PR #33, opened 2019-09-19
Basic metric notification implementation

Kyras:
- WebSocket server running on separated thread
- MetricsManager actor, listening for network operations "simulating" metrics creation.
- Broadcast message scheduled every second for the duration of the program.

### DIFF
--- a/metrics/src/metrics_manager.rs
+++ b/metrics/src/metrics_manager.rs
@@ -1,21 +1,30 @@
-use crate::ws_server::WsServer;
-use ws::{WebSocket, Sender as WsSender};
-use riker::{actor::*, system::Timer};
-use std::thread::Builder;
 use networking::p2p::network_channel::{NetworkChannelMsg, NetworkChannelTopic};
-use std::sync::{Arc, atomic::AtomicUsize};
-use std::time::Duration;
 use log::*;
-use std::collections::HashMap;
-use crate::metrics::{PeerMonitor, PeerMetrics};
-use networking::p2p::network_channel::NetworkChannelMsg::PeerCreated;
+use ws::{WebSocket, Sender as WsSender};
+use riker::{
+    system::SystemEvent, actor::*,
+    system::Timer, actors::SystemMsg,
+};
+use std::{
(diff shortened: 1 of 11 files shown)


metrics/src/metrics_manager.rs
```diff
@@ -0,0 +1,135 @@
...
+    fn receive(&mut self, _ctx: &Context<Self::Msg>, msg: NetworkChannelMsg, _sender: Sender) {
+        use std::mem::size_of_val;
+        match msg {
+            NetworkChannelMsg::PeerCreated(msg) => {
+                // TODO: Add field to message with peer name / identifier
+                let peer_name = msg.peer.name().to_owned();
+                let monitor = PeerMonitor::new(peer_name.clone());
+                self.peer_monitors.insert(peer_name, monitor);
+            }
+            NetworkChannelMsg::PeerDisconnected(msg) => {
```
mrjoe7: Instead of peer_disconnected you can listen to system event like we do here https://github.com/simplestaking/tezos-rs/blob/master/shell/src/peer_manager.rs#L103-L107

metrics/src/metrics_manager.rs
```diff
@@ -0,0 +1,135 @@
...
+        use serde_json::to_string;
+        use std::sync::atomic::Ordering::Relaxed;
+
+        if self.connected_clients.load(Relaxed) > 0 {
+            // There are clients connected to the WebSocket
+            let snapshot_data: HashMap<&String, PeerMetrics> = self.peer_monitors.iter_mut()
+                .map(|(k, v)| {
+                    (k, v.snapshot())
+                }).collect();
+            if let Ok(message) = to_string(&snapshot_data) {
```
mrjoe7: I would prefer to use `serde_json::to_string` here, so it's obvious we are serializing struct to json string.

Kyras: Fixed

metrics/src/metrics_manager.rs
Kyras: Fixed

PR #55, opened 2019-10-07
Mlacko@message listener

Kyras:
Implemented network message listening and storing in the RocksDB, as well as tooling to enable such recording.


monitoring/src/listener/listener.rs
```diff
@@ -0,0 +1,54 @@
...
+use log::*;
+
+type NetworkListenerRef = ActorRef<NetworkChannelMsg>;
+
+pub struct NetworkListener {
+    storage: File,
+    network_channel: NetworkChannelRef,
+}
+
+impl NetworkListener {
```
mrjoe7: Name suggests that thsi actor is listening to a network (connection?). Maybe use some other name which will be more descriptive of what the actor does.

monitoring/src/listener/listener.rs
```diff
@@ -0,0 +1,54 @@
...
+    fn name() -> &'static str { "network-listener" }
+
+    fn new((storage_path, network_channel): (&'static str, NetworkChannelRef)) -> Self {
+        Self {
+            storage: File::open(storage_path).expect("Failed to open the file"),
```
mrjoe7: Return result instead of panicking.

Kyras: This is an out-dated file implementation, fixed in later commit.

monitoring/src/listener/records.rs
```diff
@@ -0,0 +1,149 @@
...
+    type Key = RocksStamp;
+    type Value = Vec<u8>;
+}
+
+// --- Record implementation --- //
+
+/// Simple timestamp, denoting passed seconds from start of the node
+/// implementing the codec trait for simple database serialization.
+#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
+pub struct RocksStamp(f32);
```
mrjoe7: What is the reason behind using `f32` here instead of using `u32` and saying that it represents milliseconds from the start of a node? It will help you to avoid floating point precision problems splus you will be able to use https://doc.rust-lang.org/std/primitive.i32.html#method.from_le_bytes instead of `byteorder` crate functions.

monitoring/src/listener/listener.rs
Kyras: Actor does indeed listen to the Network messages sent on the NetworkChannel. NetworkChannelListener was chosen as more descriptive name.

monitoring/src/listener/records.rs
```diff
@@ -0,0 +1,149 @@
...
+    }
+}
+
+impl Codec for RocksStamp {
+    #[inline]
+    fn decode(data: &[u8]) -> Result<Self, SchemaError> {
+        if data.len() < 4 {
+            Err(SchemaError::DecodeError)
+        } else {
+            Ok(Self((&data[0..4]).read_f32::<LittleEndian>().map_err(|_| SchemaError::DecodeError)?))
```
mrjoe7: See my previous comment regarding the use of a `f32`

monitoring/src/listener/records.rs
Kyras: It is a mostly space saving solution. Millisecond counting requires 128 bit integral type. 32 bit floating point number is precise enough for this use-case, and Rust have a feature, in standard library, to serialize it with the `f32::to_le_bytes` and `f32::from_le_bytes`. Which is, in my opinion good  compromise.

monitoring/src/listener/records.rs
```diff
@@ -0,0 +1,149 @@
...
+#[derive(Debug, Clone)]
+/// Record metadata, describing incoming message.
+pub struct RecordMeta {
+    /// Description of type of incoming message, and stored format.
+    pub record_type: RecordType,
+    /// Representation of an peer actor, to whom belongs the message
+    pub peer_id: String,
+}
+
+impl Codec for RecordMeta {
```
mrjoe7: Is there a reason you stopped to use `zerocopy` crate?

monitoring/src/listener/records.rs
```diff
@@ -0,0 +1,149 @@
...
+            Some(PeerReceivedMessage)
+        } else {
+            None
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+/// Record metadata, describing incoming message.
+pub struct RecordMeta {
```
mrjoe7: did you split the record value into 2 structures as a optimalization?

Because the
```
pub struct RecordMeta {
    /// Description of type of incoming message, and stored format.
    pub record_type: RecordType,
    /// Representation of an peer actor, to whom belongs the message
    pub peer_id: String,
}
```

and the 

```
pub enum RecordType {
    /// Connection to new peer was accepted. No record in Record Storage should correspond to
    /// this meta message
    PeerCreated,
    /// Full connection to peer was established. Record should contains a string with public
    /// key of connected peer.
    PeerBootstrapped,
    /// Node received a message from a peer. Record should contain raw message, which
    /// should be deserialized manually.
    PeerReceivedMessage,
}
```

could be replaced by something like this:
```
pub enum RecordType {
    /// Connection to new peer was accepted. No record in Record Storage should correspond to
    /// this meta message
    PeerCreated { peer_id: String },
    /// Full connection to peer was established. Record should contains a string with public
    /// key of connected peer.
    PeerBootstrapped { peer_id: String },
    /// Node received a message from a peer. Record should contain raw message, which
    /// should be deserialized manually.
    PeerReceivedMessage { peer_id: String },
}
```

mrjoe7:
Agreed changes:
- change the key to syntetic counter to avoid message overwriting
- consider different name for meta data

monitoring/src/listener/records.rs
mrjoe7: Add `#[test]` for each `Codec` to make sure that the `struct_1 -> serialize -> bytes -> deserialize -> stuct_2` will produce `struct_1 == struct_2`.

monitoring/src/listener/records.rs
mrjoe7: will be replaces by counter value

monitoring/src/listener/records.rs
mrjoe7: Will be replaced by a couter value.

monitoring/src/listener/records.rs
mrjoe7: Consider removal of the `meta` from the name and replace it by some better name.

monitoring/src/listener/records.rs
Kyras: Done in [7631196](https://github.com/simplestaking/tezos-rs/pull/55/commits/7631196433a997eea4bf201dd9dfd47756a54c07)

monitoring/src/listener/records.rs
Kyras: Done in [ed32084](https://github.com/simplestaking/tezos-rs/pull/55/commits/ed3208499857724044ce266bafa851d390e74c53)

monitoring/src/listener/records.rs
Kyras: `RecordMeta` renamed to `Event`; `RecordType` to `EventType` in [7ee5446](https://github.com/simplestaking/tezos-rs/pull/55/commits/7ee544663680af1dea99a03173fb23b43c065048)

PR #87, opened 2019-11-11
Jselep@t 70 monitoring stats

jurajselep:


monitoring/Cargo.toml
```diff
@@ -22,3 +22,4 @@ slog_derive = "0.1.1"
 networking = { path = "../networking" }
 tezos_messages = { path = "../tezos/messages" }
 tezos_encoding = { path = "../tezos/encoding" }
+rand = ""
```
mrjoe7: We should be using at least a major verion number.

monitoring/src/monitors/bootstrap_monitor.rs
```diff
@@ -37,6 +37,19 @@ impl BootstrapMonitor {
...
+    pub fn level(&self) -> usize {
+        self.level
+    }
+    #[inline]
+    pub fn set_level(&mut self, level: usize)  {
+        // only set peer current head level for values higher than the current level
+        self.level = if self.level > level { self.level } else { level }
+    }
+    #[inline]
+    pub fn set_downloaded_blocks(&mut self, downloaded_blocks: usize)  {
```
mrjoe7: What is intended difference between `set_level` and this method? both are mutating the same variable `self.level`.

monitoring/src/monitors/chain_monitor.rs
```diff
@@ -0,0 +1,164 @@
...
+    // number of block per cycle
+    blocks_per_cycle: usize,
+    // cycle is used to measure time
+    chain: Vec<Cycle>,
+}
+
+impl ChainMonitor {
+    pub fn new() -> Self {
+        Self {
+            blocks_per_cycle: 4096,
```
mrjoe7: You should use [associated constants](https://doc.rust-lang.org/edition-guide/rust-2018/trait-system/associated-constants.html) here.

monitoring/src/monitors/chain_monitor.rs
```diff
@@ -0,0 +1,164 @@
...
+            chain: vec![],
+        }
+    }
+
+    // get cycle id number
+    fn cycle_id(&mut self, block_level: usize) -> usize {
```
mrjoe7: `&mut self` is not needed here. Instead use just `&self` or replace call to `self.blocks_per_cycle` by a named constant.

monitoring/src/monitors/chain_monitor.rs
```diff
@@ -0,0 +1,164 @@
...
+        block_level / self.blocks_per_cycle
+    }
+
+    // extend chain to support new cycles
+    fn update_chain(&mut self, block_level: usize) {
+        // cycle number
+        let cycle_id = self.cycle_id(block_level);
+        // get numner of cycles in chain
```
mrjoe7: typo in `numner`

monitoring/src/monitors/chain_monitor.rs
```diff
@@ -0,0 +1,164 @@
...
+                .map(|id| -> Cycle {
+                    Cycle {
+                        id: chain_len + id,
+                        headers: 0,
+                        operations: 0,
+                        applications: 0,
+                        start: Instant::now(),
+                        duration: None,
+                    }
+                })
```
mrjoe7: You can rewrite this as:
```rust
let mut chain_append = (chain_len..=cycle_id)
	.map(|id| -> Cycle {
		Cycle {
			id
			headers: 0,
			operations: 0,
			applications: 0,
			start: Instant::now(),
			duration: None,
		}
	})
```

monitoring/src/monitors/chain_monitor.rs
```diff
@@ -0,0 +1,164 @@
...
+        // cycle number
+        let cycle_id = self.cycle_id(block_level);
+        // use reference to update cycle stats
+        let cycle: &mut Cycle = &mut self.chain[cycle_id];
+        // increment block operations value
+        cycle.operations += 1;
+        // save duration for cycle download time at end of cycle
+        if cycle.operations == self.blocks_per_cycle {
+            // save duration in seconds
+            cycle.duration = Some(cycle.start.elapsed().as_secs_f32())
```
mrjoe7: Instead of using number of seconds you can create `Instant::now()` here. After doing that you will benefit from compiler type checking your time operations.

monitoring/src/monitors/chain_monitor.rs
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    // process block applications
+    pub fn process_block_application(&mut self, block_level: usize) {
+        // extend chain to support new cycles
+        self.update_chain(block_level);
+        // cycle number
+        let cycle_id = self.cycle_id(block_level);
+        // use reference to update cycle stats
+        let cycle: &mut Cycle = &mut self.chain[cycle_id];
```
mrjoe7: This block is used on all functions so I think it should be replaced by a function returning `&mut Cycle`.

monitoring/src/monitors/bootstrap_monitor.rs
Kyras: set_downloaded_blocks is more for initialization and special cases, where you don't intend to do any checks

monitoring/src/monitors/chain_monitor.rs
Kyras: Default `std::time::Duration` serialization is too wild to be used in any standard API.

monitoring/Cargo.toml
Kyras: Fixed in 8acf122

monitoring/src/monitors/chain_monitor.rs
Kyras: Fixed in 8acf122

monitoring/src/monitors/chain_monitor.rs
Kyras: Fixed in 8acf122

monitoring/src/monitors/chain_monitor.rs
Kyras: Fixed in 8acf122

monitoring/src/monitors/chain_monitor.rs
Kyras: Fixed in 8acf122

monitoring/src/monitors/chain_monitor.rs
Kyras: Fixed in 8acf122

PR #94, opened 2019-11-13
RPC: stats/memory

tisak:
shell/stats/memory module
implementation of riker ask for stats/memory
output according spec in http://tezos.gitlab.io/api/rpc.html#get-stats-memory


shell/src/stats/memory.rs
```diff
@@ -0,0 +1,221 @@
...
+
+impl fmt::Display for MemoryStatsError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.write_str((self as &dyn error::Error).description())
+    }
+}
+
+impl error::Error for MemoryStatsError {
+    fn description(&self) -> &str {
+        match *self {
```
mrjoe7: Is dereferencing required here?

shell/src/stats/memory.rs
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    /// private helpers
+
+    fn get_pid(opt_pid: Option<i32>) -> i32 {
+        opt_pid.unwrap_or(nix::unistd::getpid().as_raw())
+    }
+
+    fn get_mac_data(&self) -> MemoryStatsResult<MemoryData> {
+        let output = Command::new("sh")
```
mrjoe7: I believe it should be possible to execute `ps` without wrapping `sh -c` over it.

shell/src/stats/memory.rs
```diff
@@ -0,0 +1,221 @@
...
+use std::fs::File;
+use std::io::prelude::*;
+use std::path::Path;
+use std::process::Command;
+use core::str::{from_utf8};
+
+pub type MemoryStatsResult<T> = std::result::Result<T, MemoryStatsError>;
+
+#[derive(Serialize, PartialEq, Clone, Debug)]
+pub struct LinuxData {
```
mrjoe7: Please provide description to structure fields. For example `dt` function is not obvious to a reader.

rpc/src/rpc_actor.rs
```diff
@@ -183,4 +184,27 @@ impl Receive<GetFullCurrentHead> for RpcServer {
...
\ No newline at end of file
+}
+
+impl Receive<GetStatsMemory> for RpcServer {
+    type Msg = RpcServerMsg;
+
+    fn receive(&mut self, ctx: &Context<Self::Msg>, msg: GetStatsMemory, sender: Sender) {
+        if let GetStatsMemory::Request = msg {
+            if let Some(sender) = sender {
+                let me: Option<BasicActorRef> = ctx.myself().into();
```
mrjoe7: You don't need to specify types here
```rust
let me = ctx.myself().into();
```

shell/src/stats/memory.rs
```diff
@@ -0,0 +1,221 @@
...
+}
+
+impl From<DarwinOsData> for MemoryData {
+    fn from(data: DarwinOsData) -> Self {
+        MemoryData::DarwinOs(data)
+    }
+}
+
+#[derive(PartialEq, Debug)]
+pub enum MemoryStatsError {
```
mrjoe7: Use [failure](https://github.com/rust-lang-nursery/failure) crate to express errors.

shell/src/stats/memory.rs
```diff
@@ -0,0 +1,221 @@
...
+    static ref MAC_DATA: Regex = Regex::new(r"\d+\s+(?P<mem>[0-9.]+)\s+(?P<resident>\d+)").expect("Invalid regex");
+}
+
+pub struct Memory {
+    pub pid: i32,
+    pub page_size: usize
+}
+
+impl Memory {
+    pub fn new(opt_pid: Option<i32>) -> Self {
```
mrjoe7: `opt_pid` is never used (is always None) and should be probably removed

shell/src/stats/memory.rs
```diff
@@ -0,0 +1,221 @@
...
+    resident: String,
+    shared: String,
+    text: String,
+    lib: String,
+    data: String,
+    dt: String
+}
+
+#[derive(Serialize, PartialEq, Clone, Debug)]
+pub struct DarwinOsData {
```
mrjoe7: I noticed that `LinuxData` and `DarwinOsData` structures have different shape. This will be a problem for the consumers of the rpc api. Rpc api should be consistent and should not change between platforms eg. between linux and mac.

tisak: check spec at http://tezos.gitlab.io/api/rpc.html#get-stats-memory

there is different output for mac and linux, if we want be compatible with them

mrjoe7: cool, didn't know that

shell/src/stats/memory.rs
tisak: Done. Lesson learned: do not use apostrophes or spaces in args, they will be escaped and command will not work

rpc/src/rpc_actor.rs
tisak: `error[E0282]: type annotations needed
   --> rpc/src/rpc_actor.rs:195:21
    |
195 |                 let me = ctx.myself().into();
    |                     ^^ consider giving `me` a type`

shell/src/stats/memory.rs
```diff
@@ -54,31 +54,16 @@ impl From<DarwinOsData> for MemoryData {
     }
 }
 
-#[derive(PartialEq, Debug)]
+#[derive(PartialEq, Debug, Fail)]
 pub enum MemoryStatsError {
-    InvalidPid,
-    FileRead,
+    #[fail(display = "IO error: {}", _0)]
+    IOErr(String),
```
mrjoe7: just a nitpick - do not use abbreviation (use `IOError`)

tisak:
Pull request is now rebased to new master and after that I removed ask patern.  @mrjoe7 please review, thank you.

mrjoe7:
We require all commits to be signed. Please squash commits in this PR and sign the commit with your key. https://help.github.com/en/github/authenticating-to-github/signing-commits

PR #140, opened 2019-12-10
Te 84

Kyras:


light_node/src/main.rs
```diff
@@ -82,6 +84,8 @@ fn create_logger(env: &crate::configuration::Environment) -> Logger {
 fn block_on_actors(env: &crate::configuration::Environment, identity: Identity, actor_system: ActorSystem, init_info: TezosStorageInitInfo, persistent_storage: PersistentStorage, protocol_commands: IpcCmdServer, protocol_events: IpcEvtServer, protocol_runner_run: Arc<AtomicBool>, log: Logger) {
     let mut tokio_runtime = Runtime::new().expect("Failed to create tokio runtime");
 
+    let ctx_storage = Arc::new(RwLock::new(SkipList::new(0, &persistent_storage).expect("failed to initialize context storage")));
```
mrjoe7: We should not pass skip_list storage as a separate variable, but we should extent `PersistentStorage` to include it.

mrjoe7: Also we should hide internal mutability of the skip_list so `RwLock` won't be needed here. It's an implementation detail which the caller of the skip list doesn't need to know of.

light_node/src/main.rs
```diff
@@ -249,6 +253,8 @@ fn main() {
         context_storage::ContextPrimaryIndex::descriptor(),
         context_storage::ContextByContractIndex::descriptor(),
         SystemStorage::descriptor(),
+        SkipList::<String, Vec<u8>, HashMap<String, Vec<u8>>>::descriptor(),
```
mrjoe7: Use type alias instead of `Vec<u8>` to be more explicit about data stored in the hash map.

rpc/src/server/service.rs
```diff
@@ -357,6 +366,9 @@ mod fns {
 
     use crate::helpers::FullBlockInfo;
     use crate::rpc_actor::RpcCollectedStateRef;
+    use crate::CtxStorage;
```
mrjoe7: Do not use abbreviations in type names.

shell/src/context_listener.rs
```diff
@@ -118,35 +126,77 @@ fn listen_protocol_events(
...
         }
-
     }
 
     Ok(())
 }
 
+/// Get value contained in the dictionary, or create new entry associated with given key
+/// populated with default value
+fn get_default(state: &mut HashMap<Vec<u8>, Ctx>, block_hash: Vec<u8>) -> &mut Ctx {
```
mrjoe7: `get_default_WHAT` - use more descriptive function name

shell/src/context_listener.rs
```diff
@@ -118,35 +126,77 @@ fn listen_protocol_events(
...
+                        let state = get_default(&mut blocks, block_hash.clone());
+                        state.insert(key.join("/"), Bucket::Deleted);
+
+                        context_storage.put(&block_hash.clone(), &ContextRecordValue::new(msg))?;
+                    }
+                    ContextAction::Commit { new_context_hash, block_hash: Some(block_hash), .. } => {
+                        if let Some(block) = blocks.get(block_hash) {
+                            state.merge(block);
+                        } else {
+                            warn!(log, "Trying to commit non-existent block"; "block" => format!("{:?}", block_hash));
```
mrjoe7: use `HashType` to format block hash here

shell/src/context_listener.rs
```diff
@@ -118,35 +126,77 @@ fn listen_protocol_events(
...
+                        let from_key = from_key.join("/");
+                        let to_key = key.join("/");
+                        if let Some(value) = partial_state.get(&from_key) {
+                            let value = value.clone();
+                            state.insert(to_key, value);
+                        } else if let Some(value) = state.get(&from_key) {
+                            let value = value.clone();
+                            state.insert(to_key, value);
+                        } else {
+                            warn!(log, "Trying to copy from non-existent location"; "from" => format!("{:?}", from_key), "to" => format!("{:?}", to_key));
```
mrjoe7: is `format` macro here necessary? Please try to use 
```
"from" => from_key
```
if it works here.

storage/skip_list/src/skip_list.rs
```diff
@@ -33,17 +33,17 @@ impl<C: ListValue> KeyValueSchema for SkipList<C> {
     }
 }
 
-impl<C: ListValue> SkipList<C> {
+impl<K: Codec, V: Codec, C: ListValue<K, V>> SkipList<K, V, C> {
     /// Create new list in given database
-    pub fn new(list_id: SkipListId, db: Arc<rocksdb::DB>) -> Result<Self, SkipListError> {
-        let list_db: Arc<SkipListDatabase<C>> = db.clone();
+    pub fn new(list_id: SkipListId, persistent_storage: &storage::persistent::PersistentStorage) -> Result<Self, SkipListError> {
```
mrjoe7: Do we need to pass the whole `PersistentStorage` here? This is a specialized type which should not have a need for the whole persistent storage.

storage/skip_list/src/skip_list.rs
```diff
@@ -157,6 +191,56 @@ impl<C: ListValue> SkipList<C> {
             f64::log((index + 1) as f64, LEVEL_BASE as f64).floor() as usize
         }
     }
+
+    /// Build a difference object between two indexes.
+    /// `list.get(from).merge(list.diff(from, to)) == list.get(to)`
+    #[allow(dead_code)]
```
mrjoe7: don't forget ot remove

light_node/src/main.rs
Kyras: If we want to use SkipList as a general data structure, this is the way to go. RwLock does not hide inner mutability, just implements the missing thread synchronicity. Only other way would be thread safe rewrite, or some inner RwLock juggling.

shell/src/context_listener.rs
Kyras: `get(default)` and `set_default` are common method found on hashmap in other languages.
`get_default` -  Either gets value associated with the key or associate it with some default value and then returns it (in our case Default::default).

`set_default(k ,v)` - If key is present in hashmap, it works exactly as `get`. Otherwise it sets the key to the given value `v` first.

light_node/src/main.rs
Kyras: Changed in 4b31638 to `<String, Bucket<Vec<u8>>, ContextMap>`. The `Vec<u8>` type is most explicit type we have, as we literally store *some* binary data, we do not have full encoding for.

rpc/src/server/service.rs
Kyras: Changed in 4b31638

shell/src/context_listener.rs
Kyras: Removed in 4b31638

storage/skip_list/src/skip_list.rs
Kyras: We will surely get to back to this, I would recommend to merge this, as it shows at least base implementation, of what we are trying to do.

light_node/src/main.rs
Kyras: This is non-trivially solvable cyclical dependency problem, as it requires refactorization of whole `storage` folder.
`SkipList` requires many traits in `storage` crate, but extending `PersistentStorage` to include `SkipList` would make cyclical dependency between `storage` and `skip_list` crate.

shell/src/context_listener.rs
Kyras: Fixed in c97f893

storage/skip_list/src/skip_list.rs
Kyras: Fixed in c97f893

light_node/src/main.rs
mrjoe7: Skip list does not require the whole `PersistentStorage`. You can pass only `Arc<rocksdb::DB>` to it, that will solve dependency problem.

mrjoe7: I see you did the `Arc<rocksdb::DB>` in https://github.com/simplestaking/tezedge/commit/c97f893d88a5f82935933f4ec04a5ba42f4436ad
Now the cyclical dependency should be there there.

Kyras: Cyclical dependency happens, because `PersistentStorage` struct is in same crate as many traits like `BincodeEncoded` and `KeyValueStoreWithSchema`, which are essential to database backed structure, so even if `SkipList` does not directly requries the `PersistentStorage` the cyclical dependency still holds. Only way to somehow fix this, is to move `PersistentStorage` somewhere else.

mrjoe7: We could move `SkipList` into a separate module inside of the `storage` package.

Kyras: fixed in c694da1

PR #142, opened 2019-12-16
Mlacko@te 87 extended constants

Kyras:
Implementation of `/chains/{chain_id}/blocks/{blocks_id}/context/constants`, this should be merged AFTER TE-84


tezos/encoding/src/binary_reader.rs
```diff
@@ -222,6 +222,35 @@ impl BinaryReader {
                     Ok(Value::String(str_num))
                 }
             }
+            Encoding::Mutez => {
```
mrjoe7: Add unit tests for `Mutez` encoding.

PR #149, opened 2019-12-19
Rpc GET /chains/<chain_id>/blocks/<block_id>/helpers/baking_rights

adonagy:
Here is the implementation of the baking_rights rpc without the tezos PRNG generator. It is replaced by the generic rust rng from the rand crate. Next step is to implement the tezos PRNG generator and include it in this PR.

The goal of this Draft is to review the code to date. Most likely it will need some refactoring.

### DIFF
--- a/rpc/src/helpers.rs
+++ b/rpc/src/helpers.rs
@@ -127,3 +127,22 @@ impl<C> PagedResult<C>
         PagedResult { data, next_id, limit }
     }
 }
+#[derive(Serialize, Debug, Clone)]
+/// Object containing information about the baking rights 
+pub struct BakingRights {
+    pub level: i32,
+    pub delegate: String,
+    pub priority: i32,
+    pub estimated_time: String,
+}
+
+impl BakingRights {
+    pub fn new(level: i32, delegate: String, priority: i32, estimated_time: String) -> Self{
+        Self {
+            level,
+            delegate: delegate.to_string(),
+            priority,
+            estimated_time: estimated_time.to_string(),
+        }
+    }
+}
--- a/rpc/src/helpers.rs
+++ b/rpc/src/helpers.rs
@@ -133,18 +133,18 @@ impl<C> PagedResult<C>
 #[derive(Serialize, Debug, Clone, Getters)]
 pub struct BakingRights {
     #[get = "pub(crate)"]
-    level: i32,
+    level: i64,
     #[get = "pub(crate)"]
     delegate: String,
     #[get = "pub(crate)"]
-    priority: i32,
+    priority: i64,
     #[serde(skip_serializing_if = "Option::is_none")]
     #[get = "pub(crate)"]
     estimated_time: Option<String>,
 }
 
 impl BakingRights {
-    pub fn new(level: i32, delegate: String, priority: i32, estimated_time: Option<String>) -> Self{
+    pub fn new(level: i64, delegate: String, priority: i64, estimated_time: Option<String>) -> Self{
         Self {
             level,
             delegate: delegate.to_string(),
--- a/rpc/src/helpers.rs
+++ b/rpc/src/helpers.rs
@@ -194,3 +194,4 @@ impl StakingRightsContextData {
         }
     }
 }
+
--- a/rpc/src/helpers.rs
+++ b/rpc/src/helpers.rs
@@ -128,7 +128,6 @@ impl<C> PagedResult<C>
         PagedResult { data, next_id, limit }
     }
 }
-#[derive(Serialize, Debug, Clone)]
 /// Object containing information about the baking rights 
 #[derive(Serialize, Debug, Clone, Getters)]
 pub struct BakingRights {
--- a/rpc/src/helpers.rs
+++ b/rpc/src/helpers.rs
@@ -157,9 +157,21 @@ impl BakingRights {
 #[derive(Serialize, Debug, Clone, Getters)]
 pub struct StakingRightsContextData {
     #[get = "pub(crate)"]
+    head_level: i32,
+    #[get = "pub(crate)"]
+    block_level: i32,
+    #[get = "pub(crate)"]
     roll_snapshot: i16,
     #[get = "pub(crate)"]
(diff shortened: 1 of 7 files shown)


rpc/src/helpers.rs
```diff
@@ -106,3 +106,24 @@ impl FromStr for BlockHeaderInfo {
         })
     }
 }
+
+#[derive(Serialize, Debug, Clone)]
+/// Object containing information about the baking rights 
+pub struct BakingRights {
+    pub level: i32,
```
mrjoe7: Try to avoid pub fields if possible. We are using https://crates.io/crates/getset for easy field encapsulation.

rpc/src/server/service.rs
```diff
@@ -406,6 +446,197 @@ mod fns {
...
+        let requested_cycle = match cycle.parse() {
+            // cycle requested from query
+            Ok(cycle) => {
+                // check whether the cycle is reachable
+                // the storage stores cycle up to: current_cycle + perserved_cycles(5) + 2
+                // e.g. if head is in cycle 66, we can get the rights for up to cycle 73
+                if ((cycle as i32 - (block_level as i32 / 2048)).abs() as i32) <= 7 {
+                    cycle
+                } else {
+                    panic!("Cycle not found in block: {}", block_id)
```
mrjoe7: No panics

rpc/src/server/service.rs
```diff
@@ -406,6 +446,197 @@ mod fns {
...
+                }
+            },
+            // no cycle requested
+            Err(_e) => -1
```
mrjoe7: Please do not use magic constants. Be more expressive, Rust is not a C.

rpc/src/server/service.rs
```diff
@@ -516,6 +747,32 @@ mod fns {
         Ok(contract_address)
     }
 
+    #[inline]
+    fn address_to_contract_id(address: &str) -> Result<String, failure::Error> {
```
mrjoe7: This function is really unsafe because it's check data boundaries inside. Please refactor into a safe form.

rpc/src/server/service.rs
```diff
@@ -406,6 +446,197 @@ mod fns {
...
+        let rolls = get_rolls(block_id, &cycle, persistent_storage, list).unwrap();
+        let roll_owners = match rolls {
+            Some(r) => r,
+            None => panic!("Error getting rolls")
+        };
+
+        // iterate through the whole cycle if necessery
+        for level in requested_level..(level_to_itarate_to + 1) {
+            // 30 is from the protocol constants
+            let seconds_to_add = (level - head_level).abs() * 30;
```
Kyras: It would be probably a good idea to actually read the value from the context constants. There is already prepared encoding for all protocols to decode them from context

rpc/src/server/service.rs
```diff
@@ -406,6 +446,197 @@ mod fns {
...
+                ctx
+            } else {
+                panic!("Context not found")
+            }
+        };
+
+        let roll_data = context.clone();
+        // get all the relevant data out of the context database
+        let data: HashMap<String, Bucket<Vec<u8>>> = context.into_iter()
+            .filter(|(k, _)| k.contains("data/rolls/owner/current") || k.contains("/successor") || k.contains("/random_seed"))
```
Kyras: As an improvement, also add a check, that value is `Bucket::Exists` allowing safe `unwrap`s down the code.

adonagy: When accessing the created hash maps from the context, I always perform a check whether the value is a Bucket::Exists variant. (E.g [here](https://github.com/adonagy/tezedge/blob/baking_rights/rpc/src/server/service.rs#L515)). Should I add the check nevertheless?

rpc/src/server/service.rs
adonagy: Can you elaborate please? I'm not sure how to refactor this into a safe form.

mrjoe7: For example, you have ```&address[2..4]``` on line 755, which will panic in case address length is less than 4. This function should check length of it's input data.

rpc/src/server/service.rs
```diff
@@ -406,6 +447,213 @@ mod fns {
...
+                }
+            },
+            // no cycle requested
+            Err(_e) => NO_CYCLE_FLAG
+        };
+
+        // If no level is specified, we return the next block to be baked
+        let mut requested_level: i64 = match level {
+            Some(level) => level.parse()?,
+            None => head_level + 1,
```
tisak: maybe blok_level should be here?

adonagy: Yeah, you are right, will fix this

rpc/src/server/service.rs
```diff
@@ -406,6 +447,213 @@ mod fns {
...
+                // construct the whole owner/successor key
+                // /<rol_num little-endian 1st byte>/<roll_num little-endian 2nd byte>/<roll_num in decimal>
+                // Note: this will work up to 0xffff rolls, not sure how exactly they implement rolls over this number with this key format
+                let owner_key = format!("data/rolls/owner/current/{}/{}/{}", next_roll[3], next_roll[2], roll_num);
+                let successor_key = format!("data/rolls/index/{}/{}/{}/successor", next_roll[3], next_roll[2], roll_num);
+            
+                if let Some(Bucket::Exists(_roll)) = data.get(&owner_key) {
+                    roll_owners.insert(roll_num.into(), contract_id.clone());
+                } else {
+                    bail!("Roll owner key not found for key: {}", &owner_key)
```
tisak: there can be also Bucket::Deleted in owner_key, check for it or put break instead of bail

rpc/src/server/service.rs
adonagy: Right, I get it now, Thanks

adonagy:
I've added the implementation for the tezos prng.

There are still some bugs to sort out in this rpc call with the roll snapshots. It is still a WIP.

adonagy:
Closing this PR, moved to #161

PR #156, opened 2020-01-21
TE-89 helpers/endorsing_rights #1

tisak:
RPC /chains/:chain_id/blocks/:block_id/helpers/endorsing_rights

For now it works with first 5 cycles.

TODO:
1. verify computation of roll_snapshot - there seems to be different approach for levels 1-5, 6-10, 10 ...
2. verify computation of last_roll


rpc/src/helpers.rs
```diff
@@ -104,22 +105,89 @@ impl<C> PagedResult<C>
     }
 }
 
-#[derive(Serialize, Debug, Clone, Default)]
+// Enum defining possible response structures for RPC calls
+#[derive(Serialize, Debug, Clone)]
+pub enum RpcResponseData {
```
mrjoe7: Is there a reason why `Result<Vec<EndorsingRight>, RpcErrorMsg>` was not used here?

tezos/baking/src/helpers.rs
```diff
@@ -0,0 +1,74 @@
...
+// level 0 (genesis block) is not part of any cycle (cycle 0 starts at level 1), hence the -1
+pub fn cycle_from_level(level: i32, blocks_per_cycle: i32) -> i32 {
+    (level - 1) / blocks_per_cycle
+}
+
+// the position of the block in its cycle
+// level 0 (genesis block) is not part of any cycle (cycle 0 starts at level 1)
+// hence the blocks_per_cycle - 1 for last cycle block
+pub fn level_position(level:i32, blocks_per_cycle:i32) -> i32 {
+    let cycle_position = (level % blocks_per_cycle) - 1;
```
mrjoe7: Make sure that thsi will not panic in case `blocks_per_cycle` is zero.

tezos/baking/src/helpers.rs
```diff
@@ -0,0 +1,74 @@
...
+    if cycle_position < 0 { //for last block
+        blocks_per_cycle - 1
+    } else {
+        cycle_position
+    }
+}
+
+// tezos PRNG
+pub fn get_random_number(state: Vec<u8>, nonce_size: usize, blocks_per_cycle: i32, use_string_bytes: &[u8], level: i32, offset: i32, bound: i32) -> Result<(i32, Vec<u8>), failure::Error> {
```
mrjoe7: - document this function
- it's not clear what `(i32, Vec<u8>)` returned by this function represents. Consider using type alias or create a new type.
- try to avoid using generic `failure::Error` on public functions

rpc/src/helpers.rs
tisak: Ocaml node has specific error message format as response to wrong RPC request or when specific error occurs during request, i.e.:
[{"kind":"permanent","id":"proto.005-PsBabyM1.context.storage_error","missing_key":["cycle","4","random_seed"],"function":"get"}]
[{"kind":"permanent","id":"proto.005-PsBabyM1.seed.unknown_seed","oldest":9,"requested":20,"latest":15}]

1. @jurajselep  fist question is if we want our errors to be same as ocaml node, this is not specified in Tezos RPC documentation (at least I did not found it) and maybe we should go for own errors.
2. one or other way it can be implemented also by adding context to Error and build response msg directly in result_to_json_response function

tisak: there is description of this in code comment so I am resolving it here

PR #164, opened 2020-02-04
Get block by level, static genesis data

adonagy:
- Quick fix of getting the block when the URL paramater block_id is the level of the block. 
- Add more static genesis data for babylonnet

Do not merge into master. Will need proper rework


storage/src/lib.rs
```diff
@@ -133,6 +133,17 @@ pub fn initialize_storage_with_genesis_block(genesis_hash: &BlockHash, genesis:
     if block_storage.get(&genesis_with_hash.hash)?.is_none() {
         info!(log, "Initializing storage with genesis block");
         block_storage.put_block_header(&genesis_with_hash)?;
+        // TODO: include the data for the other chains as well (mainet, zeronet, etc.)
```
bkontur: @adonagy 
create pls. jira task and assign it to me

PR #165, opened 2020-02-04
Update rust toolchain to nightly-2019-12-05

mrjoe7:


mrjoe7:
We still have to update Drone CI images to newer version of the toolchain.
There was a change in `Condvar` api so the Drone build should fail.

PR #175, opened 2020-03-02
Add RPC GET chains/main/blocks/<block_hash>/votes/listings

adonagy:
This PR adds the RPC chains/main/blocks/<block_hash>/votes/listings. The RPC returns the delegates, that are eligible to vote, with their vote's weights in rolls

### DIFF
--- a/rpc/src/encoding/conversions.rs
+++ b/rpc/src/encoding/conversions.rs
@@ -1,8 +1,45 @@
-use failure::bail;
+use std::convert::From;
+
+use failure::Fail;
+use hex::FromHexError;
 
 use crypto::hash::{ChainId, HashType};
 use crypto::blake2b;
+use crypto::base58::FromBase58CheckError;
 use storage::context_storage::ContractAddress;
+
+#[derive(Debug, Fail, PartialEq)]
+pub enum ConversionError {
+    #[fail(display = "Invalid contract id: {}", contract_id)]
+    InvalidContractId {
+        contract_id: String
+    },
+
+    #[fail(display = "Conversion from invalid public key")]
+    InvalidPublicKey,
+
+    #[fail(display = "Invalid hash: {}", hash)]
+    InvalidHash {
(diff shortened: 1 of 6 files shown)


rpc/src/encoding/conversions.rs
```diff
@@ -76,6 +76,27 @@ pub fn public_key_to_contract_id(pk: Vec<u8>) -> Result<String, failure::Error>
...
+                HashType::ContractTz2Hash.bytes_to_string(&hex::decode(&hash)?)
+            }
+            "p256" => {
+                HashType::ContractTz3Hash.bytes_to_string(&hex::decode(&hash)?)
+            }
+            _ => bail!("Invalid curve tag")
+        };
+        Ok(contract_id)
+    } else {
+        bail!("Invalid hash")
```
Kyras: It would be better to define custom error through `failure::Fail`, rather than bail. Not crucial thou.

rpc/src/server/service.rs
```diff
@@ -668,6 +670,41 @@ pub(crate) fn get_rolls_owner_current_from_context(level: &str, list: ContextLis
...
+        Some(val) => val.try_into()?,
+        None => bail!("Block level not found")
+    };
+
+    // get the whole context
+    let ctxt = get_context(&block_level.to_string(), context_list)?;
+
+    // filter out the listings data
+    let listings_data: ContextMap = ctxt.unwrap().into_iter()
+        .filter(|(k, _)| k.contains(&"data/votes/listings/"))
```
Kyras: If it is possible, use `starts_wtih` instead of `contains`. (Don't remember if it is prefix or not).
EDIT: Just notice in further comment it should be an prefix.

adonagy:
@Kyras I've added the required changes and also reworked the bail usage and created an error enum. Please check and let me know whether it is ok :)

Also added missing unit tests in the conversions

PR #184, opened 2020-03-16
Baking rights and Endorsing rights optimizations

adonagy:
Removed the merge_slices macro where it was called more then once (rights for a whole cycle). Merge_slices caused to many unnecessary allocations.

See test in drone CI (when finished running) to compare the result of the optimization.


adonagy:
Note: signed unsigned commits

PR #185, opened 2020-03-19
Context RAM storage

adonagy:
This PR is a PoC - Store the aggregated context data 

@Kyras Please take a look at the implementation. Is there a better way to a more seamless switch between the two databse backed lists and this ram storeage?


mrjoe7:
@adonagy is this PR still active?

adonagy:
@mrjoe7 indeed, I will rebase and merge this PR once I finished my current task. We will keep the skip list implementation "active". (We can switch to ram storage just like we can switch between skiplist and flatlist)

bkontur:
@adonagy 
Is this still actual?
Pls. rebase on develop or if not needed, delete branch and PR pls.

PR #204, opened 2020-04-23
Storage stats

fafk:
Add a new RPC route /dev/stats/storage to get statistics about stored actions.


storage/src/persistent/commit_log.rs
```diff
@@ -62,7 +62,7 @@ impl slog::Value for CommitLogError {
 }
 
 type ByteLimit = usize;
-type ItemCount = u16;
+type ItemCount = u32;
```
mrjoe7, line 65: Please increase `DATABASE_VERSION` in `main.rs` to indicate that format of the data in persistent storage has changed.

fafk: Has it changed, though? The ItemCount is only used to read from the database:

`fold_consecutive_locations()` uses Range, which uses ItemCount,
and `fit_batch_read_limit()` uses it as a parameter.

I increased the size, because it was causing overflow in `fold_consecutive_locations`, but if I understand it correctly, the data format hasn't changed? 🤔

mrjoe7: Yes, you are right. For some reason I remembered it being stored but that is not the case here.

mrjoe7:
We can merge it after the issue with droneIO integration is solved. Nice code by the way :+1:

mrjoe7:
Can you please squash your code into a single signed commit?

fafk:
Commits squashed and signed! :)

rpc/src/server/router.rs
```diff
@@ -38,6 +38,7 @@ pub(crate) fn create_routes() -> PathTree<Handler> {
     routes.handle("/stats/memory", dev_handler::dev_stats_memory);
     routes.handle("/p2p/:offset/:count", dev_handler::p2p_messages);
     routes.handle("/p2p/:offset/:count/:host", dev_handler::p2p_host_messages);
+    //routes.handle("/stats/storage", dev_handler::dev_stats_storage);
```
bkontur, line 41: why is this commented?

fafk: Juraj asked me to comment it out for the time being. It's a very resource-intense endpoint and if someone starts spamming it they could bring down the server.

PR #216, opened 2020-04-30
Receive action values from protocol

fafk:
After protocol changes (in GitLab) are merged, we will need to change GIT_COMMIT_HASH in interop/build.rs, imho.


tezos/interop_callback/src/callback.rs
```diff
@@ -146,47 +146,53 @@ caml!(ml_context_commit(parent_context_hash, block_hash, new_context_hash, time_
...
+    let keyval: Tuple = keyval.into();
+    let key: ContextKey = List::from(keyval.get(0).unwrap()).convert_to();
+    let value: ContextValue = OcamlBytes::from(keyval.get(1).unwrap()).convert_to();
 
     let time_period: Tuple = time_period.into();
     let start_time: f64 = time_period.get(0).unwrap().f64_val();
     let end_time: f64 = time_period.get(1).unwrap().f64_val();
 
-    context_raw_get(context_hash, block_hash, operation_hash, key, start_time, end_time);
+    context_raw_get(context_hash, block_hash, operation_hash, key, value, start_time, end_time);
```
bkontur, line 195: @fafk @jurajselep 
1. pls. do a refactor keyval also for context_raw_set like I mentioned in gitlab PR
2. add also value_as_json to get result like in set result
```
        value: Vec<u8>,
        value_as_json: Option<String>,
```

fafk: I got stuck on missing `protocol_hash` when I tried to compute the JSON value for GET with `resolve_value_as_json()`

It does on `let protocol_hash = Option.unopt_exn (Ffi_errors.Ffi_error "protocol_hash not set!") protocol_hash in`

I dug around and it seems that it's set when a block is being applied, but not quite sure how to proceed here. 🤔

fafk: As for 1.: I think you didn't actually mention it in the GitLab PR (or I didn't mange to find it), but I aligned the GET and SET function annotations to be coherent.

light_node/src/configuration.rs
```diff
@@ -74,7 +74,8 @@ pub struct Environment {
     pub storage: Storage,
     pub identity: Identity,
 
-    pub record: bool,
+    pub store_p2p_messages: bool,
```
bkontur, line 77: @fafk 
I would put this properties to: configuration.Storage struct
pub store_p2p_messages: bool,
pub store_context_actions: bool,

networking/src/p2p/peer.rs
```diff
@@ -370,7 +382,7 @@ async fn bootstrap(msg: Bootstrap, info: Arc<Local>, log: Logger, mut storage: P
         &info.proof_of_work_stamp,
         &Nonce::random().get_bytes(),
         vec![supported_protocol_version.clone()]);
-    storage.store_connection_message(&connection_message, false, addr)?;
+    if store_messages { storage.store_connection_message(&connection_message, false, addr); }
```
bkontur, line 385: @fafk @mrjoe7 
I am not a very big fan of this one line IF's :)
what is the rust recommended convention for this?

fafk: I don't think there is one. I changed it to a multi-line if. :)

PR #230, opened 2020-05-26
Removed P2P explorer from codebase

Kyras:


light_node/etc/tezedge/tezedge.config
```diff
@@ -72,7 +72,7 @@
 
 # <Optional> A peer to bootstrap the network from. Peers are delimited by a colon. Format: IP1:PORT1,IP2:PORT2,IP3:PORT3
 # --peers <IP:PORT>
-# --peers=
+# --peers=51.15.81.27:9732
```
bkontur, line 75: @Kyras 
what is this IP for?

bkontur:
@Kyras 
Can you, remove from configuration `store-p2p-messages` and all of its usage, please?

light_node/etc/tezedge/tezedge.config
Kyras: It's one of default addresses used when node is started without `--peers` argument. It is meant to be as an example in for this argument, but I will remove it for sake of consistency, as all other optional argument does not have any "sampe" value in the commented template.

Kyras: Removed in 3721a55

Kyras:
Additionally removed `--store-p2p-messages` and all mentions of it.

PR #231, opened 2020-05-28
Feature/protocol runner

bkontur:


tezos/wrapper/src/service.rs
```diff
@@ -462,25 +441,99 @@ impl Drop for ProtocolController<'_> {
...
+    /// Starts protocol runner sub-process just once and you can take care of it
+    pub fn start(&self) -> Result<Child, ProtocolServiceError> {
+        info!(self.log, "Starting protocol runner process"; "endpoint" => self.name.clone());
+        self.runner.spawn()
+    }
+
+    /// Starts protocol runner sub-process and takes care of it automatically.
+    /// If sub-process failed, it is automatically spawned another sub-process.
+    /// Returns AtomicBool, if set to false, than terminates sub-process
+    pub fn start_in_restarting_mode(&mut self) -> Result<Arc<AtomicBool>, ProtocolServiceError> {
```
mrjoe7, line 476: I think we could utilize Riker actor system here and save one thread

bkontur: I think so also, but it was not so easy on the first look :), and also in our case, ProtocolEndpoint is splitted beetween two actors (ContextListener, ChainFeeder), so I just refactored watching thread to be able easily reused the same code for another ProtocolEndpoint at first step

PR #239, opened 2020-06-08
Mlacko@action rpc

Kyras:


storage/src/context_action_storage.rs
```diff
@@ -476,7 +874,6 @@ mod tests {
 
     use super::*;
 
-    #[test]
```
bkontur, line 479: @Kyras 
What about this removed #[test]? 
You should also remove test function, if there is no need, or implement test correctly pls. ;)

Kyras: Those tests, were not correct, as the associated RPC filters did not worked correctly, and yes they should be remade to reflect the new changes.

PR #261, opened 2020-08-13
Feature/TE-201: New Interop

tizoc:
Switch to new FFI.

- [x] `ocaml-rs` replaced with `znfe`.
- [x] fixes some instances where the OCaml GC rules where not being implemented correctly.
- [x] initialize `libtezos-ffi` by passing pointers to all `ml_context_*` functions to avoid cyclic dependency, this solves the cyclic linking issue on Linux, and the hangouts on OSX.
- [x] use `Option<hash>` instead of empty strings to represent null hashes and json values
- [x] fix bug in unboxed float conversion
- [x] use unboxed floats to receive time periods
- [x] use tagged version of znfe in the dependency list
- [x] move initialization code to `ffi.rs`
- ~get rid of initialization through `lazy_static` in `runtime.rs`~ (left for later)

There is a counterpart in libtezos-ffi (`cyclic-linking-fix` branch) to this branch that is required for it to work correctly.

### DIFF
--- a/tezos/api/Cargo.toml
+++ b/tezos/api/Cargo.toml
@@ -10,7 +10,7 @@ enum-iterator = "0.6"
 failure = "0.1"
 chrono = { version = "0.4", features = ["serde"] }
 lazy_static = "1.4"
-ocaml = "0.9.3"
+znfe = { git = "https://github.com/tizoc/znfe.git", tag="v0.1.0" }
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 slog = "2.5"
--- a/tezos/interop/src/ffi.rs
+++ b/tezos/interop/src/ffi.rs
@@ -1,7 +1,12 @@
 // Copyright (c) SimpleStaking and Tezedge Contributors
 // SPDX-License-Identifier: MIT
 
-use ocaml::{List, Str, ToValue, Tuple, Value};
+use std::sync::Once;
+
+use znfe::{
+    ocaml_alloc, ocaml_call, ocaml_frame, IntoRust, OCaml, OCamlBytes, OCamlFn1, OCamlInt32,
+    OCamlList, ToOCaml,
+};
 
 use tezos_api::ffi::*;
 use tezos_api::identity::Identity;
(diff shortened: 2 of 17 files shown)


tezos/interop/src/ffi.rs
```diff
@@ -1,100 +1,113 @@
...
-}
-
-fn copy_to_ocaml(bytes: &[u8]) -> Str {
-    let mut arr = Str::new(bytes.len());
-    let data = arr.data_mut();
-    for (n, i) in bytes.iter().enumerate() {
-        data[n] = *i;
+mod tezos_ffi {
+    use znfe::{ocaml, Intnat, OCamlBytes, OCamlInt32, OCamlList};
+    ocaml! {
```
bkontur, line 16: @tizoc Bruno, I am thinking that this:
```
// TODO: probably needs to be moved to a better place
unsafe { tezos_interop_callback::initialize_callbacks() };
```
should go somewhere here, for example: https://doc.rust-lang.org/beta/std/sync/struct.Once.html "Useful for one-time initialization for FFI or related functionality" :D

it makes sense, because `mod tezos_ffi` defines ffi interface to ocaml and this calls needs those callbacks,
so everything will be defined at one place and tied together,

and maybe runtime.rs initialize_environment() should be also called here, maybe without `lazy_static OCAML_ENV`, runtime.rs will be more reusable,
or maybe some global tezos_interop::ffi_init() function will be much better then this lazy_static,

what do you think?

tezos/api/Cargo.toml
```diff
@@ -10,7 +10,7 @@ enum-iterator = "0.6"
 failure = "0.1"
 chrono = { version = "0.4", features = ["serde"] }
 lazy_static = "1.4"
-ocaml = "0.9.3"
+znfe = { git = "https://github.com/tizoc/znfe.git" }
```
bkontur, line 13: @tizoc 
I started discussion on slack about this a how to version it :)

tizoc: Ok, now it points to a specific tag.

tezos/interop/src/ffi.rs
tizoc: I moved the initialization code around in https://github.com/simplestaking/tezedge/pull/261/commits/6ac36a9885ab1df2cae0ab9b9082dfe2161e0a2e

I'm not sure about how to go around the `lazy_static OCAML_ENV` change, we still need that handle to be a singleton, because we can only have a single OCaml runtime per process. Can you expand a bit on what would be the change there?

tizoc:
@bkontur ahhh I misunderstood you, I thought you wanted to keep all the references to the libtezos-ffi functions in a single place.

PR #267, opened 2020-08-20
OSX build fixes + binaries from CI + Travis CI for OSX

tizoc:
- [x] OSX build fixes + use binaries from Gitlab CI
- [x] Travis CI for OSX
- [x] recompile libtezos so that it works with OSX 10.13 and newer
- [x] Caching of OSX stuff for faster builds (half-done, needs more work)
- [x] Fix post-accept issue with sockets in OSX (https://bugs.python.org/issue7995)
  - https://github.com/dotnet/runtime/issues/25069 and https://bugs.python.org/issue7995
- [x] linker settings for OSX as part of the project (no need for the user to set it up)
- [x] instructions for OSX in README (homebrew deps)

### DIFF
--- a/.cargo/config
+++ b/.cargo/config
@@ -1,0 +1,3 @@
+[target.x86_64-apple-darwin]
+# Required in OSX so that the compiler allows undefined symbols when linking dynamic libraries
+rustflags = ["-C", "link-args=-Wl,-undefined,dynamic_lookup"]
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,10 +2,14 @@ language: rust
 rust:
   - nightly-2020-07-12
 
-os:
-  - linux
-
-dist: focal
+jobs:
+  include:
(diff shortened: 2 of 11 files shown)


tizoc:
There is still something weird going on with linking, because instead of using `libtezos.dylib`, the produced binaries try to use `src/lib_ffi/ffi_main.so` instead (I have to make a symlink or copy for it to work):

`dyld: Library not loaded: src/lib_ffi/ffi_main.so`

I have to figure that out and make a few more changes to `build.rs` so that it uses the `.dylib` extension when copying the lib file in OSX.

tizoc:
Ok, OSX has this thing called [install_name_tool](http://manpagez.com/man/1/install_name_tool/), what I have to do is to update the `dune` file in tezos to correct `libtezos-ffi.dylib` and set the `install_name` to `libtezos.dylib`.

PR #268, opened 2020-08-21
Enable core dumps for the node and increase max open files limit

michal-stanek:
On node startup, enable core dumps for debug builds. Also increase the limit of max open file descriptors to 64k.
Unsafe block is required to use direct libc syscall functions.
Also remove the db-cfg-max-open-files setting and leave the default RocksDB value (unlimited).
More details in commit logs.

### DIFF
--- a/light_node/Cargo.toml
+++ b/light_node/Cargo.toml
@@ -13,6 +13,7 @@ futures = "0.3"
 hex = "0.4"
 lazy_static = "1.4"
 riker = "0.4"
+rlimit = "0.4.0"
 serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 slog = { version = "2.5", features = ["max_level_trace"] }
--- a/light_node/src/main.rs
+++ b/light_node/src/main.rs
@@ -36,6 +36,7 @@ use crate::configuration::LogFormat;
 
 mod configuration;
 mod identity;
+mod system;
 
 const DATABASE_VERSION: i64 = 15;
 const SUPPORTED_DISTRIBUTED_DB_VERSION: u16 = 0;
@@ -251,6 +252,9 @@ fn main() {
 
     // Creates default logger
     let log = create_logger(&env);
+
+    // Enable core dumps
+    system::init_limits(&log);
 
     let actor_system = SystemBuilder::new().name("light-node").log(log.clone()).create().expect("Failed to create actor system");
 
--- a/light_node/src/main.rs
+++ b/light_node/src/main.rs
@@ -253,7 +253,7 @@ fn main() {
     // Creates default logger
     let log = create_logger(&env);
 
-    // Enable core dumps
+    // Enable core dumps and increase open files limit
     system::init_limits(&log);
 
     let actor_system = SystemBuilder::new().name("light-node").log(log.clone()).create().expect("Failed to create actor system");
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - New configuration parameter `--disable-bootstrap-lookup` to turn off dns lookup for peers (e.g. used for tests or sandbox)
-- New configuration parameter `--db-cfg-max-threads` and `--db-cfg-max-open-files` to better control system resources
+- New configuration parameter `--db-cfg-max-threads` to better control system resources
 - New RPCs to make baking in sandbox mode possible with tezos-client
 - Support for macOS (10.13 and newer).
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1356,6 +1356,7 @@ dependencies = [
  "monitoring",
  "networking",
  "riker",
+ "rlimit",
  "rpc",
  "serde 1.0.114",
  "serde_json",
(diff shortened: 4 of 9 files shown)


light_node/src/main.rs
```diff
@@ -242,7 +245,63 @@ fn block_on_actors(
     });
 }
 
+unsafe fn enable_core_dumps() {
```
bkontur, line 248: @michal-stanek 
could you pls. add all these stuff to the new file light_node\src\system.rs?
for example: in case we want to reuse, we just move this file elsewhere

light_node/src/main.rs
```diff
@@ -242,7 +245,63 @@ fn block_on_actors(
...
+    if libc::setrlimit(libc::RLIMIT_NOFILE, &rlim) != 0 {
+        let err = io::Error::last_os_error();
+        println!("Setting open files limit failed (setrlimit): {}", err);
+        return;
+    }
+    println!("Open files limit set to {}.", rlim.rlim_cur);
+}
+
 fn main() {
+    // Need unsafe to access direct libc syscall interface
```
bkontur, line 295: @michal-stanek 
and also here, just call something like:
```
system::init_limits();
```

or maybe:
```
system::init_limits(log, env);
```
and that constant **64 * 1024** could be put into the configuration as optional, just to let users possibility to set it according to their needs and default behavior will be your algo here

now, the constant is here: `env.storage.db_cfg.max_open_files` in `DbConfiguration`


is it ok set thouse rlimits after creating logger and env? or this rlimits stuff have to be first step in main.rs?
what do you thing?

bkontur: ok, ok, makes sense,
so pls. remove `env.storage.db_cfg.max_open_files`

light_node/src/main.rs
michal-stanek: Done

light_node/src/main.rs
michal-stanek: Updated the PR. I ended up using the rlimit crate instead of libc - it does look a bit nicer. I also moved the initialization after log and env as you suggested - it shouldn't make a difference and we can use logger instead of print.
I confirmed that protocol runner processes inherit the rlimits.
Please check if it looks okay, then I'll squash the commits. I'll also add a separate commit removing the db_cfg.max_open_files.

light_node/src/main.rs
```diff
@@ -252,6 +253,12 @@ fn main() {
     // Creates default logger
     let log = create_logger(&env);
 
+    // Need unsafe to access direct libc syscall interface
+    unsafe {
```
bkontur, line 257: @michal-stanek 
what if we put unsafe {} inside init_limits?  and change `pub unsafe fn init_limits` -> `pub fn init_limits`?
maybe we could put there more functionalities in future, which can be also "safe",
something like this: https://doc.rust-lang.org/book/ch19-01-unsafe-rust.html#creating-a-safe-abstraction-over-unsafe-code

light_node/Cargo.toml
```diff
@@ -8,11 +8,13 @@ default-run = "light-node"
 [dependencies]
 clap = "2.33"
 dirs = "3.0"
+libc = "0.2.76"
```
bkontur, line 11: @michal-stanek 
Do we need this dependency?

light_node/src/main.rs
bkontur: @michal-stanek 
ok, cool, looks much nicer,

please, remove that db_cfg.max_open_files, if we dont need to set it.
or what if we set db_cfg.max_open_files as resolved rlimit from set_file_desc_limit?
that we could avoid rocksdb too_many_open_files error?

michal-stanek: I'd remove it altogether. If we set max_open_files to -1 (recommended setting with respect to RocksDB performance) this means unlimited, but I wouldn't expect RocksDB to ever get any close to 64k open files. The RocksDB stats monitoring which we will have in the future will also include checks for how many files are open. I also expect that the other bounds which we will be setting in RocksDB (LSM file sizes, etc) will indirectly limit the number of files open to some 'small' amount - like not more than a few hundred. We will have tight control over those values hopefully :)

This also depends on whether we can get the other JIRA issue resolved (panic and crash due to too many open files). RocksDB should be handling this error correctly and adjusting (should consolidate files then), but the crash we experienced is probably due to some other reason - I guess it's not handled properly in some other place. But if it turns out that we cannot find where it happens and fix it, then we can limit db_max_open_files to <64k as a workaround.

Also bear in mind the rlimit is for all file descriptors so that also includes sockets, etc.

michal-stanek: I did some more reading (web + internal RocksDB code) and now I'm even more convinced we should keep the default RocksDB setting for max_open_files. If in any situation we would want to have a small number of files generated by RocksDB, we can adjust size multipliers for SST files.
I updated the PR.

light_node/src/main.rs
michal-stanek: Good idea, done.

light_node/Cargo.toml
michal-stanek: missed it, thanks!

light_node/src/main.rs
bkontur: ok, cool, thank you

PR #273, opened 2020-08-26
Feature/optimize decoding

rspendl:
Strings in Field and Tag data structures are replaced with FieldName and TagVariant enums.

Benchmark was added to messages/benches so the effect of optimizations can be measured.

Minor changes to encoding::binary_reader::BinaryReader::decode_value were made to improve speed of decoding List, Hash and Byte encodings.


bkontur:
@rspendl 
Are you rebased on the actual origin/develop?


@tizoc 
Because this CI https://travis-ci.com/github/simplestaking/tezedge/jobs/378175968 failed, and it is supposed to be fixed in develop.

tezos/encoding/src/de.rs
```diff
@@ -397,7 +397,7 @@ impl<'de> de::MapAccess<'de> for StructDeserializer<'de> {
                 let (ref field, ref value) = *item;
                 self.value = Some(value);
                 seed.deserialize(StringDeserializer {
-                    input: field.clone(),
+                    input: String::from(field.as_str()), // TODO: check if custom deserializer for FieldName is appropriate
```
bkontur, line 400: @rspendl 
what about this TODO?

tezos/encoding/src/encoding.rs
```diff
@@ -6,21 +6,585 @@
...
+            "SwapRequest" => Ok(TagVariant::SwapRequest),
+            "TooManyConnections" => Ok(TagVariant::TooManyConnections),
+            "UnknownChainName" => Ok(TagVariant::UnknownChainName),
+            _      => panic!(input.to_owned()),//Err(()),
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Copy)]
+pub enum FieldName {
```
bkontur, line 157: @rspendl @jurajselep 
I also investigated this library for optimizations few weeks ago :). I did some changes.
And also I stared at `Field(name: String)`, `Tag(varian: String)`, that they cause lots of cloning everywhere (ser.rs/de.rs). I tried to do it like `Field(name: &static str)`  but than you need to introduce lifetimes and so on and it needed more refactoring and I didnt have much time so I reverted it :)

Do you have any comparision results before/after? Or some percentage how this improve things?

Well, it looks ok, I also like your benches, they look cool and complex, great, and they will be usefull.
But there is one thing which I am not sure and concerned about.
Library `tezos/messages` is equal implementation of tezos ocaml messages/structs - definete set of structs.
Library `tezos/encoding` is supposed to be general rust library for encoding/decoding like tezos ocaml  - anybody can use it for anything. In the near future, we also plan to compile it with wasm(?). Or maybe we would like to  upload it crates.io.

But when we introduce here enums like FieldName and TagVariant in `tezos/encoding` we definitelly lose generalization.

I dont see real use-case now, we are at the begining.
But lets imagine, I have comercial project out-of the Tezedge. I want to write custom encoding for structs, which are not `tezos/messages` related. I want to store encoded struct as bytes in some smart contract. I want to have own names for field or tag variants, because I also serve this data as json some where.
So, with enums FieldName/TagVariant in `tezos/encoding` it would be impossible to do so.


So, is there any other **efficent** and **general** way for `Field(name: String)`, `Tag(varian: String)` ?
I think about `&static str` or maybe introduce some `trait` for names, whatever I dont know.

If not, I would prefer in this case to be **general** and **reusable** over performance.

tezos/encoding/src/de.rs
rspendl: Oops, this is a leftover comment, no need for custom deserializer. I'll delete that.

tezos/encoding/src/encoding.rs
rspendl: @bkontur , @jurajselep When I was working on master branch, the measured performance of the master branch vs. the new updates and FieldName + TagVariant yielded about 35-45% speed improvement of decoding. I've just tested the current develop branch and this branch already decodes about 80% faster and FieldName not only does not help much but it even slows down decoding, probably due to a lot of cloning.

I'll check today what were the differences between master and develop and see what's the reason for this. Possibly some other optimization has rendered enums obsolete, though they should be faster.

For using general strings for field names, maybe we can use some combination of enums for the most commonly used field names and something like CustomString(s: &str) so we associate a string with the enum for general solution. It will definitely use more memory, but with &str it's probably just 8 bytes for the pointer. How much would the performance be affected by this "just 8 bytes" is something I'd investigate.

So, let me come with an improved PR (the last one was refactored to match develop instead of the master without much testing of performance) so we can decide how to move forward.

rspendl:
I've checked the build, last night it succeed on Linux but failed on macOS, but now it's building OK on both. macOS has complained that it can't open a socket in ffi test, might have been just a network glitch.

tezos/encoding/src/encoding.rs
bkontur: @rspendl
"I've just tested the current develop branch and this branch already decodes about 80% faster".
Sorry, just to be clear :) What do you mean "this branch" ?

1. develop is 80% faster than master?
2. feature/optimize-decoding is 80% faster than develop?
3. feature/optimize-decoding is 80% faster than master?

rspendl: Sorry for vague wording: develop is 80% faster at decoding the stream of binary messages than master. feature/optimize-decoding was about 40% faster than master, but when based on develop, it's now about 50% slower than develop.
I'll look into lazy_static encodings, I'm looking into another peculiar thing that calling BinaryReader::read is 90% slower than calling BinaryReader::decode_value (ok, it's a private function, but just for test). Might be that lazy_static is the cause for a huge performance difference even though ::read is just calling decode_value and returns a Result.

rspendl:
After a lot of testing and trials to refactor calls, introducing FieldName enum instead of String names only provides about 5% of speed-up. Adding an enum variant with added String type (like CustomName(&str)) introduces lifetimes that are propagated practically to the whole codebase, so such solution is highly impractical for such minor improvement.

So I'm closing this pull request and opening another where only the benchmarks and an optimization of reading Encoding::List with multiple Uint8 values are included.

PR #275, opened 2020-08-31
Feature/optimized list

rspendl:
PR with added benchmarks for decoding binary messages and an optimized decoding of Encoding::List with UInt8 values. Optimization results in ~10% improvement of binary message reading speed.


bkontur:
@rspendl @jurajselep 
thank you,
we finally cherry-pick those benchmark commits and merge it to our develop here:
https://github.com/simplestaking/tezedge/pull/334

PR #277, opened 2020-09-01
TE-214: Call tezos-ffi with OCaml records instead of encoded messages

tizoc:
Ready for review.

TODO:
- [x] test more and verify it is working
- [x] cleanup (all conversion is now implemented in tezos/api instead of spread across multiple crates)
- [x] cleanup even more
- [x] benchmark
- [x] ~automate rust<->ocaml record mapping~ (didn't automate but made things more declarative)
- [x] remove remaining encoders that will not be needed anymore along with the testing/benchmarking code that depends on that
- [ ] make sure that every rust<->ocaml conversion is covered by tests (and if possible, implement specific tests for that)
- [x] release updated tezos-ffi


tizoc:
Note: I had to move the "tezos_messages" dependency from "dev" to regular dependencies, but I guess the functions that use it are functions that are only used for testing, so I guess I have to tag them as such.

tizoc:
Preliminary results:

```Benchmarking apply_block_request_decoded_roundtrip: Collecting 100 samples in e                                                                               apply_block_request_decoded_roundtrip                        
                        time:   [27.658 us 28.170 us 28.789 us]
                        change: [-2.0054% -0.1538% +2.0043%] (p = 0.88 > 0.05)
                        No change in performance detected.
Found 14 outliers among 100 measurements (14.00%)
  4 (4.00%) high mild
  10 (10.00%) high severe
Benchmarking apply_block_request_encoded_roundtrip: Collecting 100 samples in e                                                                               apply_block_request_encoded_roundtrip                        
                        time:   [61.899 us 62.947 us 64.262 us]
                        change: [-15.632% -10.960% -6.4985%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 4 outliers among 100 measurements (4.00%)
  2 (2.00%) high mild
  2 (2.00%) high severe```

Will see if I can implement automatic rust<->ocaml enums/records conversion next.

File: tezos/api/src/ocaml_conv/from_ocaml.rs [hunk 1]
```diff
@@ -0,0 +1,174 @@
...
+        } else {
+            match v.tag_value() {
+                0 => {
+                    let path: FfiPath = unsafe { v.field::<Path>(0).into_rust() };
+                    let right = unsafe { v.field::<OCamlHash>(1).into_rust() };
+
+                    FfiPath(Path::Left(Box::new(PathLeft::new(
+                        path.0,
+                        right,
+                        Default::default(), // TODO: what is body?
```
tizoc, line 157: @bkontur do you know that the `body` fields in `PathLeft` and `PathRight` are?

tezos/api/src/ocaml_conv/from_ocaml.rs
```diff
@@ -0,0 +1,194 @@
...
+            )+
+
+            $cons {
+                $($field),+
+            }
+        }
+    };
+}
+
+unsafe impl FromOCaml<ForkingTestchainData> for ForkingTestchainData {
```
bkontur, line 53: @tizoc 
What do you think about to put:
```
unsafe impl FromOCaml<ForkingTestchainData> for ForkingTestchainData {
    fn from_ocaml(v: OCaml<ForkingTestchainData>) -> Self {
```
also inside the macro unpack_ocaml_block ?

tizoc: Sounds good, I like that.

tezos/api/src/ocaml_conv/from_ocaml.rs
bkontur: @tizoc 
Do you mean BinaryDataCache::default() vs Default::default()?
I see that we use lots of `body: Default::default(),` in tezos/messages module,
so I guess `Default::default()` is ok here

```
#[derive(Clone, Serialize, Deserialize, PartialEq, Debug, Getters)]
pub struct PathLeft {
    #[get = "pub"]
    path: Path,
    #[get = "pub"]
    right: Hash,
    #[serde(skip_serializing)]
    body: BinaryDataCache,
}
```

tezos/api/src/ocaml_conv/from_ocaml.rs
```diff
@@ -0,0 +1,194 @@
...
+                    let left = unsafe { v.field::<OCamlHash>(0).into_rust() };
+                    let path: FfiPath = unsafe { v.field::<Path>(1).into_rust() };
+
+                    FfiPath(Path::Right(Box::new(PathRight::new(
+                        left,
+                        path.0,
+                        Default::default(), // TODO: what is body?
+                    ))))
+                }
+                tag => panic!("Invalid tag value for OCaml<Path>: {}", tag),
```
bkontur, line 190: @tizoc 
I am not sure about this panic, this could cause unexpected behavior, it will probably just crash protocol_runner subprocess, but there is also rust trait `impl TryFrom` which return Result for this cases

tezos/api/src/ocaml_conv/from_ocaml.rs
tizoc: My question is more about what that `body` field is, and if it is something I should care about when performing this mapping. I used `Default::default()` as a placedholder because I needed something in there, but I guess it also works as the correct value to put in there.

tezos/api/src/ocaml_conv/to_ocaml.rs
```diff
@@ -0,0 +1,252 @@
...
+    // Hashes
+
+    alloc fn alloc_operation_list_list_hash(hash: OCamlBytes) -> OCamlOperationListListHash;
+    alloc fn alloc_operation_hash(hash: OCamlBytes) -> OCamlOperationHash;
+    alloc fn alloc_block_hash(hash: OCamlBytes) -> OCamlBlockHash;
+    alloc fn alloc_context_hash(hash: OCamlBytes) -> OCamlContextHash;
+    alloc fn alloc_protocol_hash(hash: OCamlBytes) -> OCamlProtocolHash;
+}
+
+unsafe impl ToOCaml<ApplyBlockRequest> for ApplyBlockRequest {
```
bkontur, line 104: @tizoc 
Do we have all this converions/roundtrips catched in test_bytes_roundtrips.rs? 
It would be very usefull to have all this conversions catched in some test to be sure it works.
We would like to catch any incompatibility between structs (rust vs ocaml)  at least at tests.
What would you recommend?

tezos/api/src/ocaml_conv/from_ocaml.rs
tizoc: In theory this conversion will never fail, *unless* tezos changes the representation for its `Path` values or memory gets corrupted somehow (which then would crash things anyway at some other place).

tezos/api/src/ocaml_conv/to_ocaml.rs
tizoc: I think most are either directly or indirectly being converted during the tests (same as when the values where being encoded/decoded), but let me doublecheck.

When I asked some time ago how often these structs changed it was because I was worrying about this. I agree that it should be validated somehow in tests, but I haven't figured out the best way to do it yet. I will think about it more.

tezos/api/src/ocaml_conv/from_ocaml.rs
bkontur: it is used for kind of cache of encoded/decoded data,
Default::default() is ok, so please, just remove that TODO comment

tezos/api/src/ocaml_conv/from_ocaml.rs
bkontur: ok, ok, I understand,
so please, just add here some comment, that this `tag =>` is reached only if tezos changes internally something,
maybe we will come back here later when we will check clippy or something like that, so that comment would be usefull

tizoc: Ok, will do. At some point I will come up with a good safe abstraction for mapping between Rust<->OCaml structs and enums without the need for so much manually written code, and then this will not be needed anymore.

tezos/api/src/ocaml_conv/to_ocaml.rs
bkontur: I agree, they are tested indirectly in other tests.
This is important part "border between rust/ocaml", so I think it deserves direct dedicated tests :)
I am ok, if just test_bytes_roundtrips.rs covers that

tezos/api/src/ocaml_conv/from_ocaml.rs
tizoc: Done by adding a new macro, looks much better now: https://github.com/simplestaking/tezedge/pull/277/commits/6d1c4d2cfdb1adef968c69ba4a2cbfa91831e76e

bkontur: cool, yes, yes, much better, looks like more declarative style like encodings :)

tizoc:
@bkontur `as_rust_bytes` is not used anywhere anymore, and `from_rust_bytes` only shows up 3 times in `actors_apply_blocks_test.rs` for decoding `ApplyBlockRequest` values.

So just to confirm, any encoder that is not being used right now can be removed?

tezos/api/src/ocaml_conv/to_ocaml.rs
```diff
@@ -0,0 +1,297 @@
...
+                &FfiBlockHeader(self.block_header.clone()),
+                &FfiBlockHeader(self.pred_header.clone()),
+                &self.max_operations_ttl,
+                &operations,
+            )
+        }
+    }
+}
+
+unsafe impl ToOCaml<ApplyBlockResponse> for ApplyBlockResponse {
```
bkontur, line 142: @tizoc 
Bruno, could we refactor this also with macro `to_from_ocaml_block_mapping` like we have `impl_from_ocaml_block_mapping`, than it will look like the same thing

tezos/api/src/ocaml_conv/to_ocaml.rs
```diff
@@ -0,0 +1,297 @@
...
+                &self.max_operations_ttl,
+                &self.last_allowed_fork_level,
+                &self.forking_testchain,
+                &self.forking_testchain_data,
+            )
+        }
+    }
+}
+
+unsafe impl ToOCaml<ForkingTestchainData> for ForkingTestchainData {
```
bkontur, line 161: @tizoc 
if there is inside-block wrapped with unsafe, do we need also `unsafe impl` ?

tezos/api/src/ocaml_conv/to_ocaml.rs
bkontur: maybe after that, we could join `impl_from_ocaml_block_mapping` and `impl_to_ocaml_block_mapping` behind one macro like `impl_ocaml_block_mapping` where `from` and `to` conversion could be on one place,
and also maybe than we could generate inside `impl_ocaml_block_mapping` some test for conversion from/to automatically, so if we add new structs it will need some test_roundrobin stuff on ocaml side

but this is just my thought, lets just refactor it to `impl_to_ocaml_block_mapping` at first

tizoc: It would work for very simple cases like this one, where nothing has to be done with the fields, and there are no "FFI wrapper" structs involved, but for all others I just think it would make things less readable. The case above for example already needs to perform extra work, and is not a direct mapping of fields. I don't think it adds as much value here as it did for the implementation of the `FromOCaml` trait.

Regarding the tests, I thought about this a while yesterday, and I think I will just have to write all that manually, couldn't figure a good way to automate it, at least not with what we have right now. My plan is to write some functions on the OCaml side that takes one argument for each field of the structs, and construct an instance of that struct (in OCaml). In addition to those fields, as it's first argument this function would receive an instance of that struct, but constructed by Rust, and then it would compare these two structs structurally to verify that they are equal.

I will start working on that once everything else related to this PR is solved.

tezos/api/src/ocaml_conv/to_ocaml.rs
```diff
@@ -0,0 +1,297 @@
...
+        let operations: Vec<Vec<FfiOperation>> = self
+            .operations
+            .iter()
+            .map(|ops| ops.iter().map(|op| FfiOperation(op.clone())).collect())
+            .collect();
+        unsafe {
+            alloc_apply_block_request(
+                token,
+                &self.chain_id,
+                &FfiBlockHeader(self.block_header.clone()),
```
bkontur, line 133: @tizoc 
and last thing is there anyway to get rid of this clones?

PR #278, opened 2020-09-08
Te-74 Generate identity in rust tezedge + check PoW

vlad9486:


crypto/src/proof_of_work.rs
```diff
@@ -0,0 +1,152 @@
+use num_bigint::BigUint;
+use hex::{FromHex, FromHexError};
+use std::convert::TryFrom;
+use sodiumoxide::randombytes::randombytes;
+use super::{blake2b, crypto_box::PublicKey};
+
+#[derive(Clone, PartialEq)]
+pub struct ProofOfWork([u8; 0x18]);
```
bkontur, line 8: @vlad9486 
PR looks cool :),

there are several usage of 0x18 / 0x20, what if you introduce those as constants?
I dont know, if it will be readable than?

at least, some short comments/description should here, where you mention 0x18 / 0x20, that they are used as indexes or "number of bytes" or something like that

crypto/src/proof_of_work.rs
```diff
@@ -0,0 +1,152 @@
...
+                    data[0x28..0x30].clone_from_slice(b.to_be_bytes().as_ref());
+                } else {
+                    c += 1;
+                }
+                data[0x30..0x38].clone_from_slice(c.to_be_bytes().as_ref());
+            }
+        }
+    }
+
+    pub fn check(&self, pk: &[u8; 0x20], target: f64) -> Result<(), ()> {
```
bkontur, line 73: @vlad9486 
do you use this somewhere?
if not, please remove, or remove just pub,

if anybody wants to use it, he would wonder about why `&[u8; 0x20]`, because probably he will have just some PubKey or Vec<u8>

PR #281, opened 2020-09-16
Introduce merkle storage for context

michal-stanek:
Import, enhance and integrate Merkle-tree-like (git-like) data structure as context storage.
This replaces the previous skip-list implementation.

The new storage is able to handle conflicts at specific levels and is also more space-efficient.


light_node/src/main.rs
```diff
@@ -284,6 +284,9 @@ fn block_on_actors(
         drop(tezos_readonly_api);
         debug!(log, "Shutdown tezos_readonly_api complete");
 
+        persistent_storage.flush_dbs();
+        info!(log, "Flushing databases");
```
bkontur, line 288: @michal-stanek 
I think this log should go before `persistent_storage.flush_dbs();` :)

storage/src/persistent/database.rs
```diff
@@ -148,25 +142,6 @@ impl<S: KeyValueSchema> KeyValueStoreWithSchema<S> for DB {
 
         Ok(IteratorWithSchema(self.prefix_iterator_cf(cf, key), PhantomData))
     }
-
-    fn contains(&self, key: &S::Key) -> Result<bool, DBError> {
```
bkontur, line 152: @michal-stanek 
I would leave `contains` here, and just implement it with get here, something like:
```
fn contains(&self, key: &S::Key) -> Result<bool, DBError> {
    get(key).is_some()
}
```

maybe in the future, there will be better rocksdb way to do it than just with get,
and also it will be much readable to have distinction where do you need `get` or just `contains`, 
maybe get touches data and contains not

michal-stanek: ok, I'll wrap it like that

storage/src/context.rs
```diff
@@ -1,218 +1,73 @@
...
-    pub fn new(level: Option<usize>, context_hash: Option<ContextHash>) -> Self {
-        ContextIndex { level, context_hash }
-    }
+    // set key-value
+    fn set(&mut self, context_hash: &Option<ContextHash>, key: &[String], value: &[u8]) -> Result<(), ContextError>;
+    // checkout context for hash
+    fn checkout(&self, context_hash: &ContextHash) -> Result<(), ContextError>;
+    // commit current context diff to storage
+    // if parent_context_hash is empty, it means that it's a commit_genesis and we don't assign context_hash to header
+    fn commit(&mut self, block_hash: &BlockHash, parent_context_hash: &Option<ContextHash>,
```
bkontur, line 22: @michal-stanek 
I am thinking about to clean this function `commit`,
I would remove arg `new_context_hash: &ContextHash`,

and change `Result<()` -> `Result<ContextHash`
and put that assert_eq in caller (context_listener)


and aslo we will later remove `self.block_storage.assign_to_context(` need to check other things before

storage/src/merkle_storage.rs
```diff
@@ -0,0 +1,937 @@
...
+            }
+            Some(entry) => Ok(entry.clone()),
+        }
+    }
+
+    fn get_non_leaf(&self, hash: EntryHash) -> Node {
+        Node { node_kind: NodeKind::NonLeaf, entry_hash: hash }
+    }
+
+    fn get_db<P: AsRef<Path>>(path: P, cache: &Cache) -> DB {
```
bkontur, line 628: @michal-stanek 
can you please check this?
http://ci.tezedge.com/simplestaking/tezedge/1597/1/3
image

storage/src/context.rs
```diff
@@ -1,218 +1,73 @@
...
+              new_context_hash: &ContextHash, author: String, message: String,
+              date: i64) -> Result<(), ContextError>;
+    fn delete_to_diff(&self, context_hash: &Option<ContextHash>, key_prefix_to_delete: &[String]) -> Result<(), ContextError>;
+    fn remove_recursively_to_diff(&self, context_hash: &Option<ContextHash>, key_prefix_to_remove: &[String]) -> Result<(), ContextError>;
+    // copies subtree under 'from_key' to new subtree under 'to_key'
+    fn copy_to_diff(&self, context_hash: &Option<ContextHash>, from_key: &[String], to_key: &[String]) -> Result<(), ContextError>;
+    // get value for key
+    fn get_key(&self, key: &Vec<String>) -> Result<Vec<u8>, ContextError>;
```
bkontur, line 30: @michal-stanek 
what is the best way?
`key: &Vec<String>` vs `key: &[String]` vs `prefix: &ContextKey`

I think this is a trait/api and we should align this somehow

michal-stanek: I think &ContextKey would be best for both key and prefix. I will change it.

rpc/src/services/base_services.rs
```diff
@@ -239,18 +236,10 @@ pub(crate) fn get_cycle_from_context(block_id: &str, list: ContextList, persiste
...
-
-    // get cylce list from context storage
-    // let cycle_lists: HashMap<String, Bucket<Vec<u8>>> = context_data.clone().into_iter()
-    //     .filter(|(k, _)| k.contains("cycle"))
-    //     .filter(|(_, v)| match v {
-    //         Bucket::Exists(_) => true,
-    //         _ => false
-    //     })
-    //     .collect();
+    let context = TezedgeContext::new(BlockStorage::new(&persistent_storage), persistent_storage.merkle());
```
bkontur, line 239: @michal-stanek 
I am just thinking, do we need to make instance of TezedgeContext all the time?
Couldnt we just store reference in RpcServiceEnvironment like persistent_storage is?

storage/src/persistent/mod.rs
```diff
@@ -158,12 +157,12 @@ impl PersistentStorage {
     }
 
     #[inline]
-    pub fn context_storage(&self) -> ContextList { self.cs.clone() }
-}
+    pub fn merkle(&self) -> Arc<RwLock<MerkleStorage>> {
+        self.merkle.clone()
+    }
 
-impl Drop for PersistentStorage {
```
bkontur, line 164: @michal-stanek 
I dont remember, what was the problem with this drop?
Without that, databases want be flushed in tests or if light-node ends (panic) without ctrl-c, than flush wont be triggered, 
I will wrap it, and all tests worked
```
impl Drop for PersistentStorage {
    fn drop(&mut self) {
        self.flush_dbs();
    }
}
```

bkontur: another problem is that if self.clog.flush() fails, than self.kv.flush() is also not triggered

storage/src/context.rs
```diff
@@ -1,218 +1,73 @@
...
-}
 
-/// Actual context implementation with context skip list
-pub struct TezedgeContext {
-    block_storage: BlockStorage,
-    storage: ContextList,
-}
+    fn checkout(&self, context_hash: &ContextHash) -> Result<(), ContextError> {
+        let mut merkle = self.merkle.write().expect("lock poisoning");
+        let context_hash_arr: EntryHash = context_hash.as_slice().try_into().expect("EntryHash conversion error");
```
bkontur, line 55: @michal-stanek 
`.expect("EntryHash conversion error")` should be better map to ContextError,
I saw `.expect("EntryHash conversion error")` in several places

storage/src/persistent/mod.rs
michal-stanek: I remember context tests used to crash because of Drop for PersistentStorage in the old implementation. If the tests work now then it's fine.
Yeah we should try flushing both, I'll get rid of expect()

rpc/src/services/base_services.rs
michal-stanek: Everything under TezedgeContext was Arc<> so construction wasn't expensive but I changed it to use a reference anyway so that it looks a bit nicer in base_services.

storage/src/persistent/mod.rs
michal-stanek: Fixed in the new version

storage/src/context.rs
michal-stanek: Fixed

storage/src/context.rs
```diff
@@ -153,10 +153,24 @@ impl ContextApi for TezedgeContext {
...
+    fn get_context_tree_by_prefix(&self, context_hash: &ContextHash, prefix: &Option<&str>) -> Result<StringTree, MerkleError> {
+        let merkle = self.merkle.read().expect("lock poisoning");
+        // clients may pass in a prefix with elements containing slashes (expecting us to split)
+        // we need to join with '/' and split again
+        let mut prefix: ContextKey = match prefix {
+            Some(prefix) => prefix.split('/').map(|s| s.to_string()).collect(),
+            None => vec![],
+        };
+        // clients assume that root is at /data
+        prefix.insert(0, "data".to_string());
```
bkontur, line 169: @michal-stanek 
I moved this prefix parsing and "data" to the client call, so it could be used to get another key prefixes : test_chain, protocol or other roots in future (if any)


**rpc -> base_services.rs:**
```
    // we assume that root is at "/data"
    let mut key_prefix = vec!["data".to_string()];

    // clients may pass in a prefix (without /data) with elements containing slashes (expecting us to split)
    // we need to join with '/' and split again
    if let Some(prefix) = prefix {
        key_prefix.extend(prefix.split('/').map(|s| s.to_string()));
    };

    match context.get_context_tree_by_prefix(&ctx_hash.unwrap(), &key_prefix) {
```

storage/src/context.rs
```diff
@@ -52,32 +54,30 @@ impl ContextApi for TezedgeContext {
 
     fn checkout(&self, context_hash: &ContextHash) -> Result<(), ContextError> {
         let mut merkle = self.merkle.write().expect("lock poisoning");
```
bkontur, line 56: @michal-stanek 
I also changed order:
```
let mut merkle = self.merkle.write().expect("lock poisoning");
let context_hash_arr: EntryHash = context_hash.as_slice().try_into()?;
```
to 
```
let context_hash_arr: EntryHash = context_hash.as_slice().try_into()?;
let mut merkle = self.merkle.write().expect("lock poisoning");
```

there is no need to do locking, if hash is bad,
but I think this is just a very very very very corner case :)

PR #285, opened 2020-09-18
Add tests for Rust<->OCaml value conversions

tizoc:


tizoc:
@bkontur I will not add tests for the `FromOCaml` conversions for now, because those are far more trivial and less likely to break. I will focus on the ocaml-interop improvements + release, once that is done I will then take advantage of such improvements on tezedge and see if I can solve those tests in a better way.

Btw this requires a new tezos-ffi release (and merge of !13)

PR #314, opened 2020-11-04
Add tezos python test framework tests to CI

adonagy:
Modified the node accordingly:

- add RPC to describe all the RPCs
- always connect all peers defined by the --peers argument
- actually disable the mempool when --disable_mempool true argument is present
- fix few rpc return values
- add a context rpc: /context/contracts/<contract>/balance

Few things before merge:

- [x] analyze the failing test in CI, fix accordingly
- [x] perform a last cleanup of the code
- [x] sign all somehow unsigned commits

**TODO:Code**
- [x] ~~@adonagy - rename to `synchronisation-threshold` (+ add it to sandbox cfg files?)~~
- [x] @adonagy - refactor `impl Stream for MonitorHeadStream` to new file `services/stream_services.rs`
-- split everything to: `HeadMonitorStream` and `OperationMonitorStream`  (then we dont need enum MonitorRpc)
-- put stuff from `helpers.rs` and `base_services` directly to `services/stream_services.rs`
- [x] @bkontur - refactor `CurrentMempoolState` and remove msg: `MempoolStateChanged(Arc<RwLock<CurrentMempoolState>>)`
- [x] @adonagy  - router.rs - multiple methods `// TODO: rework this, this should accept more methods`
-- at least for protocol we need: `GET, POST, OPTIONS, PUT`
- [x] @bkontur - impl for `get_user_activated_upgrades` it is in tezos_env
- [x] @adonagy  - peer_manager.rs  initialize connections on startup
- [ ] @bkontur  - add rpc/shell to fmt
- [x] @bkontur - correct usage of disable_mempool (just for p2p CurrentHead/GetCurrentHead)
- [x] @adonagy - (do as last point) for `inject/block` `inject/operation` get chain id from query

**TODO:Jiras**
- [x] @bkontur  - speed up injected block distribution
- [x] @adonagy  - add async parameter to `inject_operation`

**TODO:Infra**
- [x] @adonagy  - remove binaries tezos-client / tezos-admin-client from tezedge
- [x] ? - remove all debugs/logs/printlns at the end


rpc/src/server/router.rs
```diff
@@ -14,73 +14,79 @@ pub(crate) fn create_routes(is_sandbox: bool) -> PathTree<Handler> {
...
         T: Fn(Request<Body>, Params, Query, RpcServiceEnvironment) -> F + Send + Sync + 'static,
         F: Future<Output=HResult> + Send + 'static
 {
-    fn handle(&mut self, path: &str, f: T) {
+    fn handle(&mut self, method: Method, path: &str, f: T) {
         self.insert(path, Arc::new(move |req, params, query, env| {
             Box::new(f(req, params, query, env))
         }));
+        self.insert(&format!("/describe{}", path), Arc::new(move |req, params, query, env| {
+            Box::new(shell_handler::describe(method.clone(), req, params, query, env))
```
bkontur, line 89: @adonagy 
I guess this is not necessery clone: `method.clone()`

rpc/src/server/router.rs
```diff
@@ -14,73 +14,79 @@ pub(crate) fn create_routes(is_sandbox: bool) -> PathTree<Handler> {
...
 
+    // DEPRECATED in ocaml but still used by python tests
+    routes.handle(Method::GET, "/network/version", shell_handler::node_version);
+    
     routes
 }
 
 trait Routes<Fut> {
-    fn handle(&mut self, path: &str, f: Fut);
+    fn handle(&mut self, method: Method, path: &str, f: Fut);
```
bkontur, line 76: @adonagy 
cool, I think we will need this method also for router

rpc/src/server/shell_handler.rs
```diff
@@ -263,4 +263,66 @@ pub async fn node_version(_: Request<Body>, _: Params, _: Query, env: RpcService
...
         env.log(),
     )
-}
\ No newline at end of file
+}
+
+// TODO: remove. This is a 'fake it till you make it' handler
+// Handler faking the describe routes in ocaml to be compatible with tezoses python test framework
+pub async fn describe(method: Method, req: Request<Body>, _: Params, _: Query, env: RpcServiceEnvironment) -> ServiceResult {
+    //let method = req.method();
```
bkontur, line 271: @adonagy remove please commented code

rpc/src/server/shell_handler.rs
```diff
@@ -263,4 +263,66 @@ pub async fn node_version(_: Request<Body>, _: Params, _: Query, env: RpcService
...
+    let path: Vec<String> = req.uri().path().split("/").skip(2).map(|v| v.to_string()).collect();
+
+    let service_fields = serde_json::json!({
+        "meth": method.as_str(),
+        "path": path,
+        "description": "Lorem ipsum dolor sit amet, consectetur adipiscing elit.",
```
bkontur, line 277: @adonagy :)
change this pls. to:
```
// TODO: remove. This is a 'fake it till you make it' handler
// Handler faking the describe routes in ocaml to be compatible with tezoses python test framework
```

rpc/src/server/shell_handler.rs
```diff
@@ -263,4 +263,66 @@ pub async fn node_version(_: Request<Body>, _: Params, _: Query, env: RpcService
         base_services::get_node_version(env.network_version()),
         env.log(),
     )
-}
\ No newline at end of file
+}
+
+// TODO: remove. This is a 'fake it till you make it' handler
+// Handler faking the describe routes in ocaml to be compatible with tezoses python test framework
```
bkontur, line 269: @adonagy 
function comments starts ///

rpc/src/services/base_services.rs
```diff
@@ -176,6 +179,27 @@ pub(crate) fn live_blocks(_chain_param: &str, block_param: &str, env: &RpcServic
     Ok(live_blocks)
 }
 
+#[derive(Serialize, Debug)]
+pub(crate) struct Prevalidators {
```
bkontur, line 183: @adonagy 
I guess, just one `Prevalidators` -> `Prevalidator`

rpc/src/services/protocol/mod.rs
```diff
@@ -424,6 +424,47 @@ pub(crate) fn proto_get_contract_manager_key(
...
+
+    // get protocol and constants
+    let context_proto_params = get_context_protocol_params(
+        block_id,
+        None,
+        persistent_storage,
+        state,
+    )?;
+
+    let context = TezedgeContext::new(BlockStorage::new(&persistent_storage), persistent_storage.merkle());
```
bkontur, line 442: @adonagy 
there is a shared reference for TezedgeContext in rpc module,
please use that instead of creating new one here,
and also please check another usages in this file

tezos/messages/src/protocol/proto_006/contract.rs
```diff
@@ -30,3 +30,22 @@ has_encoding!(Counter, COUNTER_ENCODING, {
...
 });
+
+#[derive(Serialize, Deserialize, Debug, Clone, Getters)]
+pub struct Balance {
+    #[get = "pub"]
+    balance: BigInt,
+}
+
+impl Balance {
+    pub fn to_string(&self) -> String {
```
bkontur, line 41: @adonagy 
this is againts clippy: https://rust-lang.github.io/rust-clippy/master/#inherent_to_string
I would rename that function to something like: `to_json_format`,
and I also see another usage here: `impl Counter {` 

and also please check other usage `fn to_string` and rename them specificcally

rpc/src/services/protocol/mod.rs
adonagy: I see, cool, will change the usages.

tezos/messages/src/protocol/proto_006/contract.rs
adonagy: Oh right, I had a hunch this was not entirely ok. Will do 👍

rpc/src/lib.rs
```diff
@@ -47,7 +47,7 @@ pub(crate) fn make_json_stream_response<T: futures::Stream<Item=Result<String, s
         .header(hyper::header::ACCESS_CONTROL_ALLOW_HEADERS, "Content-Type")
         .header(hyper::header::ACCESS_CONTROL_ALLOW_HEADERS, "content-type")
         .header(hyper::header::ACCESS_CONTROL_ALLOW_METHODS, "GET, POST, OPTIONS, PUT")
-        .header(hyper::header::TRANSFER_ENCODING, "chunked")
+        // .header(hyper::header::TRANSFER_ENCODING, "chunked")
```
bkontur, line 50: @adonagy 
do we need it or not?

rpc/src/server/router.rs
```diff
@@ -14,66 +14,80 @@ pub(crate) fn create_routes(is_sandbox: bool) -> PathTree<Handler> {
...
-    routes.handle("/stats/database_mem", dev_handler::database_memstats);
-    //routes.handle("/stats/storage", dev_handler::dev_stats_storage);
+    routes.handle(Method::GET, "/dev/chains/main/blocks", dev_handler::dev_blocks);
+    routes.handle(Method::GET, "/dev/chains/main/actions/blocks/:block_hash", dev_handler::dev_action_cursor);
+    routes.handle(Method::GET, "/dev/chains/main/actions/contracts/:contract_address", dev_handler::dev_action_cursor);
+    routes.handle(Method::GET, "/stats/memory", dev_handler::dev_stats_memory);
+    routes.handle(Method::GET, "/stats/database_mem", dev_handler::database_memstats);
+    //routes.handle(Method::GET, "/stats/storage", dev_handler::dev_stats_storage);
 
+    routes.handle(Method::GET, "config/network/user_activated_upgrades", shell_handler::config_user_activated_upgrades);
```
bkontur, line 68: @adonagy 
please move it after "/workers/prevalidators"

rpc/src/services/mempool_services.rs
```diff
@@ -263,6 +268,28 @@ pub fn inject_block(
     Ok(block_hash)
 }
 
+pub fn request_operations(
+    _data: &str,
```
bkontur, line 272: @adonagy 
remove please _data arg, if dont need it

light_node/src/configuration.rs
```diff
@@ -283,6 +283,12 @@ pub fn tezos_app() -> App<'static, 'static> {
             .value_name("NUM")
             .help("Maximal number of peers to connect to")
             .validator(parse_validator_fn!(usize, "Value must be a valid number")))
+        .arg(Arg::with_name("synchronization-thresh")
```
bkontur, line 286: @adonagy 
Ado, pls. rename it according to tezos: `synchronisation-threshold`

rpc/src/helpers.rs
```diff
@@ -125,62 +134,272 @@ pub struct BlockHeaderMonitorInfo {
...
+    pub branch: String,
+    pub contents: Value,
+
+    #[serde(skip_deserializing)]
+    pub protocol: Option<String>,
+    #[serde(skip_serializing)]
+    pub hash: String,
+}
+
 pub struct MonitorHeadStream {
```
bkontur, line 162: @adonagy 
Ado, we should to refactor this streaming a little bit - see description of this PR

rpc/src/server/router.rs
```diff
@@ -14,66 +14,80 @@ pub(crate) fn create_routes(is_sandbox: bool) -> PathTree<Handler> {
...
-    routes.handle("/chains/:chain_id/blocks/:block_id/helpers/endorsing_rights", protocol_handler::endorsing_rights);
-    routes.handle("/chains/:chain_id/blocks/:block_id/votes/listings", protocol_handler::votes_listings);
+    routes.handle(Method::GET, "/chains/:chain_id/blocks/:block_id/context/constants", protocol_handler::context_constants);
+    routes.handle(Method::GET, "/chains/:chain_id/blocks/:block_id/helpers/baking_rights", protocol_handler::baking_rights);
+    routes.handle(Method::GET, "/chains/:chain_id/blocks/:block_id/helpers/endorsing_rights", protocol_handler::endorsing_rights);
+    routes.handle(Method::GET, "/chains/:chain_id/blocks/:block_id/votes/listings", protocol_handler::votes_listings);
 
     // Other Protocol rpcs - routed through ffi calls
-    routes.handle("/chains/:chain_id/blocks/:block_id/*any", protocol_handler::call_protocol_rpc);
+    // TODO: rework this, this should accept more methods
```
bkontur, line 57: @adonagy 
refactor this methods, to support multiple, at least for protocol we need to support them

rpc/src/server/shell_handler.rs
```diff
@@ -67,9 +67,42 @@ pub async fn valid_blocks(_: Request<Body>, _: Params, _: Query, _: RpcServiceEn
...
+        applied: applied == Some("yes"),
+        branch_refused: branch_refused == Some("yes"),
+        branch_delayed: branch_delayed == Some("yes"),
+        refused: refused == Some("yes"),
+    };
+
+    make_json_stream_response(base_services::get_operations_monitor(&chain_id, &env, Some(mempool_query))?.unwrap())
+}
+
+pub async fn blocks(_: Request<Body>, params: Params, query: Query, env: RpcServiceEnvironment) -> ServiceResult {
```
bkontur, line 94: @adonagy 
hmm, looks suspicious,
do you have stored some calls? I would like to see the whole rpc path with parameters

rpc/src/services/base_services.rs
```diff
@@ -143,6 +226,11 @@ pub(crate) fn get_block_by_block_id(chain_id: &ChainId, block_hash: &BlockHash,
     )
 }
 
+// TODO: This requires further investigation, for now, just mock an empty vector for python tests' sake
```
bkontur, line 229: @bkontur 
add impl for get_user_activated_upgrades

shell/src/chain_manager.rs
```diff
@@ -746,7 +759,34 @@ impl ChainManager {
...
+                                let ChainManager {
+                                    peers,
+                                    chain_state,
+                                    identity_peer_id,
+                                    ..
+                                } = self;
+                                let header: &BlockHeader = &message.header().header;
+                                let chain_id = chain_state.get_chain_id();
+
+                                let block_hash: BlockHash = header.message_hash()?;
```
bkontur, line 772: @adonagy 
you dont need to compute hash `header.message_hash()?` (it is probably cached there, maybe we remove that cache in future),
you have it here already: `message.header().hash.clone()`

shell/src/peer_manager.rs
```diff
@@ -336,6 +336,13 @@ impl Actor for PeerManager {
         });
     }
 
+    fn post_start(&mut self, ctx: &Context<Self::Msg>) {
+        if !self.initial_peers.is_empty() {
+            info!(ctx.system.log(), "Connecting to defined peers");
+            self.initial_peers.drain().for_each(|address| ctx.myself().tell(ConnectToPeer { address: address.clone() }, ctx.myself().into()));
```
bkontur, line 342: @adonagy 
hmm, drain will clear self.initial_peers, so later they wont be used on `discover_peers(`,
this is not good, we need to find betrer way to initialize connections on startup

networking/src/p2p/peer.rs
```diff
@@ -29,7 +29,7 @@ use super::network_channel::{NetworkChannelRef, NetworkChannelTopic, PeerBootstr
 use super::stream::{EncryptedMessageReader, EncryptedMessageWriter, MessageStream, StreamError};
 
 const IO_TIMEOUT: Duration = Duration::from_secs(6);
-const READ_TIMEOUT_LONG: Duration = Duration::from_secs(30);
+const READ_TIMEOUT_LONG: Duration = Duration::from_secs(120);
```
bkontur, line 32: @adonagy 
what is original tezos timeout for this?

rpc/src/services/base_services.rs
bkontur: done

rpc/src/services/protocol/proto_006/rights_service.rs
```diff
@@ -175,7 +175,12 @@ fn baking_rights_assign_rolls(parameters: &RightsParams, constants: &RightsConst
 
         // we omit the estimated_time field if the block on the requested level is already baked
         let priority_timestamp = if block_level < (level as i64) {
-            Some(estimated_head_timestamp + (priority as i64 * time_between_blocks[1]))
+            let time = if time_between_blocks.len() == 1 {
```
bkontur, line 178: @adonagy 
we should fix this in all protos

rpc/src/services/mempool_services.rs
```diff
@@ -141,10 +141,15 @@ pub fn inject_operation(
     let block_meta_storage: Box<dyn BlockMetaStorageReader> = Box::new(BlockMetaStorage::new(persistent_storage));
     let state = env.state();
 
+    let state = state.read().unwrap();
+
+    if state.disable_mempool() {
```
bkontur, line 146: @bkontur 
remove this

shell/src/peer_manager.rs
adonagy: @bkontur  Well, we could just iterate through `self.initial_peers` instead of drain. That would ensure the Set is not cleared. The idea behind drain was, that when a peer (from `self.initial_peers`) disconnects, or won't even connect there is no point keeping it in initial peers. Ideally when this happens, we would want to look for an another peer. As the name suggests, they are the initial peers (or "starting" peers)

rpc/src/services/protocol/proto_006/rights_service.rs
adonagy: Definitely. Will do

networking/src/p2p/peer.rs
adonagy: Hmm, good question. Tezos sends a `get_current_head` to a peer which is not responding for 90s. In tezedge, we schedule and send a `get_current_head` message every 90s. So I set the timeout 120s here so it has  30s (120 - 90) to respond.

shell/src/chain_manager.rs
adonagy: Oh right! :)

rpc/src/lib.rs
adonagy: We don't. It is the default configuration. No point setting it explicitly.

adonagy: But still, no idea why I left there a same, but commented line

adonagy:
@bkontur About `synchronization-thresh`. The name diverges from the tezos one to be consistent with our other two args (`--peer-thresh-high` `--peer-thresh-high`)

PR #317, opened 2020-11-06
New ffi version

bkontur:


storage/src/merkle_storage.rs
```diff
@@ -407,13 +407,13 @@ impl MerkleStorage {
         };
         let entry = Entry::Commit(new_commit.clone());
 
-        self.put_to_staging_area(&self.hash_commit(&new_commit)?, entry.clone());
```
bkontur, line 410: @michal-stanek 
Michal, just asking, was there any reason to calculate context_hash twice here?:
```
self.put_to_staging_area(&self.hash_commit(&new_commit)?, entry.clone());
..
let last_commit_hash = self.hash_commit(&new_commit)?;
```
I changed just to one computation

PR #318, opened 2020-11-10
Use new RPC router

tizoc:
Some notes:

- ocaml-interop updated to latest version
- error handling is the same as before (OK responses get their body encoded as json, everything else is an error)
- content-type header is not handled for now, json is assumed on the OCaml-side


tezos/api/src/ffi.rs
```diff
@@ -562,38 +562,88 @@ pub type Json = String;
...
+
+    fn try_from(s: &str) -> Result<Self, Self::Error> {
+        let upper = s.to_uppercase();
+        match upper.as_ref() {
+            "DELETE" => Ok(RpcMethod::DELETE),
+            "GET" => Ok(RpcMethod::GET),
+            "PATCH" => Ok(RpcMethod::PATCH),
+            "POST" => Ok(RpcMethod::POST),
+            "PUT" => Ok(RpcMethod::PUT),
+            _ => Err("Invalid RPC method"),
```
bkontur, line 607: @tizoc please, add `s`  to `Err("Invalid RPC method")` like
`Err(format!("Invalid RPC method: {}", s))`

tizoc: Done.

rpc/src/helpers.rs
```diff
@@ -569,11 +569,35 @@ pub(crate) fn current_time_timestamp() -> TimeStamp {
 
 pub(crate) async fn create_ffi_json_request(req: Request<Body>) -> Result<JsonRpcRequest, failure::Error> {
     let context_path = req.uri().path_and_query().unwrap().as_str().to_string();
+    let meth = RpcMethod::try_from(req.method().to_string().as_str()).unwrap(); // TODO: handle correctly
+    let content_type = None; // TODO: get from request
```
bkontur, line 573: @tizoc what about this content-type?

rpc/src/services/protocol/mod.rs
```diff
@@ -463,53 +463,61 @@ pub(crate) fn proto_get_contract_manager_key(
...
+fn handle_rpc_response(response: &ProtocolRpcResponse) -> Result<serde_json::value::Value, failure::Error> {
+    match response {
+        ProtocolRpcResponse::RPCOk(body) => Ok(serde_json::from_str(&body)?),
+        other => Err(failure::err_msg(format!("Got non-OK response from protocol-RPC service: {:?}", other))),
+    }
+}
+
 pub(crate) fn run_operation(chain_param: &str, block_param: &str, json_request: JsonRpcRequest, env: &RpcServiceEnvironment) -> Result<serde_json::value::Value, failure::Error> {
-    let request = create_protocol_json_rpc_request(chain_param, block_param, json_request, FfiRpcService::HelpersRunOperation, &env)?;
+    let request = create_protocol_json_rpc_request(chain_param, block_param, json_request, &env)?;
```
bkontur, line 475: @tizoc 
I think we could also allign this name:
`create_protocol_json_rpc_request` -> `create_protocol_rpc_request`
`call_protocol_json_rpc` -> `call_protocol_json_rpc`
like you did in ffi.rs

rpc/src/helpers.rs
tizoc: @bkontur right now it is not used, and on the OCaml side it defaults to JSON encoding when the value is `None`. There is something else missing I just noticed, the "Accept" header (so, content-type => what to parse the input as, "accept" => what output encoding to use).

bkontur:
this could be also a separate commit:
```
- here is a predefined request mappings:
https://github.com/simplestaking/tezedge/blob/master/rpc/src/server/router.rs#L13
- here is a handling of routing for predefined request mappings
https://github.com/simplestaking/tezedge/blob/master/rpc/src/server/mod.rs#L116

my idea for handling is:
1. first check request with "predefined request mappings", if matched any, than handle
- this is how it is working now
2. if not found in predefined and request starts with "/chains/:chain_id/blocks/:block_id/*" than trigger protocol rpc router

there is kind of "*any: feature - Michal used it and it looks it is working:
routes.handle("/chains/:chain_id/blocks/:block_id/context/raw/bytes", shell_handler::context_raw_bytes);
routes.handle("/chains/:chain_id/blocks/:block_id/context/raw/bytes/*any", shell_handler::context_raw_bytes);

so maybe it is just enought to change this:
// remove all this:
// Protocol rpcs - routed through ffi calls
routes.handle("/chains/:chain_id/blocks/:block_id/helpers/scripts/run_operation", protocol_handler::run_operation);
routes.handle("/chains/:chain_id/blocks/:block_id/helpers/forge/operations", protocol_handler::forge_operations);
routes.handle("/chains/:chain_id/blocks/:block_id/helpers/current_level", protocol_handler::current_level);
routes.handle("/chains/:chain_id/blocks/:block_id/minimal_valid_time", protocol_handler::minimal_valid_time);

// add just something like this:
// Protocol rpcs - routed through ffi calls
routes.handle("/chains/:chain_id/blocks/:block_id/*any", protocol_handler::call_protocol_rpc);
```

rpc/src/helpers.rs
tizoc: @bkontur rechecking the code, I now remember why that part was a `TODO`. Currently every function in there has a `serde_json::value::Value` return type, so to be able to make use of the content type specification headers, this would have to change too. I will update the code to pass the value from the headers, but for now that will be ignored.

tizoc:
@bkontur what about `context_contract` ? I see it in the "implemented" section, but it still goes through FFI-RPC:

https://github.com/simplestaking/tezedge/blob/master/rpc/src/services/protocol/mod.rs#L450

bkontur:
> 
> 
> @bkontur what about `context_contract` ? I see it in the "implemented" section, but it still goes through FFI-RPC:
> 
> https://github.com/simplestaking/tezedge/blob/master/rpc/src/services/protocol/mod.rs#L450

@tizoc Bruno, yes, good point, please remove it from router.rs, it should go with protocol router

and also check this please:
```
// TODO: TE-226 - uncomment after protocol rpc router is done
// test_rpc_compare_json(&format!("{}/{}/{}", "chains/main/blocks", level, "helpers/current_level")).await;
// test_rpc_compare_json(&format!("{}/{}/{}", "chains/main/blocks", level, "minimal_valid_time")).await;
```
please, remove TODO and uncomment this two lines, it should work now (befere that it was hardcoded just for 006 protocol, so now it should work, because in drone first blocks are on 005 protocol)

rpc/src/services/protocol/mod.rs
tizoc: Done, see https://github.com/simplestaking/tezedge/pull/318/commits/f4c7bf8ed5a710c1f2da75c33306c5e0ba532f87 . I also updated the few remaining cases in the OCaml side. Technically everything is still "json", because thats the only thing supported right now, but at least when things change the names will not have to be taken care about too.

rpc/src/helpers.rs
bkontur: ok, ok

tizoc: https://github.com/simplestaking/tezedge/pull/318/files#diff-b4be62d44950d767fd17e69305f99558119c01ab676c79b557041ee102faa398R570-R589 <- these values are ignored on the OCaml side for now, but the information is passed on now.

bkontur: ok, cool, I see you also renamed: JsonRpcRequest -> RpcRequest, nice :)

tizoc:
@bkontur what about this: https://github.com/simplestaking/tezedge/blob/develop/rpc/src/services/protocol/mod.rs#L555-L563

Anything else to do there or just get rid of the comment?

tizoc:
Tests just passed fine locally after uncommenting those two lines btw.

bkontur:
> 
> 
> @bkontur what about this: https://github.com/simplestaking/tezedge/blob/develop/rpc/src/services/protocol/mod.rs#L555-L563
> 
> Anything else to do there or just get rid of the comment?


@tizoc 
we should remove the whole stuff:
```
routes.handle("/chains/:chain_id/blocks/:block_id/context/raw/json/cycle/:cycle_id", protocol_handler::cycle);
```

bkontur:
@tizoc and the very last thing, please change, all the time we forget to update it:
```
const GIT_COMMIT_HASH: &str = "5b2fe841c65901fd657fada0c6cc3e9b0b057641";
```
to https://gitlab.com/simplestaking/tezos/-/commit/ef6c692d19b5222d09ece59dc424906f8ee44a3a
```
const GIT_COMMIT_HASH: &str = "ef6c692d19b5222d09ece59dc424906f8ee44a3a";
```

I dont know if anybody is using it, I dont, maybe we could remove it also with UPDATE_GIT
```
            // check we want to update git updates or just skip updates, because of development process and changes on ocaml side, which are not yet in git
            let update_git = env::var("UPDATE_GIT").unwrap_or_else(|_| "true".to_string()).parse::<bool>().unwrap();
            if update_git {
                update_git_repository();
            }
```

what do you think?

tizoc:
@bkontur lol I always disable that because the git update is never what I want to do. My vote is for just removing it.

tizoc:
@bkontur if somebody is doing some work that doesn't use the released binaries, it is because he is making changes to the tezos-ffi repo, and in that case the "working commit hash" is irrelevant, and doing a git update just gets in the way. That is exactly what happens to me every time I have to depend on the repository instead of the released binaries.

adonagy:
@bkontur @tizoc Yeah, never used the "UPDATE_GIT" before, I think it can be removed

tizoc:
Also as a simplification we can remove the `OCAML_BUILD_CHAIN="local"/"remote"` variable and just use a single variable containing the tezos path, if it is set, then it will be used, otherwise the binary will be fetched.

bkontur:
> 
> 
> @bkontur if somebody is doing some work that doesn't use the released binaries, it is because he is making changes to the tezos-ffi repo, and in that case the "working commit hash" is irrelevant, and doing a git update just gets in the way. That is exactly what happens to me every time I have to depend on the repository instead of the released binaries.

ok, cool, thats what I wanted to hear and confirm my doubts, I also use UPDATE_GIT=false when I work with ffi,
ok, so please remove it and also check README.md, thx

tizoc:
Ok, will do

bkontur:
> 
> 
> Also as a simplification we can remove the `OCAML_BUILD_CHAIN="local"/"remote"` variable and just use a single variable containing the tezos path, if it is set, then it will be used, otherwise the binary will be fetched.

you mean something like:
- if present TEZOS_BASE_DIR -> means local build with own tezos repo directory
- if NOT present TEZOS_BASE_DIR -> means download dependencies?

than I would like to also remove/change Makefile,
I ususally build tezos separatelly and just copy libtezos.a to `lib_tezos/artifacts`

hmm, lets simplify thing for us

tizoc:
@bkontur exactly. See https://github.com/simplestaking/tezedge/pull/318/commits/eb2de0f29baac87f34232be3005e3acd1e85c40c

I will look into removing the Makefile too.

tizoc:
Ok, I will replace the makefile with a simple copy in the build script + a check to see if the file is there, and if not, ask the user to build it (with instructions on how to do so, and how to install the build dependencies)

tizoc:
New version force-pushed here: https://github.com/simplestaking/tezedge/commit/be9d9f385c1cd836f4d0da6b68cb61f2de8aef4a

Doesn't try to build anything, if it doesn't find `$TEZOS_BASE_DIR/libtezos-ffi.so` it asks you to build it (and points to the `README.md` file which contains basic instructions). If it does, it just copies the file.

PR #320, opened 2020-11-18
Add GitHub actions

tizoc:
- [x] build and tests
- [x] cargo check step before actual build (or instead of), it is faster
- [x] cargo audit - tested and commented out for now (because of `ws` dependency)
- [x] clippy
- [x] rustfmt - tested but commented out for now (we need to rustfmt the current codebase first, because it will fail as it is and the produced diff is huge)
- [ ] ~switch to rust stable (should be possible now) - TE-171~ (`const_fn` still not in stable, cannot be done yet)
- [ ] ~https://github.com/cirrus-actions/rebase~ (I think it is better to leave this one for later, I'm still not sure about how safe it is, and since the hard rebases still require manual intervention I'm not sure it will be that useful. Still an interesting example of what can be done)
- [x] badges in README
- [x] on repo's config, remove travis as "required", add the github actions stuff


tizoc:
Added rustfmt and cargo audit, disabled for now.

Sec audit will run periodically (not as part of PRs). For now it seems it will not be able to post anything once enabled, and the logs will need to be checked, see https://github.com/actions-rs/audit-check#limitations

tizoc:
Example of badge for README (would be `test` instead of `build` here unless we rename `test.yml`):

```
![build](https://github.com/simplestaking/ocaml-interop/workflows/build/badge.svg?branch=master)
```

build

tizoc:
About the Rustfmt one, there will be some synchronization required there, because for it to stop failing the whole codebase has to be run through rustfmt and then committed. This is probably going to be a PITA when rebasing any unmerged branches because it is going to generate conflicts on any code that has been modified in that branch.

bkontur:
> 
> 
> Example of badge for README (would be `test` instead of `build` here unless we rename `test.yml`):
> 
> ```
> build
> ```
> 
> build
> 
> Edit: even easier, just saw that if you go to the workflow, and click on the "..." icon, there is an option to create the badge. Also the name is not the name of the file, it is the value of "name:" or the whole path to the workflow file from the root of the project)


hmm, I am not sure about `test.yml`, I would prefer `build.yml` or `build-and-test.yml`,
is there any "good practice" for naming this wfl files?

bkontur:
> 
> 
> About the Rustfmt one, there will be some synchronization required there, because for it to stop failing the whole codebase has to be run through rustfmt and then committed. This is probably going to be a PITA when rebasing any unmerged branches because it is going to generate conflicts on any code that has been modified in that branch.

uff, to re-format project is always ... :D, lots of people will like you much more after that :D 
but we will need to do it, maybe we can do this when there wont be any opened PR or maybe during the christmas,
or we can do it partially by modules, rpc/shell/storage they are changed all the time, but the rest no,

we can re-format at least, rigth before merge this PR
`ipc`
`fuzz`
`logging`
`protocol_runner`
`sandbox`
`tezos/api`
`tezos/client`
`tezos/context`
`tezos/identity`
`tezos/identity`
`tezos/interop`
`tezos/interop_callback`
`tezos/messages`
`tezos/wrapper`

tizoc:
> hmm, I am not sure about `test.yml`, I would prefer `build.yml` or `build-and-test.yml`,
> is there any "good practice" for naming this wfl files?

No idea, it is not that the names are terribly relevant. I think `build-and-test.yml` is fine.

re: format partially by modules, thats an excellent idea, lets do that.

tizoc:
Added badges to README, and left audit and fmt in their final state but disabled (can be enabled here: https://github.com/simplestaking/tezedge/actions )

tizoc:
Seems to work fine without the OCaml install, but I don't know what is up with clippy now, it is failing without doing anything.

tizoc:
Nevermind, it was failing in the same before, just not as quickly because of the OCaml instal step. Not sure why.

tizoc:
Ok, seems to be working now (switched to plain nightly), will squash all the github actions stuff into a single commit once I see the final results from the run.

bkontur:
> 
> 
> Ok, seems to be working now (switched to plain nightly), will squash all the github actions stuff into a single commit once I see the final results from the run.

ok, cool, very nice,
so, if it everything works, please, just remove travis and add build.yml actions as required instead (we will add clippy later and others) and you just merge it to develop, or I can do it also in the morning

tizoc:
Ok, disabled clippy, will wait until this run finishes successfuly and then update the settings + merge into develop

PR #325, opened 2020-11-25
Extended statistics in Merkle Storage

michal-stanek:
- [ ] check disk usage stats for test, if possible

Gather latency stats (avg/min/max) for each operation (action) in the context storage. 
Both global and per-path (first node under /data) statistics are collected.

E.g. output (printed at the end of actors test)

Checkout: avg: 98692ns, min: 85095ns, max: 184275ns, times: 1325
GetKeyFromHistory: avg: 154697ns, min: 122461ns, max: 196316ns, times: 3
Commit: avg: 9542351ns, min: 397378ns, max: 8158597828ns, times: 1325
Set: avg: 1434198ns, min: 60163ns, max: 167741880ns, times: 51996
Delete: avg: 1441716ns, min: 92412ns, max: 20147907ns, times: 883
CopyToDiff: avg: 2779932ns, min: 1359952ns, max: 4371206ns, times: 10

And per-path statistics (truncated):

block_priority:
	Set: avg: 296119ns, min: 157054ns, max: 555984ns, times: 1324
v1:
	Set: avg: 145658ns, min: 106391ns, max: 178293ns, times: 4
commitments:
	Set: avg: 1738113ns, min: 248130ns, max: 167741880ns, times: 30001
	Delete: avg: 18302312ns, min: 17355364ns, max: 20147907ns, times: 59

### DIFF
--- a/storage/src/context.rs
+++ b/storage/src/context.rs
@@ -128,7 +128,7 @@ impl ContextApi for TezedgeContext {
 
     fn get_key_from_history(&self, context_hash: &ContextHash, key: &ContextKey) -> Result<Option<ContextValue>, ContextError> {
         let context_hash_arr: EntryHash = context_hash.as_slice().try_into()?;
-        let merkle = self.merkle.read().expect("lock poisoning");
+        let mut merkle = self.merkle.write().expect("lock poisoning");
         match merkle.get_history(&context_hash_arr, key) {
             Err(MerkleError::ValueNotFound { key: _ }) => Ok(None),
             Err(MerkleError::EntryNotFound { hash: _ }) => {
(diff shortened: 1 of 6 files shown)


storage/src/context.rs
```diff
@@ -126,7 +126,7 @@ impl ContextApi for TezedgeContext {
 
     fn get_key_from_history(&self, context_hash: &ContextHash, key: &ContextKey) -> Result<Option<ContextValue>, ContextError> {
         let context_hash_arr: EntryHash = context_hash.as_slice().try_into()?;
-        let merkle = self.merkle.read().expect("lock poisoning");
+        let mut merkle = self.merkle.write().expect("lock poisoning");
```
bkontur, line 129: @michal-stanek 
why do we need `write lock` for just reading from history fixed data?

michal-stanek: I had to change all read locks to write because we're updating statistics on each operation.

PR #346, opened 2020-12-16
Feature/python test todos

adonagy:
TODO: switch base to feature/python-feature-compatibility after tests

- [x] add async parameter to inject_operation
- [x] for inject/block inject/operation get chain id from query: TODO: TE-221 - add optional chain_id to params mapping
- [x] peer_manager.rs initialize connections on startup
- [ ]  TE-210 - remove from storage, block after inject failed or mempool operation injetion failed thourght rpc


rpc/src/server/shell_handler.rs
```diff
@@ -224,9 +231,14 @@ pub async fn inject_block(req: Request<Body>, _: Params, query: Query, env: RpcS
 
     let shell_channel = env.shell_channel();
 
-    // TODO: TE-221 - add optional chain_id to params mapping
-    let chain_id_param = "main";
-    let chain_id = parse_chain_id(chain_id_param, &env)?;
+    // defaults to "main"
+    let chain_id_query = if let Some(chain_id_query) = query.get_str("chain_id") {
```
bkontur, line 235: @adonagy 
could be also written like this with constant:
```
let chain_id_query = query.get_str("chain_id").unwrap_or(MAIN_CHAIN_ID);
```

PR #360, opened 2021-01-06
Addressing clippy findings for storage

mat13mn:


storage/src/lib.rs
```diff
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: MIT
 #![forbid(unsafe_code)]
 #![feature(const_fn)]
+#![allow(clippy::ptr_arg)]
```
bkontur, line 5: @mat13mn @jurajselep 
hmm, I am not sure, that we want allowing anything right now,
There is no main goal to get rid off clippy now (at the end it will be, sure),
I would prefer to solve partial stuff in every PR, 
because you solve some stuff in storate module, but lots of these things repeat in other modules also,
so thats why I dont like this partial allowing (do it in whole project or not),
and also ptr_arg is this warning: 
`writing `&Vec<_>` instead of `&[_]` involves one more reference and cannot be used with non-Vec-based slices. `
and for this we have global jira to be solved https://simplestakingcom.atlassian.net/browse/TE-228

mat13mn: sure, no problem

PR #361, opened 2021-01-07
[TE-280] dedicated database instances for context and context actions

mat13mn:
dedicated database instances for context and context actions


storage/src/persistent/mod.rs
```diff
@@ -125,20 +130,32 @@ pub struct PersistentStorage {
...
             seq,
-            merkle: Arc::new(RwLock::new(MerkleStorage::new(kv))),
+            merkle: Arc::new(RwLock::new(MerkleStorage::new(kv_context))),
         }
     }
 
     #[inline]
-    pub fn kv(&self) -> Arc<DB> {
-        self.kv.clone()
+    pub fn kv(&self, storage: StorageType) -> Arc<DB> {
```
bkontur, line 153: @mat13mn 
wow, this PR looks easier then I thought :D,


I am just thinking about, if it would be possible to return here not `Arc<DB>` but something like:
```
Arc<dyn KeyValueStoreWithSchema<GenericSomething> + Sync + Send>
```

I mean `DB` is rocksdb struct, and `KeyValueStoreWithSchema` is our KV trait, so we could cover implementation detail, becasuse we use it like:
```
pub type BlockMetaStorageKV = dyn KeyValueStoreWithSchema<BlockMetaStorage> + Sync + Send;
...
pub struct BlockMetaStorage {
    kv: Arc<BlockMetaStorageKV>,
...
pub fn new(persistent_storage: &PersistentStorage) -> Self {
        BlockMetaStorage {
            kv: persistent_storage.kv(),
```


I am just thinking, never try it, and I dont know if it is even doable and not so generic-complicated

mat13mn: i tried that but i run into problems, i ended up with 
```
    pub fn kv(&self, storage: StorageType) -> Arc<dyn KeyValueStoreWithSchema<KeyValueSchema<Key=[u8; 32],Value=Vec<u8>>> + Sync + Send> {
```

but compiler is still complaining 

```
storage/src/persistent/mod.rs:160:79
    |
160 |     pub fn kv(&self, storage: StorageType) -> Arc<dyn KeyValueStoreWithSchema<KeyValueSchema<Key=[u8; 32],Value=Vec<u8>>> + Sync + Send> {
    |                                                                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: use `dyn`: `dyn KeyValueSchema<Key=[u8; 32],Value=Vec<u8>>`
    |
    = note: `#[warn(bare_trait_objects)]` on by default

error[E0038]: the trait `schema::KeyValueSchema` cannot be made into an object
   --> storage/src/persistent/mod.rs:160:47
    |
160 |     pub fn kv(&self, storage: StorageType) -> Arc<dyn KeyValueStoreWithSchema<KeyValueSchema<Key=[u8; 32],Value=Vec<u8>>> + Sync + Send> {
    |                                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `schema::KeyValueSchema` cannot be made into an object
    |
note: for a trait to be "object safe" it needs to allow building a vtable to allow the call to be resolvable dynamically; for more information visit <https://doc.rust-lang.org/reference/items/traits.html#object-safety>
   --> storage/src/persistent/schema.rs:14:8
    |
10  | pub trait KeyValueSchema {
    |           -------------- this trait cannot be made into an object...
...
14  |     fn descriptor(cache: &Cache) -> ColumnFamilyDescriptor {
    |        ^^^^^^^^^^ ...because associated function `descriptor` has no `self` parameter
...
18  |     fn name() -> &'static str;
    |        ^^^^ ...because associated function `name` has no `self` parameter

```

i`m not sure how to fix taht - cause i dont think changing this static function into non-static actually makes sense ...

bkontur: ok, ok, no problem, just leave it as it is now,
we will change it in future, when needed

light_node/src/main.rs
```diff
@@ -469,6 +471,54 @@ fn block_on_actors(
...
 
+fn initialize_databases(caches: &[Cache; 3], env: &Environment) -> Result<Vec<Arc<DB>>, DBError> {
+    let paths = vec![
+        env.storage.db_path.join("db"),
+        env.storage.db_path.join("context"),
+        env.storage.db_path.join("context_actions"),
+    ];
+
+    let main_cache = &caches[0];
+    let context_cache = &caches[0];
```
bkontur, line 482: @mat13mn 
all the time you use just `caches[0]`

light_node/src/main.rs
```diff
@@ -518,85 +568,81 @@ fn main() {
...
-        context_action_storage::ContextActionByContractIndex::descriptor(&cache),
-        context_action_storage::ContextActionByTypeIndex::descriptor(&cache),
-        ContextActionStorage::descriptor(&cache),
-        MerkleStorage::descriptor(&cache),
-        SystemStorage::descriptor(&cache),
-        Sequences::descriptor(&cache),
-        MempoolStorage::descriptor(&cache),
-        ChainMetaStorage::descriptor(&cache),
-        PredecessorStorage::descriptor(&cache),
+    let cache = [
```
bkontur, line 571: @mat13mn 
if we dont need cache later, I would move this cache initialization inside `initialize_databases`

light_node/src/main.rs
```diff
@@ -518,85 +568,81 @@ fn main() {
...
-                actor_system,
-                persistent_storage,
-                tezedge_context,
-                log,
-            ),
-            Err(e) => shutdown_and_exit!(
-                error!(log, "Failed to resolve init storage chain data."; "reason" => e),
-                actor_system
-            ),
+        if let [kv, kv_context, kv_actions] = &storages[..] {
```
bkontur, line 610: @mat13mn 
what if we remove `&` here,
than we dont need to copy kv.clone(),
or something like 'to give owner ship to PersistentStorage"

light_node/src/main.rs
mat13mn: that was also my idea but i found this note in the code :/
```
    // create common RocksDB block cache to be shared among column families
    // IMPORTANT: Cache object must live at least as long as DB (returned by open_kv)
```

https://github.com/simplestaking/tezedge/blob/master/light_node/src/main.rs#L513

light_node/src/main.rs
mat13mn: i tried - results with compilation error
```
error[E0508]: cannot move out of type `[Arc<DB>]`, a non-copy slice
   --> light_node/src/main.rs:610:47
    |
610 |         if let [kv, kv_context, kv_actions] = storages[..] {
    |                 --  ----------  ----------    ^^^^^^^^^^^^
    |                 |   |           |             |
    |                 |   |           |             cannot move out of here
    |                 |   |           |             help: consider borrowing here: `&storages[..]`
    |                 |   |           ...and here
    |                 |   ...and here
    |                 data moved here
    |
    = note: move occurs because these variables have types that don't implement the `Copy` trait

```

bkontur: ok, no problem

storage/src/persistent/mod.rs
```diff
@@ -125,20 +130,37 @@ pub struct PersistentStorage {
     merkle: Arc<RwLock<MerkleStorage>>,
 }
 
+pub enum StorageType {
+    Database,
+    Context,
+    ContxtAction,
```
bkontur, line 136: @mat13mn 
just a small typo: ContxtAction -> ContextAction :)

light_node/src/main.rs
```diff
@@ -516,85 +566,81 @@ fn main() {
...
-            crit!(log, "Database incompatibility detected"),
-            actor_system
-        ),
-        Err(e) => shutdown_and_exit!(
-            error!(log, "Failed to verify database compatibility"; "reason" => e),
-            actor_system
-        ),
-        _ => (),
+    for db in storages.iter() {
+        match check_database_compatibility(db.clone(), DATABASE_VERSION, &tezos_env, &log) {
```
bkontur, line 587: @mat13mn 
I like the fix with that split SystemStorage to every subpart, I didnt think of that before :), but make sense,
so I think, that we should split also constant DATABASE_VERSION for every storage, that it make totally sense,

bkontur: @mat13mn 
and last thing, constant LRU_CACHE_SIZE_128MB I would split also for every storate, and set it somehow,
context_actions storage does not need 128, maybe just half, or just 32MB

bkontur:
@mat13mn 
closing, it was merge with another PR

PR #362, opened 2021-01-11
Fixing ipc

bkontur:


tizoc:
@bkontur `IpcCmdServer::accept` calls `self.0.accept` which doesn't exist anymore (should be `try_accept` instead), same for `IpcEvtServer::accept`.

bkontur:
> 
> 
> @bkontur `IpcCmdServer::accept` calls `self.0.accept` which doesn't exist anymore (should be `try_accept` instead), same for `IpcEvtServer::accept`.

I know, I just fixed `ipc` module, tomorrow continue with other stuff,
but `ipc` tests works

PR #363, opened 2021-01-12
TE-264 Use checked arithmetics for encoded size

akoptelov:
Not sure if it is possible to add any tests for testing usize overflow though...

### DIFF
--- a/tezos/encoding/src/binary_writer.rs
+++ b/tezos/encoding/src/binary_writer.rs
@@ -79,7 +79,11 @@ fn encode_record(data: &mut Vec<u8>, value: &Value, schema: &[Field]) -> Result<
                     .unwrap_or_else(|| panic!("No values found for {}", name));
                 let encoding = field.get_encoding();
 
-                bytes_sz += encode_any(data, value, encoding)?;
+                bytes_sz = bytes_sz
+                    .checked_add(encode_any(data, value, encoding)?)
+                    .ok_or_else(|| {
+                        Error::custom("Encoded message size overflow while encoding a record field")
+                    })?;
             }
 
             Ok(bytes_sz)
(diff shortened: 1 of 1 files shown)


tezos/encoding/src/binary_writer.rs
```diff
@@ -472,6 +472,16 @@ fn find_value_in_record_values<'a>(
...
 }
 
+fn encoded_size_add(size: usize, val: usize) -> Result<usize, Error> {
+    size.checked_add(val)
+        .ok_or(Error::custom("Encoded message size overflow"))
+}
+
+fn encoded_size_sub(size: usize, val: usize) -> Result<usize, Error> {
+    size.checked_sub(val)
+        .ok_or(Error::custom("Encoded message size overflow"))
```
bkontur, line 482: @akoptelov 
could you please just change: overflow -> underflow?

akoptelov: @bkontur From CPU point of view this is still overflow, see the std documentation: https://doc.rust-lang.org/std/primitive.usize.html#method.checked_sub

bkontur: @akoptelov 
aha, ok, ok, I just want to have different error messages for different cases - that we could differ directly from the log in case of error

akoptelov: To my understanding this is very unlikely that this ever happen. And if it will, it will be quite hard reproducing it with backtrace option. So probably we can use different messages for each place instead. I'll do that.

akoptelov: Please take a look if there is enough information now for possible troubleshooting.

bkontur: cool, looks ok :)

bkontur (approved):

bkontur:
@akoptelov 
merged, close please jira

PR #368, opened 2021-01-13
[TE-251] Remove expect from encoding

akoptelov:
Replaced `expect()` with proper error reporting.

However, I cannot imagine how that can be reproduced in a usual way of using `write()` call since it is impossible to instantiate an enum with no variants, and for `Tags` only `Value::Tag` is used, not `Value::Enum`

### DIFF
--- a/tezos/encoding/src/binary_writer.rs
+++ b/tezos/encoding/src/binary_writer.rs
@@ -3,8 +3,8 @@
 
 //! Tezos binary data writer.
 
-use std::cmp;
 use std::mem::size_of;
+use std::{cmp, convert::TryFrom};
 
 use bit_vec::BitVec;
 use byteorder::{BigEndian, WriteBytesExt};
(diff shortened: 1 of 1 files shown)


bkontur:
> 
> 
> Replaced `expect()` with proper error reporting.
> 
> However, I cannot imagine how that can be reproduced in a usual way of using `write()` call since it is impossible to instantiate an enum with no variants, and for `Tags` only `Value::Tag` is used, not `Value::Enum`

you are right, some of these expect/unwrap it is impossible to "trigger",
well, we are now on the wave "to remove expect/unwrap everywhere, because they could cause thread panic or application shutdown", this is maybe kind of cleaning for future, future refactors, if something change...

akoptelov:
@bkontur so, is it ok to merge?

tezos/encoding/src/binary_writer.rs
```diff
@@ -226,7 +226,10 @@ fn encode_value(data: &mut Vec<u8>, value: &Value, encoding: &Encoding) -> Resul
             match value {
                 Value::Enum(_, ordinal) => {
                     // TODO: handle enum of different sizes
-                    data.put_u8(ordinal.expect("Was expecting enum ordinal value") as u8);
+                    data.put_u8(
+                        ordinal.ok_or_else(|| Error::custom("Was expecting enum ordinal value"))?
+                            as u8,
```
tizoc, line 231: Why is ordinal `u32` but being casted to `u8` ? Can we use `TryFrom` or something like that here?

akoptelov: Good point. Let me check that.

bkontur (approved):
@akoptelov 

PR #371, opened 2021-01-14
[TE-276] Introduce an error on exceeding 128 size

akoptelov:
This is the fix in the case we don't want to replace `base58` crate. Now our base58 refuses to accept data/strings longer than 128 bytes.

This is still a compromised solution. It is possible to introduce a completely new hash type longer than 128 bytes that will cause panic, but that will be discovered very early during development. Also user input does not affect the length of the hashes being encoded. For decoding the error will be reported.

### DIFF
--- a/crypto/src/hash.rs
+++ b/crypto/src/hash.rs
@@ -237,7 +237,12 @@ impl HashType {
             let mut hash = Vec::with_capacity(self.base58check_prefix().len() + data.len());
             hash.extend(self.base58check_prefix());
             hash.extend(data);
-            Ok(hash.to_base58check())
+            hash.to_base58check()
+                // currently the error is returned if the input lenght exceeds 128 bytes
+                // that is the limitation of base58 crate, and there are no hash types
+                // exceeding that limit.
+                // TODO: remove this when TE-373 is implemented
+                .map_err(|_| unreachable!("Hash size should not exceed allowed 128 bytes"))
         }
     }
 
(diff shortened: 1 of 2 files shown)


bkontur:
> 
> 
> This is the fix in the case we don't want to replace `base58` crate. Now our base58 refuses to accept data/strings longer than 128 bytes.
> 
> This is still a compromised solution. It is possible to introduce a completely new hash type longer than 128 bytes that will cause panic, but that will be discovered very early during development. Also user input does not affect the length of the hashes being encoded. For decoding the error will be reported.

ok, I will check, and we also need to refactor hash.rs, see this https://simplestakingcom.atlassian.net/browse/TE-228

crypto/src/hash.rs
```diff
@@ -237,7 +237,9 @@ impl HashType {
             let mut hash = Vec::with_capacity(self.base58check_prefix().len() + data.len());
             hash.extend(self.base58check_prefix());
             hash.extend(data);
-            Ok(hash.to_base58check())
+            Ok(hash
+                .to_base58check()
+                .unwrap_or_else(|_| unreachable!("Hash size should not exceed allowed 128 bytes")))
```
bkontur, line 242: @akoptelov 
can we just map_err to FromBytesError here?

@akoptelov 
also please, check all usage of `unreachable!()` and add message there please, why is it unreachable

bkontur (approved):
@akoptelov 
see comment,

PR #377, opened 2021-01-25
[TE-247] Remove path codec recursion

akoptelov:
This is quite big PR. It addresses several issues with recursive data structure.

**binary_reader/writer**. `Lazy` encoding was used for recursive data structures that implied recursion in runtime. Now `Custom` encoding is used that supplies plain functions for encoding/decoding.
**Serde**.  Serialization and deserialization are recursive, so they are also vulnerable for stack overflow. Now custom encoding is used for `Path` that checks that the size does not exceed maximal value. The `depth` field is used to store path size for each `PathRight/PathLeft` basing on sub path, so to properly deserialize these structures also custom deserializer is implemented.

Fixes alto TE-232.

### DIFF
--- a/tezos/messages/src/p2p/encoding/operations_for_blocks.rs
+++ b/tezos/messages/src/p2p/encoding/operations_for_blocks.rs
@@ -1,19 +1,31 @@
 // Copyright (c) SimpleStaking and Tezedge Contributors
 // SPDX-License-Identifier: MIT
 
-use std::mem::size_of;
+use std::collections::VecDeque;
 use std::sync::Arc;
 
+use bytes::Buf;
+
 use getset::{CopyGetters, Getters};
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Serialize, Serializer};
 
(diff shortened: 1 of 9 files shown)


tezos/messages/src/p2p/encoding/operations_for_blocks.rs
```diff
@@ -1,19 +1,28 @@
...
+use tezos_encoding::types::{RecursiveDataSize, Value};
+use tezos_encoding::{has_encoding, safe};
 
 use crate::cached_data;
 use crate::p2p::binary_message::cache::BinaryDataCache;
 use crate::p2p::encoding::operation::Operation;
+use tezos_encoding::json_writer::JsonWriter;
+
+/// Maximal depth for path to be decoded
+pub const PATH_MAX_DEPTH: RecursiveDataSize = 512;
```
bkontur, line 25: @akoptelov 
please add there detailed comments, why this is 512 (like we talked on slack)
\and TODO: with jira linked, where we will continue on that

then please rebase and merge

bkontur (approved):

PR #378, opened 2021-01-25
[TE-262] Use pattern binding instead of explicit unwrapping

akoptelov:
Unwrapping is removed.


bkontur (approved):
@akoptelov 
please rebase and merge to develop

PR #380, opened 2021-01-25
Refactor peer handshake, bootstrap peer before create peer actor

vlad9486:


shell/src/peer_manager.rs
```diff
@@ -604,9 +605,16 @@ impl Receive<ConnectToPeer> for PeerManager {
             match timeout(CONNECT_TIMEOUT, TcpStream::connect(&msg.address)).await {
                 Ok(Ok(stream)) => {
                     debug!(system.log(), "(Outgoing) Connection to peer successful"; "ip" => msg.address);
-                    match Self::create_peer(&system, local_node_info, network_channel, tokio_executor, &msg.address) {
+                    match Self::create_peer(&system, network_channel.clone(), tokio_executor, &msg.address) {
                         Ok(peer) => {
-                            peer.tell(Bootstrap::outgoing(stream, msg.address.clone(), disable_mempool, private_node), None);
+                            debug!(system.log(), "Bootstrapping"; "incoming" => false, "ip" => &msg.address);
+                            match bootstrap(Bootstrap::outgoing(stream, msg.address.clone(), disable_mempool, private_node), local_node_info, &system.log()).await {
+                                Ok(bootstrap_output) => peer.tell(bootstrap_output, None),
```
bkontur, line 612: @vlad9486 
can we now call Self::create_peer after bootstrap.await ?
```
Ok(bootstrap_output) => {
    Self::create_peer(...).tell(bootstrap_output, None)
}

File: networking/src/p2p/peer.rs [hunk 2]
```diff
@@ -217,80 +214,49 @@ impl Actor for Peer {
...
+
         let myself = ctx.myself();
         let system = ctx.system.clone();
-        let net = self.net.clone();
+        let mut net = self.net.clone();
         let network_channel = self.network_channel.clone();
-        self.remote_addr = msg.address;
+        self.remote_addr = peer_address;
 
         self.tokio_executor.spawn(async move {
```
bkontur, line 229: @vlad9486 
ok, very cool,
and last thing I would like to move all this stuff `self.tokio_executor.spawn(async move {` to `fn create_args`,
and remove `impl Receive<BootstrapOutput>`, so BootstrapOutput will be an argument of `fn actor(`

so in peer_manager, there will be just `Self::create_peer(` and no `peer.tell(bootstrap_output, None)`,
so this will be the final: peer_manager just calls `create_peer`

bkontur (approved):

PR #376, opened 2021-01-22
New context action recording

tizoc:
OCaml-side: https://gitlab.com/simplestaking/tezos/-/merge_requests/16

Changes:

- Simplified FFI callbacks. A single "send" function is used now, with "record" and "perform" flags.
- `ContextActionMessage` wrapper for `ContextAction` that carries along these flags, which are checked by the context listener to see what to do. Read actions will contain both flags. Write actions will contain the "record" flag when initially sent  from the OCaml side, and the "perform" flag when replayed (but not the "record" one).
- New `tree_hash` (hash of the tree to which the action is being applied) and `new_tree_hash` (on actions that produce a new tree, this is the hash of the new tree) fields on `ContextAction`.

PR #332, opened 2020-12-03
Upgrade ocaml interop

tizoc:
NOTE: this was made on top of #383, which should go first (initially it was the other way around).

Changes:

- Upgraded to newer ocaml-interop 
- Switched to an exception-less tezos-ffi. Tezos errors are represented by the new `TezosError` and `TezosErrorTrace` types.
- Mapping of errors doesn't require any more parsing, instead the name of the exception/error is matched.
- Unwrapped all FFI return values that had the form `Result<Result<T, CustomError>, OcamlError>`, which now look like `Result<T, CustomError>`.
- Initialization and shutdown of the OCaml runtime is handled by the OCaml executor thread
- Removed GC threshold setting
- `OcamlError` has been renamed `OCamlBlockPanic`, and with exceptions not being raised now this should only happen when the block passed to the executor panics.

Matching tezos fork MR: https://gitlab.com/simplestaking/tezos/-/merge_requests/15

## What changed on ocaml-interop

- Switched to a caller-rooting calling convention, which means that now values have to be rooted before passing them to OCaml functions (which now receive `OCamlRef<T>` instead of `OCaml<T>` arguments. That is why you will not see the `ocaml_call!` macro anymore.
- `ocaml_frame!` doesn't greate a `gc` handle anymore, but instead it takes as its first argument a `&mut OCamlRuntime` that has to be passed by the caller. In the case of tezedge, the executor thread creates that handle (upon initialization of the OCaml runtime), and the closure that is passed to the `runtime::execute` function now must accept one `&mut OCamlRuntime` argument that gets passed by `runtime::execute`.
- Now any exception raised by OCaml code will result in a `panic!`. The tezedge-ffi code has been modified to not raise exceptions but instead return `Result<T, ErrorTrace>` values to signal errors. Also `runtime::execute` now returns `Result<T, OCamlBlockPanic>`, with `OCamlBlockPanic` signaling that a panic ocurred inside the executed closure (either raised by Rust code, or a rogue OCaml exception).

PR #383, opened 2021-01-26
[TE-320] New context action recording develop

tizoc:
https://github.com/simplestaking/tezedge/pull/376 but against develop

OCaml-side https://gitlab.com/simplestaking/tezos/-/merge_requests/17

*description copied from #376*:

Changes:

- Simplified FFI callbacks. A single "send" function is used now, with "record" and "perform" flags.
- `ContextActionMessage` wrapper for `ContextAction` that carries along these flags, which are checked by the context listener to see what to do. Read actions will contain both flags. Write actions will contain the "record" flag when initially sent  from the OCaml side, and the "perform" flag when replayed (but not the "record" one).
- New `tree_hash` (hash of the tree to which the action is being applied) and `new_tree_hash` (on actions that produce a new tree, this is the hash of the new tree) fields on `ContextAction`.


tizoc:
Cancelled the Drone build, but it already passed here: http://ci.tezedge.com/simplestaking/tezedge/2346

tizoc:
@bkontur when do you think you will have time to go through this one (the easy one) and #332 (harder one)? it would be better to not have to spend time to keep them up to date with develop. If there is anything you need from me to ease the review process just let me know and I will do it.

storage/tests/context_action_storage.rs
```diff
@@ -28,39 +28,43 @@ fn context_get_values_by_block_hash() -> Result<(), Error> {
         operation_hash: None,
         block_hash: Some(str_block_hash_1.into()),
         context_hash: None,
+        tree_hash: "ignored".into(),
```
bkontur, line 31: @tizoc 
here it is :),
I guess Option will be better

tezos/context/src/channel.rs
```diff
@@ -46,40 +46,44 @@ pub enum ContextAction {
         context_hash: Option<Hash>,
         block_hash: Option<Hash>,
         operation_hash: Option<Hash>,
+        tree_hash: Hash,
```
bkontur, line 49: @tizoc 
Option<Hash>

tezos/context/src/lib.rs
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) SimpleStaking and Tezedge Contributors
 // SPDX-License-Identifier: MIT
-#![forbid(unsafe_code)]
+
+// TODO: re-enable this once ocaml-interop has been upgraded
+//#![forbid(unsafe_code)]
```
bkontur, line 5: @tizoc 
what about this?

tizoc: @bkontur the ocaml-interop-upgrade branch re-enables it. Remember that originally this branch depended on the other, but things had to be turned around. That required some of the newer ocaml-interop features to be imported directly into that branch, and that code requires `unsafe` features. The ocaml-interop upgrade branch removes the imported code and re-enables forbid-unsafe

tizoc:
@bkontur added the new flag + make those hashes into Option here:
- Rust side: https://github.com/simplestaking/tezedge/pull/383/commits/51d247c12ca078d9bc204ad9de66b4f1142d90d9
- OCaml side: https://gitlab.com/simplestaking/tezos/-/merge_requests/17/diffs?commit_id=4161117791f68a463fec9cf97c069e66e55638f7 (sorry, there are some formatting changes because of ocaml-format too)

Test suite will fail right now because I need to produce a new libtezos binary, but I will just do it for the ocaml-interop upgrade branch because it includes everything.

tizoc:
I am very confused about the failure in drone when running the benchmarks (a segfault!).

@bkontur how does it work exactly? I saw that it builds the benchmarks in both current and target branches.

Is it possible that when rebuilding from two different branches it is overriding the tezos ffi library? I was not failing before, and the only thing that has changed now is that the runtime configuration function has an extra boolean argument, so that is my guess of what is going on.

I have not been able to reproduce it locally either.

tizoc:
I'm more confused now, because the ocaml-interop PR (which includes everything from this PR) did not fail, so my theory is wrong.

bkontur (approved):
@tizoc 
there are few clippy stuff in tezos/client/tests/*
please, squash a little bit, and merge to develop,

light_node/src/configuration.rs
```diff
@@ -844,17 +850,24 @@ impl Environment {
...
                     threads: db_context_actions_threads_count,
                 };
+                let store_context_actions = args
+                    .value_of("store-context-actions")
+                    .unwrap_or("true")
+                    .parse::<bool>()
+                    .expect("Provided value cannot be converted to bool");
+                let compute_context_action_tree_hashes = args
+                    .value_of("compute-context-action-tree-hashes")
+                    .unwrap_or("true")
```
bkontur, line 860: @tizoc 
please change "true" to "false" here

tizoc:
@bkontur yes, that SIGSEV is what I mentioned in the comment above. I cannot reproduce it locally, and it is even weirder when you consider that in the upgrade-ocaml-interop branch it doesn't fail, but it includes everything that is on this branch (and the matching tezedge-ffi branch too). My theory was that maybe drone was overriding libtezos-ffi when building the other branch, but if that were the case. the upgrade-ocaml-interop branch would fail too, and it doesn't.

Yesterday I made a dummy PR from the commit before the new flag and everything passed: http://ci.tezedge.com/simplestaking/tezedge/2668

tizoc:
Closing this one, will be merged as part of #332 .

PR #390, opened 2021-01-28
Context action file storage

mambisi:


action_sync/src/file.rs
```diff
@@ -0,0 +1,263 @@
...
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct Block {
+    pub block_level: u32,
+    pub block_hash_hex: String,
+    pub block_hash: [u8; BLOCK_HASH_HEADER_LEN],
+    pub predecessor: [u8; BLOCK_HASH_HEADER_LEN],
+}
+
+impl Block {
+    pub fn new(block_level: u32, raw_block_hash: Vec<u8>, raw_predecessor: Vec<u8>) -> Self {
```
mat13mn, line 25: someone on security channel said that he is working on removeing Vec<u8> as hash representation instead array are used. It might be good idea to follow that approach with new development :thinking:

action_sync/src/file.rs
```diff
@@ -0,0 +1,263 @@
...
+    }
+
+    pub fn header(&self) -> ActionsFileHeader {
+        self.header
+    }
+}
+
+unsafe impl Send for ActionsFileWriter {}
+
+unsafe impl Sync for ActionsFileWriter {}
```
mat13mn, line 203: i dont think this one is really sync... as cannot we wrap it into Mutex ?

action_sync/src/file.rs
```diff
@@ -0,0 +1,263 @@
...
+        }
+
+        let mut out = Vec::new();
+        let writer = snap::write::FrameEncoder::new(&mut out);
+        bincode::serialize_into(writer, &(block, actions))?;
+
+        // Writes the header if its not already set
+        if self.header.block_count <= 0 {
+            let header_bytes = self.header.to_vec();
+            let _ = self.file.seek(SeekFrom::Start(0));
```
mat13mn, line 224: consider unwraping result instead of ignoring it

action_sync/src/file.rs
```diff
@@ -0,0 +1,263 @@
...
+    pub fn new<P: AsRef<Path>>(path: P) -> Result<Self> {
+        let file = OpenOptions::new()
+            .write(true)
+            .create(true)
+            .read(true)
+            .open(path)?;
+        let mut reader = BufReader::new(file.try_clone()?);
+        let _ = reader.seek(SeekFrom::Start(0));
+        let mut h = [0_u8; HEADER_LEN];
+        let _ = reader.read_exact(&mut h);
```
mat13mn, line 191: unwrap

action_sync/src/file.rs
```diff
@@ -0,0 +1,263 @@
...
+
+impl ActionsFileWriter {
+    pub fn new<P: AsRef<Path>>(path: P) -> Result<Self> {
+        let file = OpenOptions::new()
+            .write(true)
+            .create(true)
+            .read(true)
+            .open(path)?;
+        let mut reader = BufReader::new(file.try_clone()?);
+        let _ = reader.seek(SeekFrom::Start(0));
```
mat13mn, line 189: unwrap or pass the error to outher scope

action_sync/src/file.rs
```diff
@@ -0,0 +1,263 @@
...
+        actions_count: u32,
+        block_hash: [u8; BLOCK_HASH_HEADER_LEN],
+    ) {
+        self.header.block_height = block_level;
+        self.header.actions_count += actions_count;
+        self.header.block_count += 1;
+        self.header.current_block_hash = block_hash;
+
+        let header_bytes = self.header.to_vec();
+        let _ = self.file.seek(SeekFrom::Start(0));
```
mat13mn, line 244: unwrap or pass error to outher scope

action_sync/src/file.rs
```diff
@@ -0,0 +1,263 @@
...
+        let _ = self.file.write(&header_bytes);
+    }
+
+    fn _fetch_header(&mut self) {
+        let _ = self.file.seek(SeekFrom::Start(0));
```
mat13mn, line 249: here and in other place unwrap or pass error to outher scope

shell/src/context_listener.rs
```diff
@@ -188,6 +191,7 @@ fn listen_protocol_events(
     apply_block_run: &AtomicBool,
     event_server: &mut IpcEvtServer,
     event_server_accept_timeout: Duration,
+    actions_file_storage: &mut Option<ActionFileStorage>,
```
mat13mn, line 194: arent `actions_file_storage` kind of the same thing ? they seems to store actions on demand. Maybe ActionFileStorage and `ContextActionStorage` should be implementing some sort of abstraction and throught configuration it should be possible to choose one of those ?

mambisi: Its kinda the same thing but its much faster to fetch from ActionFile and RockDB use a lot memory so it might squee memory statistics , but we can build an abstraction over it.

action_sync/src/file.rs
mambisi: Ohk

action_sync/src/file.rs
mambisi: alright

action_sync/src/file.rs
mambisi: Yeah, i would make some of the code better. is there a pull request for it?

shell/src/context_listener.rs
mambisi: `actions_file_storage` is also deleted when the node restarts

tizoc:
@mambisi @mat13mn is this PR still relevant?

PR #393, opened 2021-01-28
[TE-268] Remove assert from base58 to hash

akoptelov:
Replace `assert_eq!` with error reporting.

This PR is built on top of the #387

### DIFF
--- a/crypto/src/base58.rs
+++ b/crypto/src/base58.rs
@@ -20,6 +20,11 @@ pub enum FromBase58CheckError {
     /// Data is too long
     #[fail(display = "data too long")]
     DataTooLong,
+    #[fail(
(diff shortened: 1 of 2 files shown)


crypto/src/base58.rs
```diff
@@ -17,6 +17,8 @@ pub enum FromBase58CheckError {
     /// The input is missing checksum.
     #[fail(display = "missing checksum")]
     MissingChecksum,
+    #[fail(display = "mismatched data lenght")]
+    MismatchedLength,
```
bkontur, line 21: @akoptelov 
please, add here two attribtues:
- expected_length
- actual_length

it will really help to better investigate later error

bkontur: @akoptelov 

bkontur (approved):

PR #399, opened 2021-01-30
fix(storage_backend) changed `Vec<EntryHash>` to `HashSet<EntryHash>`

mambisi:


storage/src/backend/mark_sweep_gced.rs
```diff
@@ -46,8 +46,7 @@ impl<T: 'static + KVStore + Default> MarkSweepGCed<T> {
     }
 
     fn sweep_entries(&mut self, todo: HashSet<EntryHash>)  -> Result<(), KVStoreError> {
-        let p = todo.into_iter().collect::<Vec<_>>();
-        self.retain(p);
+        self.retain(todo);
```
tizoc, line 49: @mambisi how does sweep_entries work? (I'm confused about the contrast between `sweep` and `retain`)

mambisi: `retain` is storage backend function that removes all items that is not in the `HashSet`,
it is similar to
https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.drain_filter

PR #402, opened 2021-02-02
[TE-253] Result for bitset

akoptelov:
Introduced `Result` return type for operations that can fail.


bkontur (approved):
@akoptelov 

PR #404, opened 2021-02-02
Sort out assertions [TE-266]

akoptelov:


storage/src/block_meta_storage.rs
```diff
@@ -401,13 +401,6 @@ impl Decoder for Meta {
             // predecessor
             let predecessor = if has_predecessor!(mask) {
                 let block_hash = bytes[IDX_PREDECESSOR..IDX_LEVEL].to_vec();
-                assert_eq!(
```
akoptelov, line 404: This should newer be triggered as the resulting length is  `IDX_LEVEL-IDX_PREDECESSOR` that is `LEN_BLOCK_HASH`, and also correct length is checked below in `block_hash.try_into()?`.

storage/src/block_meta_storage.rs
```diff
@@ -427,26 +420,12 @@ impl Decoder for Meta {
                         let block_hash = bytes
                             [next_successor_index..(next_successor_index + LEN_BLOCK_HASH)]
                             .to_vec();
-                        assert_eq!(
```
akoptelov, line 430: This should newer be triggered as the resulting length is  `next_successor_index + LEN_BLOCK_HASH-next_successor_index` that is `LEN_BLOCK_HASH`, and also correct length is checked below in `BlockHash::try_from(block_hash)?`.

storage/src/block_meta_storage.rs
```diff
@@ -427,26 +420,12 @@ impl Decoder for Meta {
...
-                            LEN_BLOCK_HASH,
-                            block_hash.len()
-                        );
                         successors.push(BlockHash::try_from(block_hash)?);
                     }
                 }
                 successors
             };
 
-            assert_eq!(
```
akoptelov, line 443: This should newer be triggered as the resulting length is  `LEN_CHAIN_ID`, and also correct length is checked below in `chain_id.try_into()?`.

storage/src/block_meta_storage.rs
```diff
@@ -489,7 +468,7 @@ impl Encoder for Meta {
                 value.extend(successor.as_ref());
             });
         }
-        assert_eq!(
+        debug_assert_eq!(
```
akoptelov, line 471: This looks rather like a debug assertion that the result has correct size.

storage/src/block_meta_storage.rs
```diff
@@ -525,17 +504,20 @@ fn merge_meta_value(
...
+            return None;
+        }
+    }
+
     let mut result = existing_val.map(|v| v.to_vec());
 
     for op in operands {
         match result {
             Some(ref mut val) => {
-                assert!(
```
akoptelov, line 533: To me it looks like the length constraint should be checked for both existing value and merge operands. Also we cannot do anything apart returning `None` in case of invalid value.

storage/src/block_meta_storage.rs
```diff
@@ -565,7 +547,7 @@ fn merge_meta_value(
                 }
 
                 let total_len = total_len(op_successors_count);
-                assert_eq!(total_len, val.len(), "Invalid length after merge operator was applied. Was expecting {} but found {}.", total_len, val.len());
+                debug_assert_eq!(total_len, val.len(), "Invalid length after merge operator was applied. Was expecting {} but found {}.", total_len, val.len());
```
akoptelov, line 550: This looks like a check for correct calculation, so `debug_assert` looks appropriate.

bkontur (approved):
@akoptelov 

PR #406, opened 2021-02-02
Replace try_from/unwrap with as cast [TE-307]

akoptelov:
Looks like replacing `try_into().unwrap()` with plain `as` makes the code clearer.


bkontur (changes requested):
@akoptelov 
sorry, this file was merged to block_state.rs in actual develop,
please check it

akoptelov:
The file is gone for good.

PR #409, opened 2021-02-03
Properly handle connection pool timeout [TE-330]

akoptelov:


light_node/src/main.rs
```diff
@@ -240,27 +243,45 @@ fn block_on_actors(
...
         "tezos_readonly_api_pool",
         env.ffi.tezos_readonly_api_pool.clone(),
         &env,
         tezos_env.clone(),
         log.clone(),
-    ));
-    let tezos_readonly_prevalidation_api_pool = Arc::new(create_tezos_readonly_api_pool(
+    ) {
+        Ok(pool) => Arc::new(pool),
+        Err(e) => shutdown_and_exit!(
```
bkontur, line 254: @akoptelov 
please, rebase on actual develop,
because this is changed a litttle bit in actual develop,
shutdown_and_exit is removed and actor system is create when all is ok

bkontur (changes requested):
@akoptelov 
please, rebase on actual develop,
because this is changed a litttle bit in actual develop,
shutdown_and_exit is removed and actor system is create when all is ok

akoptelov:
@bkontur Updated changes. Now everything is using `expect()` in the `block_on_actors()`, I think we should sort out which error should be reported in other way, in a separate Jira.

bkontur (approved):
@akoptelov 
ok, please rebase and merge to develop

PR #410, opened 2021-02-03
Report error instead of unwrap [TE-331]

akoptelov:
I'll create a separate tasks for creating tests on lock poisoning (if possible) and invalid data location.


bkontur (approved):
@akoptelov 

PR #423, opened 2021-02-03
Pre commit rustfmt check

tizoc:
To install: `ln git-hooks/pre-commit .git/hooks/`.

But I guess we need something better than having everyone run this command manually?


tizoc:
Possible alternative https://github.com/swellaby/rusty-hook

tizoc:
Ok, edited to use https://crates.io/crates/cargo-husky

This dependency's build script will copy our custom hooks to where they should be, no need to manually link anything

tizoc:
So... I don't really like this empty-subcrate-with-that-dependency approach much tbh, maybe a `Makefile` (or with https://sagiegurari.github.io/cargo-make/) command to set the hooks would be better.

bkontur:
> 
> 
> So... I don't really like this empty-subcrate-with-that-dependency approach much tbh, maybe a `Makefile` (or with https://sagiegurari.github.io/cargo-make/) command to set the hooks would be better.

I dont like Makefile here, cargo make looks better, but I am thinking, if it is not easier just to run manually: `cargo fmt`

tizoc:
@bkontur what I mean is for setting up the git pre-commit hook, because it cannot be part of the repository (one could put something dangerous in there). On this PR I use an empty crate that depends o `cargo-husky`. What that does is that when cargo-husky is being built, it checks if there are any hooks that it must copy, and does so. This ensures that everybody has the hook set without requiring manual intervention.

The `make` thing would be to have a command that sets up the repository for development (right now all we have is a pre-commit hook, could do more things in the future), but requiring a single command from the developer (`make setup-development-repo` or something like that).

PR #435, opened 2021-02-11
Fb storage extended actions backtracking support

mat13mn:


shell/src/context_listener.rs
```diff
@@ -195,12 +205,54 @@ fn listen_protocol_events(
...
     context: &mut Box<dyn ContextApi>,
 ) -> Result<(), Error> {
+
+    if let Some(pre_hash) = get_tree_hash(&action){
+        context.set_merkle_root(pre_hash)?;
+    }
+    
     match action {
-        ContextAction::Get { key, .. } => {
+        ContextAction::Get { key,tree_hash, .. } => {
```
tizoc, line 255: this `tree_hash` is not being used for anything right? it is handled above if I'm not misunderstanding

storage/src/context.rs
```diff
@@ -126,6 +129,7 @@ impl ContextApi for TezedgeContext {
         let commit_hash = &commit_hash[..].to_vec();
 
         // associate block and context_hash
+        // TODO uncomment
```
tizoc, line 132: its uncommented already

storage/src/merkle_storage.rs
```diff
@@ -542,16 +440,14 @@ impl MerkleStorage {
 
     /// Get value from current staged root
     pub fn get(&mut self, key: &ContextKey) -> Result<ContextValue, MerkleError> {
-        let instant = Instant::now();
+        let _ = StatUpdater::new(& mut self.stats, MerkleStorageAction::Delete, Some(key));
+
         // build staging tree from saved list of actions (set/copy/delete)
```
tizoc, line 445: this comment is not relevant anymore, right?

storage/src/merkle_storage.rs
```diff
@@ -561,31 +457,27 @@ impl MerkleStorage {
 
     /// Check if value exists in current staged root
     pub fn mem(&mut self, key: &ContextKey) -> Result<bool, MerkleError> {
-        let instant = Instant::now();
+        let _ = StatUpdater::new(& mut self.stats, MerkleStorageAction::Mem, Some(key));
         // build staging tree from saved list of actions (set/copy/delete)
         // note: this can be slow if there are a lot of actions
```
tizoc, line 462: same here, doesn't apply to this new implementation

storage/src/merkle_storage.rs
```diff
@@ -561,31 +457,27 @@ impl MerkleStorage {
...
-        self.update_execution_stats("Mem".to_string(), Some(&key), &instant);
         rv
     }
 
     /// Check if directory exists in current staged root
     pub fn dirmem(&mut self, key: &ContextKey) -> Result<bool, MerkleError> {
-        let instant = Instant::now();
+        let _ = StatUpdater::new(& mut self.stats, MerkleStorageAction::DirMem, Some(key));
         // build staging tree from saved list of actions (set/copy/delete)
         // note: this can be slow if there are a lot of actions
```
tizoc, line 475: Outdated comment

storage/src/merkle_storage.rs
```diff
@@ -877,177 +757,85 @@ impl MerkleStorage {
...
+        key: &ContextKey,
+        value: &ContextValue,
+    ) -> Result<EntryHash, MerkleError> {
+        let blob_hash = hash_blob(&value);
+        self.put_to_staging_area(&blob_hash, Entry::Blob(value.clone()));
+        let new_node = Node {
+            entry_hash: blob_hash,
+            node_kind: NodeKind::Leaf,
+        };
+        let rv = self.compute_new_root_with_change(root, &key, Some(new_node));
```
tizoc, line 794: Other than for being able to play well with the current way actions are recorded, do we need to calculate the hash every time? cc @bkontur

storage/src/merkle_storage.rs
```diff
@@ -877,177 +757,85 @@ impl MerkleStorage {
...
+        rv
+    }
+
     /// Delete an item from the staging area.
+    //TODO: pelight
```
tizoc, line 799: what is this about?

storage/src/merkle_storage.rs
```diff
@@ -877,177 +757,85 @@ impl MerkleStorage {
...
-                self.staged[idx].1 += 1;
-            }
-            None => {
-                return Err(MerkleError::EntryNotFoundInStaging {
-                    hash: HashType::ContextHash.hash_to_b58check(hash),
-                });
-            }
-        }
+    /// Copy subtree under a new path.
+    /// TODO Consider copying values!
```
tizoc, line 818: can you expand on this? what do you mean? how is it not copied now?

storage/src/merkle_storage.rs
```diff
@@ -1059,201 +847,48 @@ impl MerkleStorage {
...
+    /// * `new_entry_hash` - None for deletion, Some for inserting a hash under the key.
     fn compute_new_root_with_change(
         &mut self,
-        root_hash: &EntryHash,
+        root: &Tree,
         key: &[String],
         new_node: Option<Node>,
     ) -> Result<EntryHash, MerkleError> {
-        assert_eq!(key.is_empty(), false);
         if key.is_empty() {
```
tizoc, line 863: If I am understanding this right, if the key is empty, we just replace the tree with a new one, right? would be good to have a comment about that here.

storage/src/merkle_storage.rs
```diff
@@ -2123,6 +1656,95 @@ mod tests {
...
+        assert_eq!(storage.dirmem(&key_abc).unwrap(), true);
+        assert_eq!(storage.mem(&key_abcd).unwrap(), true);
+
+        assert_eq!(storage.mem(&key_ax).unwrap(), true);
+
+        storage.delete(key_abcd);
+
+        assert_eq!(storage.dirmem(&key_a).unwrap(), true);
+        assert_eq!(storage.mem(&key_ax).unwrap(), true);
+        assert_eq!(storage.dirmem(&key_ab).unwrap(), false);
```
tizoc, line 1744: I am confused, shouldn't these be `true` ? `a/b/c/d` is being deleted, which should still keep the upper "directories" in place, or is there something I am missing?

tizoc: Oh, empty directories just shouldn't be there? I get it now.

shell/src/context_listener.rs
```diff
@@ -195,12 +205,54 @@ fn listen_protocol_events(
...
+        | ContextAction::Copy {tree_hash, ..}
+        | ContextAction::Delete {tree_hash, ..}
+        | ContextAction::RemoveRecursively {tree_hash, ..}
+        | ContextAction::Commit {tree_hash, ..}
+        | ContextAction::Fold {tree_hash, ..} => {
+            let mut hash:[u8;32] = [0;32];
+            tree_hash.clone().reader().read_exact(& mut hash).unwrap();
+            Some(hash)
+        }
+        _ => {None}
```
tizoc, line 226: Whenever possible I would list everything explicitly here instead of using a wildcard, makes it easier to avoid mistakes related to updating code when a new tag is added to an enum. When using wildcards to match the rest of the enum possibilities we lost the exhaustiveness check from the compiler.

shell/src/context_listener.rs
```diff
@@ -195,12 +205,54 @@ fn listen_protocol_events(
...
+    match &action{
+        ContextAction::Set {new_tree_hash, ..}
+        | ContextAction::Copy {new_tree_hash, ..}
+        | ContextAction::Delete {new_tree_hash, ..}
+        | ContextAction::RemoveRecursively {new_tree_hash, ..} => {
+            let mut hash:[u8;32] = [0;32];
+            new_tree_hash.clone().reader().read_exact(& mut hash).unwrap();
+            Some(hash)
+        }
+        _ => {None}
```
tizoc, line 240: Same here (re: enum tags and exhaustiveness checks)

shell/tests/apply_and_verify_context_actions_from_file.rs
```diff
@@ -0,0 +1,172 @@
...
+        .build()
+        .unwrap();
+
+    let header_stub = BlockHeaderWithHash::new(block_header_stub).unwrap();
+
+    let cache = Cache::new_lru_cache(128 * 1024 * 1024).unwrap(); // 128 MB
+    let commit_log_db_path = PathBuf::from("/tmp/commit_log/");
+    let key_value_db_path = PathBuf::from("/tmp/key_value_store/");
+    // let actions_storage_path = PathBuf::from("/mnt/hdd/milion.bin");
+    let actions_storage_path = PathBuf::from("/mnt/hdd/node-data/actionfile.bin");
```
tizoc, line 99: This will not work on other people's computers. @bkontur what should be used here? `OUT_DIR` ? (I think that is what is used in other tests)

tizoc: Oh wait, just noticed that the file is read, not written to, and also not generated by the build process or tests. Can it be a path related to the repo directory? something not absolute (and if not there, it should be reported with an error). Not sure how this is going to be handled on CI.

tizoc:
Ok, I don't think I have anything else to add for now, at least not until everything has been packed up together and rebased against `develop`. Maybe @bkontur has something to add.

storage/src/merkle_storage.rs
mat13mn: its some outdated comment

shell/tests/apply_and_verify_context_actions_from_file.rs
mat13mn: Right i forgot to disable it - the idea is to create network storage where this file will be accessible on CI - but not as a part of this PR ;)

storage/src/merkle_storage.rs
mat13mn: that gonna be optimized a little bit later ;)

PR #436, opened 2021-02-12
enable sanitization for tezos messages

mat13mn:
- New version for develop
- - run 'cargo check' once per OS - run 'cargo test' for every combination except macos with address sanitizer

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -45,18 +45,13 @@ jobs:
         run: brew install hidapi
         if: runner.os == 'macOS'
       - name: cargo check
+        if: matrix.sanitizer == 'none'
         uses: actions-rs/cargo@v1
         with:
           command: check
       - name: cargo test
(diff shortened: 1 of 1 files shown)


bkontur (approved):

bkontur:
@mat13mn 
rebase to develop and merge please

tizoc:
@mat13mn this one was approved seems to have been forgotten. Btw, can you also include the simplification we discussed? (using either `include` or `exclude` in the matrix to avoid the need to filter macos+address with an `if`)

bkontur:
@mat13mn @tizoc 
guys, what about this PR?

mat13mn:
there is still one comment from bruno that i need to adress - haven't had time yet - will do that soon

tizoc:
Closing this, will solve it as part of  #499 because I have to fix other things on the github actions file anyway because of the upgrade to ocaml-sys 0.20

PR #438, opened 2021-02-15
Report an error on wrong parameter

akoptelov:
Now error is reported instead of `panic`ing.


bkontur (approved):
@akoptelov 

PR #440, opened 2021-02-15
Replace lazy static with function [TE-327]

akoptelov:
@bkontur Can't use `const fn` here, but don't think this is a problem, since the value is used in production only once per session. So now this is a plain function.


bkontur (approved):
@akoptelov 
please rebase develop and merge

PR #443, opened 2021-02-15
Replace From with TryFrom [TE-246]

akoptelov:


bkontur (approved):
@akoptelov 
I see rustfmt check failed, please run `cargo fmt` before commit,

PR #439, opened 2021-02-15
[WIP] Merkle storage update

mat13mn:
- backtracking support
- configurable merkle storage backends (sled,in memory, rocksdb)
- configurable action storage backend (rocksdb, actionfile)
- using crypto::hash in merkle storage (and backends)
- using extended actions format (for backtracking)
- refactored statistics related code (merkle storage)
- integration test for applying actions from recorded actionfile

edit:
- still working on some failing tests

PR #444, opened 2021-02-15
Merkle storage update (rebased, was #439)

mat13mn:


storage/src/context_action_storage.rs
```diff
@@ -89,7 +89,7 @@ impl ContextActionStorage {
     #[inline]
     pub fn put_action(
         &mut self,
-        block_hash: &BlockHash,
+        block_hash: BlockHash,
```
tizoc, line 92: Why is this no longer a reference?

crypto/src/hash.rs
```diff
@@ -53,6 +54,9 @@ macro_rules! define_hash {
             std::cmp::Ord,
             std::hash::Hash,
         )]
+        // TODO: Hash -> Array, in current shape it doesnt seem to be
+        // super effective object is still allocated on heap and clone is
+        // costful, even
```
tizoc, line 59: Still has to be evaluated, but we have this: https://github.com/simplestaking/tezedge/pull/424

shell/src/context_listener.rs
```diff
@@ -3,21 +3,31 @@
 
 //! Listens for events from the `protocol_runner`.
 
+use bytes::{Buf, BufMut, BytesMut};
```
tizoc, line 6: There are many clippy warnings regarding unused and duplicated imports on this file

light_node/src/configuration.rs
```diff
@@ -850,11 +858,21 @@ impl Environment {
...
-                        .unwrap_or("true")
-                        .parse::<bool>()
-                        .expect("Provided value cannot be converted to bool"),
+                    action_store_backend: {
+                        let backend = args
+                            .value_of("actions-store-backend")
+                            .unwrap_or("none")
+                            .parse::<String>()
+                            .expect("Provided value cannot is not a string");
+                        match &backend[..] {
```
tizoc, line 867: `backend.as_str()` instead?

shell/tests/apply_and_verify_context_actions_from_file.rs
```diff
@@ -0,0 +1,194 @@
...
+                | ContextAction::Copy { .. }
+                | ContextAction::Delete { .. }
+                | ContextAction::RemoveRecursively { .. }
+                | ContextAction::Commit { .. }
+                | ContextAction::Checkout { .. } => {
+                    if let Err(e) = perform_context_action(&msg.action, &mut context) {
+                        panic!("cannot perform action {:?} error: '{}'", &msg, e);
+                    }
+                }
+                _ => {}
```
tizoc, line 184: Please, use an explicit enumeration here, we should avoid wildcard matches on enums whenever possible.

storage/Cargo.toml
```diff
@@ -12,21 +12,34 @@ derive_builder = "0.9"
 failure = "0.1"
 getset = "0.1"
 hex = "0.4"
+im = { version = "15.0.0", features = ["serde"] }
 itertools = "0.10"
 num_cpus = "1.13"
 rocksdb = {version = "0.15", features = ["snappy", "lz4", "zstd", "zlib"], default-features = false }
 serde = { version = "1.0", features = ["derive", "rc"] }
-slog = "2.7"
+slog = "2.5"
```
tizoc, line 20: Version is being lowered here,  but shouldn't

storage/scripts/actions_tool.rs
```diff
@@ -0,0 +1,257 @@
+use bytes::{Buf, BufMut, BytesMut};
+use std::error::Error;
+use std::fs::{File, OpenOptions};
+use std::io::{BufReader, Read, Seek, SeekFrom};
+use std::path::Path;
+// use bytes::buf::BufExt;
```
tizoc, line 6: The commented imports should be removed instead.

storage/scripts/actions_tool.rs
```diff
@@ -0,0 +1,257 @@
...
+
+#[derive(Clone, Copy, Debug)]
+pub struct ActionsFileHeader {
+    pub block_height: u32,
+    pub actions_count: u32,
+    pub block_count: u32,
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug)]
+pub enum ContextAction {
```
tizoc, line 28: This seems to be in the old format, right? I don't see any mention of `tree_hash` and `new_tree_hash`.

storage/scripts/merkle_storage_stats.rs
```diff
@@ -0,0 +1,251 @@
...
+            .arg(Arg::with_name("actions_file")
+                 .required(true)
+                 .help("path to the actions.bin")
+                 .index(1))
+            .arg(Arg::with_name("backend")
+                 .short("b")
+                 .long("backend")
+                 .default_value("in-memory-gced")
+                 .help("backend to use for storing merkle storage. Possible values: in-memory-gced, in-memory-mark-sweep-gced")
+                       // + "rocksdb, sled, in-memory-gced, in-memory-mark-sweep-gced")
```
tizoc, line 93: What is this comment for?

storage/scripts/merkle_storage_stats.rs
```diff
@@ -0,0 +1,251 @@
...
+            }
+
+            match &action {
+                ContextAction::Set {
+                    key,
+                    value,
+                    ignored,
+                    ..
+                } => {
+                    if !ignored {
```
tizoc, line 173: This looks outdated too, we don't have an `ignored` flag amymore.

crypto/src/hash.rs
```diff
@@ -68,6 +72,10 @@ macro_rules! define_hash {
                 }
             }
 
+            pub fn to_bytes(&self) -> &[u8] {
```
tizoc, line 75: Since we are borrowing, this should be `as_bytes` and not `to_bytes`

tizoc:
@binier @mat13mn @mambisi I left a bunch of comments, but there are also many clippy warnings, please address those that are related to the code each of you wrote.

Btw, I'm confused about `storage/scripts/actions_tool.rs` having it's own definitions of context actions, shouldn't it import the definitions that already exist? Those also don't include `tree_hash` and `new_tree_hash`, so it seems like things are out of sync. Is there anything am I missing?

binier:
> Btw, I'm confused about `storage/scripts/actions_tool.rs` having it's own definitions of context actions, shouldn't it import the definitions that already exist? Those also don't include `tree_hash` and `new_tree_hash`, so it seems like things are out of sync. Is there anything am I missing?

That's just to support the old actions.bin file. That duplicate definition should be removed so that script supports the new format.

storage/scripts/merkle_storage_stats.rs
binier: Right now that script only supports in-memory stores. It probably should support rocksdb and sled as well, but those implementations don't implement measurements for memory usage, so it won't create relevant statistics.

Probably we should just  replace it with todo comment.

storage/scripts/actions_tool.rs
binier: Yes its for old format.

tizoc:
@binier understood, thank you!

storage/scripts/merkle_storage_stats.rs
tizoc: Yes, please. Add a meaningful comment or remove that one if there is nothing to say.

crypto/src/hash.rs
mat13mn: my impression is that thread safe Arc as well as heap allocation might cause noticeable overhead...

crypto/src/hash.rs
mat13mn: will do

tizoc:
Closing this one, sone parts have been merged to develop already (#458 and #453) , others are being handled in #463 and #454

PR #445, opened 2021-02-15
Properly report lock poisoning errors [TE-332]

akoptelov:
The only exception is `resolve_node_from_request` that is used in
`warp::any` filter that should be infallible.


bkontur (approved):
@akoptelov 
please rebase/merge to develop

PR #446, opened 2021-02-16
[TE-420] Rework dev (explorer) blocks RPC

adonagy:


rpc/src/helpers.rs
```diff
@@ -630,6 +630,24 @@ pub(crate) fn get_prevalidators(
     Ok(prevalidators)
 }
 
+/// Struct to show in tezedge explorer to lower data flow
+#[derive(Serialize, Debug, Clone)]
+pub struct SlimBlockData {
+    pub level: i32,
+    pub hash: String,
```
bkontur, line 637: @adonagy 
hash -> block_hash

rpc/src/server/dev_handler.rs
```diff
@@ -42,14 +42,21 @@ pub async fn dev_blocks(
     };
     let limit = query.get_usize("limit").unwrap_or(50);
 
+    let blocks = base_services::get_blocks(
```
bkontur, line 45: @adonagy 
I would refactor get_blocks, to take mapping closure as parameter:
```
map(|(header, json_data)| map_header_and_json_to_full_block_info(header, json_data, &chain_id))
```
```
block_result
                .into_iter()
                .map(|block| block.into())
                .collect::<Vec<SlimBlockData>>()
```

or maybe easier to change get_blocks to return:
```
Vec<(BlockHeaderWithHash, BlockJsonData)>
```
and then map it, because now there is a big overhead:
`Vec<(BlockHeaderWithHash, BlockJsonData)>` -> is mapped to FullBlockInfo (there is a lots of hashes computation) -> and then mapped to SlimBlockData

```

bkontur (changes requested):
@adonagy 
please see comment about get_blocks

bkontur (approved):
@adonagy 
wow, very nice with that generic,
please rebase and merge to develop

adonagy:
@bkontur rebased, will merge after checks passed

PR #448, opened 2021-02-16
Use Result for blake2b functions [TE-277]

akoptelov:


crypto/src/crypto_box.rs
```diff
@@ -51,11 +51,13 @@ pub struct PublicKey(box_::PublicKey);
...
-    pub fn public_key_hash(&self) -> CryptoboxPublicKeyHash {
-        CryptoboxPublicKeyHash::try_from(crate::blake2b::digest_128(self.0.as_ref()))
-            .unwrap_or_else(|_| {
-                unreachable!("digest_128 result lenght should match CryptoboxPublicKeyHash")
-            })
+    pub fn public_key_hash(&self) -> Result<CryptoboxPublicKeyHash, Blake2bError> {
+        Ok(
+            CryptoboxPublicKeyHash::try_from(crate::blake2b::digest_128(self.0.as_ref())?)
+                .unwrap_or_else(|_| {
+                    unreachable!("digest_128 result lenght should match CryptoboxPublicKeyHash")
```
bkontur, line 58: @akoptelov 
what about this unreachable?
I just dont like them :D :D

crypto/src/proof_of_work.rs
```diff
@@ -95,11 +95,11 @@ impl ProofOfWork {
 // `check_proof_of_work(&received_raw_data[4..60], target)`
 pub fn check_proof_of_work(data: &[u8], target: f64) -> PowResult {
     let target_number = make_target(target);
-    check_proof_of_work_inner(data, &target_number)
+    check_proof_of_work_inner(data, &target_number).map_err(|_| ())
```
bkontur, line 98: @akoptelov @vlad9486 
this PowResult looks little bit hacky :D

can we do it here better with From trait:
`.map_err(PowResult::from)` instead of `.map_err(|_| ())` ?

bkontur: @akoptelov 
there is some compilation error,
please run `cargo bench` locally and see

bkontur (changes requested):
@akoptelov 
please check comments in PR

crypto/src/proof_of_work.rs
akoptelov: > @akoptelov @vlad9486
> this PowResult looks little bit hacky :D
> 
> can we do it here better with From trait:
> `.map_err(PowResult::from)` instead of `.map_err(|_| ())` ?

Hmm... we just don't need that at all.

And below we need to convert `Blake2bError` explicitly since `PowResult` is an alias for `Result<(), ()>`.

crypto/src/proof_of_work.rs
```diff
@@ -109,7 +109,7 @@ pub fn check_proof_of_work(data: &[u8], target: f64) -> PowResult {
 }
 
 fn check_proof_of_work_inner(data: &[u8], target_number: &BigUint) -> PowResult {
-    let hash = blake2b::digest_256(data);
+    let hash = blake2b::digest_256(data).map_err(|_| ())?;
```
bkontur, line 112: @akoptelov 
now when digest_256 returns Blake2bError,
I would add here also:
```
impl From<Blake2bError> for PowResult {
    fn from(...) -> Self {
...
    }
}
```

and then we can use: `blake2b::digest_256(data)?`

bkontur (changes requested):
@akoptelov 
please, check one comment and rebase to develop and merge

bkontur (approved):
@akoptelov 

PR #451, opened 2021-02-18
Cheaper identifier for identifying to which tree a context action is being applied

tizoc:
With this we will avoid the need to compute tree hashes all the time.

libtezos-ffi change: https://gitlab.com/simplestaking/tezos/-/merge_requests/18


tizoc:
Closing this one, will be part of #454

PR #452, opened 2021-02-18
Update limits basing on Tezos updates [TE-422, TE-405]

akoptelov:
- Limits are updated according to Tezos code
- Encoded data generator is added to test limits
- Now there is the `PeerMessageResponse.message` that contains exactly one specific message

@bkontur Please take a look at the `HISTORY_MAX_SIZE`. It is `200` despite that `CURRENT_BRANCH_HISTORY_MAX_LENGTH` is now `1000`. Should that also be updated?

This addresses following Jiras:
TE-422, TE-405, TE-407


tezos/encoding/src/binary_reader.rs
```diff
@@ -251,7 +251,7 @@ impl BinaryReader {
                 let mut buf_slice = safe!(buf, upper, buf.take(upper));
                 let res = self.decode_value(&mut buf_slice, inner_encoding);
                 match res {
-                    Err(BinaryReaderError::Underflow { bytes: _ }) => {
+                    Err(BinaryReaderError::Underflow { bytes: _ }) if upper == *max => {
```
tizoc, line 254: It is weird seeing things go from an underflow error to a "boundary exceeded" one. Reasoning about what the code is doing here makes it clear, but a comment explaining that in this case the inner underflow is caused by the outer-imposed limit would make it easier to understand.

tizoc: @akoptelov looks good, thanks

tezos/messages/src/p2p/encoding/current_branch.rs
```diff
@@ -16,7 +16,7 @@ use crate::p2p::encoding::block_header::BlockHeader;
 
 use super::limits::CURRENT_BRANCH_HISTORY_MAX_LENGTH;
 
-pub const HISTORY_MAX_SIZE: u8 = CURRENT_BRANCH_HISTORY_MAX_LENGTH as u8; // 200
+pub const HISTORY_MAX_SIZE: u8 = 200;
```
tizoc, line 19: if this is supposed to match `CURRENT_BRANCH_HISTORY_MAX_LENGTH`, can we add a static assert to verify that it remains true all the time?

I use this crate in ocaml-interop for that https://docs.rs/static_assertions/1.1.0/static_assertions/index.html

tizoc: Oh wait, I just saw the new value for `CURRENT_BRANCH_HISTORY_MAX_LENGTH`, ignore this comment.

tezos/messages/src/p2p/encoding/peer.rs
```diff
@@ -47,74 +49,80 @@ pub struct PeerMessageResponse {
...
-                Tag::new(
-                    0x60,
-                    "GetOperationsForBlocks",
-                    GetOperationsForBlocksMessage::encoding().clone(),
-                ),
-                Tag::new(
-                    0x61,
-                    "OperationsForBlocks",
-                    OperationsForBlocksMessage::encoding().clone(),
+    Encoding::bounded(
```
tizoc, line 52: @akoptelov this is the thing we discussed some time ago about a mismatch between our implementation representing this as a list while in OCaml it was a single dynamically-sized element, is that correct?

tizoc: What I am thinking is, why does this need to be a list at all? (I am not that familiar with our encoding library, so there may be details that I am missing)

akoptelov: Well, just limiting the list a very fast fix, with minimal changes. But I agree, replacing the list with single value is more clear and firm solution.

tizoc:
@akoptelov last change looks good, but something happened with the drone run and it failed.

tizoc:
Looks good to me.

@bkontur @akoptelov I'm thinking that I may be worth it to have a small OCaml program that just prints the values of all limits to easily detect when something changes. Probably as part of the idea I mentioned some time ago about using OCaml's PPX to produce parseable definitions of the encodings. I'm still dealing with storage integration related stuff, but it is something I can look into doing once I'm free from other things.

tezos/messages/src/p2p/encoding/current_branch.rs
bkontur, line 19: @akoptelov 
could we remove HISTORY_MAX_SIZE now?  CURRENT_BRANCH_HISTORY_MAX_LENGTH is what we want

bkontur (changes requested):
@akoptelov 
please to develop,
and check comment about  HISTORY_MAX_SIZE

tezos/messages/src/p2p/encoding/current_branch.rs
akoptelov: That constant is still used in Tezos code:
```
(* FIXME: this should not be hard-coded *)
let max_locator_size = 200
```

bkontur: @akoptelov 
I know, but we have now:
```
pub const HISTORY_MAX_SIZE: u8 = 200;
```
and 
```
pub const CURRENT_BRANCH_HISTORY_MAX_LENGTH: usize = 200;
```

it is the same, so I just want to remove HISTORY_MAX_SIZE and use CURRENT_BRANCH_HISTORY_MAX_LENGTH instead

akoptelov: Not exactly. The `CURRENT_BRANCH_HISTORY_MAX_LENGTH` is actually `1000`.

bkontur (approved):
@akoptelov 
wait for CI and please rebase and merge to develop

PR #458, opened 2021-02-19
Merkle storage multiple backends

mat13mn:


storage/src/storage_backend.rs
```diff
@@ -0,0 +1,181 @@
...
+        serializer: &mut dyn slog::Serializer,
+    ) -> slog::Result {
+        serializer.emit_arguments(key, &format_args!("{}", self))
+    }
+}
+
+pub trait StorageBackend: Send + Sync {
+    fn is_persisted(&self) -> bool;
+    fn get(&self, key: &EntryHash) -> Result<Option<ContextValue>, StorageBackendError>;
+    fn put(&mut self, key: EntryHash, value: ContextValue) -> Result<bool, StorageBackendError>;
```
tizoc, line 72: `put` and `merge` take ownership of the key while the others borrow from it, why is that?

mat13mn: fixed

storage/src/storage_backend.rs
```diff
@@ -0,0 +1,181 @@
...
+}
+
+pub trait StorageBackend: Send + Sync {
+    fn is_persisted(&self) -> bool;
+    fn get(&self, key: &EntryHash) -> Result<Option<ContextValue>, StorageBackendError>;
+    fn put(&mut self, key: &EntryHash, value: ContextValue) -> Result<bool, StorageBackendError>;
+    fn merge(&mut self, key: &EntryHash, value: ContextValue) -> Result<(), StorageBackendError>;
+    fn delete(&mut self, key: &EntryHash) -> Result<Option<ContextValue>, StorageBackendError>;
+    fn contains(&self, key: &EntryHash) -> Result<bool, StorageBackendError>;
+    fn get_mem_use_stats(&self) -> Result<RocksDBStats, StorageBackendError>;
```
tizoc, line 76: It is not always `RocksDBStats`, right? I saw that the `in_memory_storage` backend updates a different kind of stats, but always returns an empty `RocksDBStats` here

mat13mn: yes - the stats part is  its coupled with GCs(then return type is extended). I just checked and it looks like for some of the backends `get_stats` was not implemented on the feature branch.

storage/src/merkle_storage.rs
```diff
@@ -1163,11 +1174,11 @@ impl MerkleStorage {
     fn get_entries_recursively(
         &self,
         entry: &Entry,
-        batch: &mut WriteBatch,
+        batch: &mut Vec<(EntryHash, ContextValue)>,
     ) -> Result<(), MerkleError> {
         // add entry to batch
-        self.db
-            .put_batch(batch, &self.hash_entry(entry)?, &bincode::serialize(entry)?)?;
+        // self.db.put(self.hash_entry(entry)?, bincode::serialize(entry)?)?;
```
tizoc, line 1180: Can this comment be removed or was it for something that is pending?

storage/src/backend/mod.rs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) SimpleStaking and Tezedge Contributors
+// SPDX-License-Identifier: MIT/
```
bkontur, line 2: @mat13mn 
remove please just "/" at the end "MIT/" :)

storage/src/persistent/database.rs
```diff
@@ -123,6 +123,22 @@ pub trait KeyValueStoreWithSchema<S: KeyValueSchema> {
     fn get_mem_use_stats(&self) -> Result<RocksDBStats, DBError>;
 }
 
+pub trait GetInMemStats {
```
tizoc, line 126: @bkontur FYI this had to be separated to it's own trait because of some weird error in the type inferencer when the `get_mem_use_stats` method was called:

```
cannot infer type for type parameter `S` declared on the trait `KeyValueStoreWithSchema`
```

I wouldn't expect the inferencer to need that being that there are no parameters that relate to `S` in that method, but it seems it does.

storage/src/persistent/database.rs
```diff
@@ -123,6 +123,22 @@ pub trait KeyValueStoreWithSchema<S: KeyValueSchema> {
     fn get_mem_use_stats(&self) -> Result<RocksDBStats, DBError>;
```
tizoc, line 123: @mat13mn btw with the new trait for this method, it should be removed from here (and the duplicated implementation)

storage/src/storage_backend.rs
tizoc: @mat13mn sounds fine to me, and I guess those stats don't make as much sense as when the GC is involved. @bkontur  is that ok with you?

storage/src/storage_backend.rs
```diff
@@ -0,0 +1,181 @@
...
+        &self,
+        _record: &slog::Record,
+        key: slog::Key,
+        serializer: &mut dyn slog::Serializer,
+    ) -> slog::Result {
+        serializer.emit_arguments(key, &format_args!("{}", self))
+    }
+}
+
+pub trait StorageBackend: Send + Sync {
```
tizoc, line 69: Dumb question, but what is the difference between this and `KeyValueStoreWithSchema` ? just noticed that the interfaces are very similar.

mat13mn: AFAIK original author was not happy with some RocksDB types leaking into abstract interface - not really related to rocksdb.

mat13mn: ah right - also StorageBackend comes from GC part where it has additional methods like `start_new_cycle` `retain` etc. So probably we dont wont to infect `KeyValueStoreWithSchema` with those :thinking:

storage/src/merkle_storage.rs
```diff
@@ -1163,11 +1174,10 @@ impl MerkleStorage {
     fn get_entries_recursively(
         &self,
         entry: &Entry,
-        batch: &mut WriteBatch,
+        batch: &mut Vec<(EntryHash, ContextValue)>,
     ) -> Result<(), MerkleError> {
         // add entry to batch
-        self.db
-            .put_batch(batch, &self.hash_entry(entry)?, &bincode::serialize(entry)?)?;
```
tizoc, line 1170: Can this cause problems? I would guess that the RocksDB implementation for this is optimized, while the new approach isn't (at least not as much).

tizoc:
Other than the current clippy warnings about cloning arrays instead of using `*key` it looks ok to me, but a few things remaining:

- I don't get the difference between `StorageBackend` and `KeyValueStoreWithSchema`, they are similar, why do we need both?
- Only the RocksDB backend returns stats when asked.
- Some `write_batch` functionality from RocksDB was used in the old version, but has now been replaced by a custom implementation (that is compatible with all backends I assume). Can this be an issue? ([here](https://github.com/simplestaking/tezedge/pull/458/files#diff-8e32c1599e2683630b3d12ff50bfbc818afbd7ac8205c662925ba5dfd1b1743eR1180))
- Are the other stores being tested? I see that the `KVBACKEND` can be used to control what backend is used in the tests, but that is not being used right now, right?

I'm ok if the first two are addressed later, but I'm a bit worried about the last two.

storage/src/storage_backend.rs
tizoc: Make sense. But it is weird to have both I guess things have to be unified/reworked at some point (maybe adding extra traits in addition to `KeyValueStoreWithSchema`), having this duplication is suspicious.

mat13mn:
>Some write_batch functionality from RocksDB was used in the old version, but has now been replaced by a custom implementation (that is compatible with all backends I assume). Can this be an issue?

just checked and docs says that write_batch is atomic write - i`ll try update `StorageBackend` interface to make use of that

>Are the other stores being tested? I see that the KVBACKEND can be used to control what backend is used in the tests, but that is not being used right now, right? i can try to integrate that into our github pipline somehow

storage/src/persistent/database.rs
bkontur: @tizoc 
ok, lets get it working as it is, and we can next week clean this up,
just please do some jira with TODOs for that

tizoc:
Update on the tests: I'm adding a macro to expand to calls to all same tests but with each of the different backends. @mat13mn has tested them locally using the env var, but we still want them on CI.

storage/src/persistent/database.rs
tizoc: @bkontur ok, will do that👍

tizoc: This is probably also related to the other duplication issue I mentioned, we have two traits that look very similar.

storage/src/storage_backend.rs
bkontur: @tizoc @mat13mn 
ok, just add here `TODO: TE-XYZ <some-desc-what-should-be-done>` please,
we can do that later

storage/src/merkle_storage.rs
mat13mn: i have already exteneded trait and made use of write batch :+1:

storage/src/storage_backend.rs
mat13mn: done

storage/src/storage_backend.rs
mat13mn: https://simplestakingcom.atlassian.net/browse/TE-432

storage/src/persistent/database.rs
mat13mn: https://simplestakingcom.atlassian.net/browse/TE-432

storage/src/storage_backend.rs
mat13mn: https://simplestakingcom.atlassian.net/browse/TE-432

PR #463, opened 2021-02-23
Merkle storage gc backends

mat13mn:
- fixes data race in mark&move gc (https://simplestakingcom.atlassian.net/browse/TE-426)
- common memory usage statistics for all the backends (https://simplestakingcom.atlassian.net/browse/TE-431)
- single abstraction over all merkle storage backends (https://simplestakingcom.atlassian.net/browse/TE-432)
- GC implementations `mark&sweep` and `mark&move` with interface shared with other backends
- tests for MerkleStorageBackend for all the backends implementations

- [ ] dev stats enabled check
- [ ] context is_commited refactor
- [ ] // TODO: move to src/context/mod.rs
- [ ] TODO TE-440 subscribe to commit event - change


mat13mn:
the one last thing missing is triggering GC from MerkleStorage. There is proper trait created already i`ll add that probably tommorow morning.

shell/src/context_listener.rs
```diff
@@ -216,6 +216,13 @@ fn listen_protocol_events(
                 }
 
                 perform_context_action(&action, context)?;
+                // below logic should be driven by dedicated ContextAction events
```
bkontur, line 219: @mat13mn
is it possible to trigger this block_applied/cycle_started logic just for MarkMoveInMem/MarkSweepInMem??

storage/src/backend/btree_map.rs
```diff
@@ -22,59 +26,121 @@ impl<K: Ord, V> Default for KVStore<K, V> {
...
 
+    fn total_get_mem_usage(&self) -> Result<usize, DBError> {
+        Ok(self.stats.read().unwrap().total_as_bytes())
+    }
+
     /// put kv in map if key doesn't exist. If it does then return false.
-    fn put(&mut self, key: &EntryHash, value: ContextValue) -> Result<bool, KVStoreError> {
-        match self.kv_map.entry(key.clone()) {
+    fn put(&self, key: &EntryHash, value: &ContextValue) -> Result<(), DBError> {
+        match self.kv_map.write().unwrap().entry(*key) {
```
bkontur, line 59: @mat13mn 
please no unwraps, just map PoisonError to DBError, e.g.: 
```
impl<T> From<PoisonError<T>> for StateError {
    fn from(pe: PoisonError<T>) -> Self {
        StateError::LockError {
            reason: format!("{}", pe),
        }
    }
}
```

storage/src/backend/in_memory_backend.rs
```diff
@@ -91,21 +138,29 @@ impl StorageBackend for InMemoryBackend {
...
-        })
+    fn write_batch(&self, batch: Vec<(EntryHash, ContextValue)>) -> Result<(), DBError> {
+        for (k, v) in batch {
+            self.merge(&k, &v)?;
+        }
+        Ok(())
+    }
+
+    fn total_get_mem_usage(&self) -> Result<usize, DBError> {
+        let r = self.inner.read().map_err(|e| DBError::GuardPoison {
```
bkontur, line 157: @mat13mn 
please, unify please this PoisonError handling like I mentioned

storage/src/backend/rocksdb_backend.rs
```diff
@@ -43,101 +29,53 @@ pub struct RocksDBBackendStats {
...
-                name: &self.column_name,
-            },
-        )?;
-
-        self.inner
-            .delete_cf_opt(cf, &key.as_ref()[..], &Self::default_write_options())
-            .map_err(StorageBackendError::from)?;
-        Ok(None)
+    fn contains(&self, key: &EntryHash) -> Result<bool, DBError> {
+        (self.inner.deref() as &dyn KeyValueStoreBackend<MerkleStorage>).contains(key)
```
bkontur, line 67: @mat13mn 
I would wrap this to some fn or macro, something like that, if possible
```
fn merkle() ->&dyn KeyValueStoreBackend<MerkleStorage> {
  self.inner.deref() as &dyn KeyValueStoreBackend<MerkleStorage>
}
```

storage/src/backend/rocksdb_backend.rs
```diff
@@ -43,101 +29,53 @@ pub struct RocksDBBackendStats {
...
-        self.inner
-            .merge_cf_opt(
-                cf,
-                &key.as_ref()[..],
-                &value,
-                &Self::default_write_options(),
-            )
-            .map_err(StorageBackendError::from)
+    fn get(&self, key: &EntryHash) -> Result<Option<ContextValue>, DBError> {
+        (self.inner.deref() as &dyn KeyValueStoreBackend<MerkleStorage>).get(key)
```
bkontur, line 63: @mat13mn 
we almost never used deref() directly, I just saw few usages in merkle_storage stuff,
is not here any other solution without calling deref() ?

storage/src/persistent/database.rs
```diff
@@ -221,25 +254,49 @@ impl<S: KeyValueSchema> KeyValueStoreWithSchema<S> for DB {
...
 
+        self.write_opt(rocksb_batch, &default_write_options())?;
         Ok(())
     }
 
-    fn write_batch(&self, batch: WriteBatch) -> Result<(), DBError> {
-        self.write_opt(batch, &default_write_options())?;
+    fn total_get_mem_usage(&self) -> Result<usize, DBError> {
+        let memory_usage_stats = rocksdb::perf::get_memory_usage_stats(Some(&[&self]), None)?;
+        return Ok((memory_usage_stats.mem_table_total
```
bkontur, line 275: @mat13mn 
clippy says you dont need to use `Result<usize, DBError>` -> `usize`

bkontur (changes requested):
@mat13mn 
please check comments and clippy, which is easy

shell/src/context_listener.rs
mat13mn: for backends other then markmove/mark sweep it does nothing

tizoc: This is a good candidate for a commit-event + subscribers logic we discussed earlier today. Maybe not viable as part of this PR, but please add a TODO with TE-440 (which I just created for that task, still have to fill the description).

storage/src/backend/rocksdb_backend.rs
```diff
@@ -43,101 +29,53 @@ pub struct RocksDBBackendStats {
...
-        self.inner
-            .put_cf_opt(
-                cf,
-                &key.as_ref()[..],
-                &value,
-                &Self::default_write_options(),
-            )
-            .map_err(StorageBackendError::from)
-            .map(|_| true)
+impl GarbageCollector for RocksDBBackend {
```
tizoc, line 32: Is it possible to instead of doing this explicitly for each store, have a "NotGarbageCollected" marker trait to which this empty GC logic is implemented for? And stores that don't have GC just impl that trait.

```rust
impl NotGarbageCollected for RocksDBBackend {}
```

storage/src/merkle_storage.rs
```diff
@@ -936,17 +939,25 @@ impl MerkleStorage {
...
-        let mut batch: Vec<(EntryHash, ContextValue)> = Vec::new();
-
-        // build list of entries to be persisted
-        self.get_entries_recursively(entry, &mut batch)?;
-
-        // write all entries at once (depends on backend)
-        self.db.put_batch(batch)?;
+    /// Marks all the entries from last commit as used
+    /// so GC can know when to remove them
+    pub fn mark_entries_from_last_commit_as_used(&mut self) -> Result<(), MerkleError> {
```
tizoc, line 944: This is probably a good candidate for being moved elsewhere once the "commit event" has been implemented (TE-440). I find it smelly that this logic is in `MerkleStorage`.

storage/src/merkle_storage_stats.rs
```diff
@@ -0,0 +1,256 @@
...
+}
+
+pub struct StatUpdater<'a> {
+    stats: &'a mut MerkleStorageStatistics,
+    timer: Instant,
+    action: MerkleStorageAction,
+    key: Option<String>,
+}
+
+impl<'a> Drop for StatUpdater<'a> {
```
tizoc, line 183: Cool. Just in case, we want these stats even if the operation fails, right? (which is unlike the previous implementation)

storage/src/merkle_storage_stats.rs
```diff
@@ -0,0 +1,256 @@
...
+    }
+}
+
+impl<'a> StatUpdater<'a> {
+    pub fn new(
+        stats: &'a mut MerkleStorageStatistics,
+        action: MerkleStorageAction,
+        action_key: Option<&Vec<String>>,
+    ) -> Self {
+        let key = match action_key {
```
tizoc, line 195: @mat13mn What happens with the rest of the path here?

Shorter but not necessarily better versions:

```rust
    let key = action_key
        .filter(|path| path.len() > 1 && path[0] == "data")
        .map(|path| path[1].to_string());
```

or

```rust
    let key = match action_key {
        Some(path) if path.len() > 1 && path[0] == "data" => Some(path[1].to_string()),
        Some(_) | None => None,
    };
```

storage/src/persistent/database.rs
```diff
@@ -82,61 +133,72 @@ pub trait KeyValueStoreWithSchema<S: KeyValueSchema> {
...
-    /// * `key` - Value of key specified by schema
-    /// * `value` - Value to be inserted associated with given key, specified by schema
-    fn put_batch(
-        &self,
-        batch: &mut WriteBatch,
-        key: &S::Key,
-        value: &S::Value,
-    ) -> Result<(), DBError>;
+    /// * `key` - Key (specified by schema), to be checked for existence
+    fn retain(&self, predicate: &dyn Fn(&S::Key) -> bool) -> Result<(), DBError>;
```
tizoc, line 146: I am confused by this name, it is called `retain` but the description says it just checks if a given key is there.

storage/src/persistent/mod.rs
```diff
@@ -31,6 +32,8 @@ pub mod database;
 pub mod schema;
 pub mod sequence;
 
+const PRESERVE_CYCLE_COUNT: usize = 7;
```
tizoc, line 35: If we have a source for this number it would be really good to link it here.

storage/src/persistent/mod.rs
```diff
@@ -153,24 +156,32 @@ impl PersistentStorage {
...
-            KeyValueStoreBackend::BTreeMap => MerkleStorage::new(Box::new(BTreeMapBackend::new())),
-        };
+        let merkle =
+            match merkle_backend {
+                KeyValueStoreBackend::RocksDB => {
+                    MerkleStorage::new(Box::new(RocksDBBackend::new(db_context.clone())))
+                }
+                KeyValueStoreBackend::InMem => MerkleStorage::new(Box::new(InMemoryBackend::new())),
+                KeyValueStoreBackend::Sled { path } => {
+                    let sled = sled::Config::new().path(path).open().unwrap();
```
tizoc, line 166: Is there anything that can be done about this `unwrap` here?

storage/src/persistent/database.rs
tizoc: Ohhhh, it is the same comment as `contains` above. This one needs to be updated.

storage/src/backend/rocksdb_backend.rs
mat13mn: i couldnt figure out anything else ;/

tizoc: Another thing about this that I find confusing, is that the methods on `inner` return `rocksdb::Error`, but `DBError` is expected here.

tizoc: Ok I think I now see why all this is needed: the `KeyValueStoreBackend` defines methods with the same names that RocksDB uses in `DB`, and that makes things ambiguous. I think `KeyValueStoreBackend` should be implemented directly over a wrapper over `DB` (newtype pattern), and not for `DB` itself (and probably good for the other trait implementations too).

What do you think?

mat13mn: Thing is KeyValueStoreBackend is former `KeyValueStoreWithSchema` that is widely used in our codebase. One of the comment on previous PR was why to create separate trait(`StorageBackend`) as there is one already defined. What i did i splitted old `KeyValueStoreWithSchema` into 2 traits `KeyValueStoreBackend` and `KeyValueStoreWithSchemaIterator`. And created alias for old name 

```pub trait KeyValueStoreWithSchema<S: KeyValueSchema>:
    KeyValueStoreBackend<S> + KeyValueStoreWithSchemaIterator<S>
{                                        
}                                                   
```
to preserve backward compatibility. Then i just implemented already existing `KeyValueStoreBackend` for other types( inmem, btree, etc). I didnt want to mix old `DB` type with other merkle storage backends so i wrapped DB into RocksDBBackend and just forwarded all the method calls for that one. Does it make sense ?

storage/src/backend/rocksdb_backend.rs
mat13mn: i ended up using default empty impl;/ what i tried is (https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=6ccadea5565853fe30c6ba5a0a038d05) but it doesnt seem to work way i expected it to work

tizoc: @mat13mn I was thinking of something like (using your example):

```rust
impl<T: DefaultFoo> Foo for T {
    fn blah(&self){}
}
```

I don't see why `dyn` is needed.

storage/src/backend/rocksdb_backend.rs
tizoc: Yes, but there is still an implementation of that trait `impl<S: KeyValueSchema> KeyValueStoreWithSchemaIterator<S> for DB`, which is what I say I think is causing problems. If instead of implementing anything `for DB` you instead create a new wrapper type and move that implementation over there (so that it is not implemented `for DB` anymore), the ambiguity (which is what I think is causing problems) goes away.

tizoc: That applies to all traits we implement `for DB`.

storage/src/backend/rocksdb_backend.rs
mat13mn: yeah - for some reason compiler wanted it to be `dyn` in my non templated/generic impl (which was bad of course)

```
 --> src/main.rs:7:14
  |
7 | impl Foo for DefaultFoo{
  |              ^^^^^^^^^^ help: use `dyn`: `dyn DefaultFoo`
  |
  = note: `#[warn(bare_trait_objects)]` on by default
```

tizoc: Yeah, because in that case you are trying to implement it for a trait, and not an actual type, in my example I'm implementing it for any type that implements the `DefaultFoo` trait.

storage/src/merkle_storage_stats.rs
mat13mn: >@mat13mn What happens with the rest of the path here?

the idea was to store information about the first key only - something like `max-depth=1` parameter for `find`

storage/src/context.rs
```diff
@@ -290,6 +294,21 @@ impl ContextApi for TezedgeContext {
         let merkle = self.merkle.read().expect("lock poisoning");
         merkle.get_staged_root_hash()
     }
+
+    fn block_applied(&self) -> Result<(), ContextError> {
+        let mut merkle = self.merkle.write().expect("lock poisoning");
```
bkontur, line 299: @mat13mn 
please change expect to error handling,

anyway, do we need to do anything here?
if we do not support gc here now, I would just leaeve it as `Ok(())` to avoid unnecessecery locking

storage/src/context.rs
```diff
@@ -290,6 +294,21 @@ impl ContextApi for TezedgeContext {
...
+        Ok(merkle.block_applied()?)
+    }
+
+    fn cycle_started(&self) -> Result<(), ContextError> {
+        let mut merkle = self.merkle.write().expect("lock poisoning");
+        Ok(merkle.start_new_cycle()?)
+    }
+
+    fn get_memory_usage(&self) -> Result<usize, ContextError> {
+        let merkle = self.merkle.write().expect("lock poisoning");
```
bkontur, line 309: @mat13mn 
please change expect to error handling

storage/src/backend/rocksdb_backend.rs
```diff
@@ -2,36 +2,26 @@
...
     }
-}
 
-impl RocksDBBackend {
-    fn default_write_options() -> WriteOptions {
-        let mut opts = WriteOptions::default();
-        opts.set_sync(false);
-        opts
+    fn merkle_ref(&self) -> &dyn KeyValueStoreBackend<MerkleStorage> {
+        self.inner.deref() as &dyn KeyValueStoreBackend<MerkleStorage>
```
tizoc, line 24: @mat13mn what prevents the creation of a newtype to avoid the use of `DB` directly? this here just hides the problem under the rug instead of fixing the ambiguity. I see a few other places where `as &dyn` casts are still being made.

storage/src/persistent/mod.rs
tizoc: @mat13mn this one was marked as resolved but I don't see any change. Since the return type of this function is `Result<...>`, I think this can be mapped to an error.

storage/src/backend/rocksdb_backend.rs
tizoc: @mat13mn @bkontur although I am commenting here to understand why things are currently like that, this refactoring I'm suggesting here (and on the older comment) is probably not to be made as part of this PR because there is a lot going on already. Just trying to understand if there is anything that prevents us from taking the cleaner approach.

bkontur: @tizoc @mat13mn 
I am not sure about that dyn stuff, and we use it in different place,
last week I create anther jira which is related to this: https://simplestakingcom.atlassian.net/browse/TE-437
we can do it as as part of this jira,

just pls. add to the description, what we should look at

storage/src/persistent/database.rs
```diff
@@ -221,25 +265,49 @@ impl<S: KeyValueSchema> KeyValueStoreWithSchema<S> for DB {
...
 
+        self.write_opt(rocksb_batch, &default_write_options())?;
         Ok(())
     }
 
-    fn write_batch(&self, batch: WriteBatch) -> Result<(), DBError> {
-        self.write_opt(batch, &default_write_options())?;
+    fn total_get_mem_usage(&self) -> Result<usize, DBError> {
+        let memory_usage_stats = rocksdb::perf::get_memory_usage_stats(Some(&[&self]), None)?;
+        Ok((memory_usage_stats.mem_table_total
```
akoptelov, line 286: Is it possible to get an integer overflow?

storage/src/persistent/database.rs
```diff
@@ -221,25 +265,49 @@ impl<S: KeyValueSchema> KeyValueStoreWithSchema<S> for DB {
...
+            + memory_usage_stats.mem_table_unflushed
+            + memory_usage_stats.mem_table_readers_total
+            + memory_usage_stats.cache_total) as usize)
+    }
+
+    fn retain(&self, predicate: &dyn Fn(&S::Key) -> bool) -> Result<(), DBError> {
+        let garbage: Vec<_> = (self as &dyn KeyValueStoreWithSchemaIterator<S>)
+            .iterator(IteratorMode::Start)
+            .unwrap()
```
akoptelov, line 295: Please remove this `unwrap()`.

akoptelov (changes requested):
There are some `unwrap()`s still. Please remove them.

Also there is a place with possible integer overflow. @pventuzelo would recommend using `checked_add()` for such cases.

storage/src/persistent/mod.rs
mat13mn: ouch - i changed the return type but forgot to remove unwrap ...

storage/src/context.rs
mat13mn: we support GC but its implementation detail of KV backend passed to merkle storage so at this point we dont know if its supported or not

storage/src/backend/rocksdb_backend.rs
mat13mn: added todo note TE-437

mat13mn:
i`ll squash and sign commit before delivery

storage/src/persistent/database.rs
```diff
@@ -29,6 +32,20 @@ pub enum DBError {
     MissingColumnFamily { name: &'static str },
     #[fail(display = "Database incompatibility {}", name)]
     DatabaseIncompatibility { name: String },
+    #[fail(display = "Value already exists {}", key)]
+    ValueExists { key: String },
+    #[fail(display = "Guard Poison {} ", error)]
+    GuardPoison { error: String },
```
bkontur, line 38: @mat13mn 
what is difference between GuardPoison  vs LockError?
I think it is the same, could we just have just LockError?

storage/src/persistent/database.rs
```diff
@@ -82,61 +155,72 @@ pub trait KeyValueStoreWithSchema<S: KeyValueSchema> {
...
+
+        Ok(IteratorWithSchema(
+            self.prefix_iterator_cf(cf, key),
+            PhantomData,
+        ))
     }
 }
 
-impl<S: KeyValueSchema> KeyValueStoreWithSchema<S> for DB {
+impl<S: KeyValueSchema> KeyValueStoreWithSchema<S> for DB {}
```
bkontur, line 217: @mat13mn 
do we need this empty line?
`impl<S: KeyValueSchema> KeyValueStoreWithSchema<S> for DB {}`

storage/src/persistent/mod.rs
bkontur: @mat13mn 
please add here TODO with jira for real integartion to GC trigger

shell/scripts/storage_stats.rs
```diff
@@ -0,0 +1,468 @@
...
+    let _ = fs::remove_dir_all(&out_dir);
+    let _ = fs::create_dir(&out_dir);
+
+    let logger = create_logger();
+
+    let kvbackend = match params.backend.as_str() {
+        "rocksdb" => {
+            let cache = Cache::new_lru_cache(128 * 1024 * 1024).unwrap(); // 128 MB
+            let kv = create_key_value_store(&key_value_db_path, &cache);
+            MerkleStorage::new(Box::new(RocksDBBackend::new(kv.clone())))
```
bkontur, line 346: @mat13mn 
clippy says to remove clone `kv.clone()`

shell/scripts/storage_stats.rs
```diff
@@ -0,0 +1,468 @@
...
+
+fn main() -> Result<(), Error> {
+    let params = Args::read_args();
+
+    let out_dir = PathBuf::from(params.output.as_str());
+    let key_value_db_path = out_dir.join("key_value_store");
+
+    let actions_storage_path = PathBuf::from(params.input.as_str());
+
+    let _ = fs::remove_dir_all(&out_dir);
```
bkontur, line 337: @mat13mn 
this i suppressed error,
please do at least: `let _ = fs::remove_dir_all(&out_dir)?;`

bkontur:
@mat13mn 
please check clippy:
https://github.com/simplestaking/tezedge/pull/463/files#diff-d54fef0fdda2f7c092f3b92f132d96f8818ffe891a2614e35cde17d9e4e77084R482

https://github.com/simplestaking/tezedge/pull/463/files#diff-11179d6fb54b787df03f4cb1f394f2c68799d324ad686086a6f088988cb0de6eR233

https://github.com/simplestaking/tezedge/pull/463/files#diff-11179d6fb54b787df03f4cb1f394f2c68799d324ad686086a6f088988cb0de6eR265

storage/src/context.rs
```diff
@@ -290,22 +270,73 @@ impl ContextApi for TezedgeContext {
...
 impl TezedgeContext {
-    pub fn new(block_storage: BlockStorage, merkle: Arc<RwLock<MerkleStorage>>) -> Self {
+    pub fn new(block_storage: Option<BlockStorage>, merkle: Arc<RwLock<MerkleStorage>>) -> Self {
         TezedgeContext {
             block_storage,
             merkle,
         }
     }
+
+    fn associate_block_and_context_hash(
```
bkontur, line 305: @mat13mn @tizoc @jurajselep 
ok, so last thing, lets discuss and finally clean this old-hack: https://simplestakingcom.atlassian.net/browse/TE-447

bkontur (changes requested):
@mat13mn 
please, check comments

bkontur:
closing this, https://github.com/simplestaking/tezedge/pull/508 will be merged in few minutes

PR #467, opened 2021-02-25
Apply non mutating actions in integration test & set output path for 'sled' kv backend

mat13mn:


light_node/src/configuration.rs
```diff
@@ -939,6 +939,13 @@ impl Environment {
                     |value| {
                         value
                             .parse::<KeyValueStoreBackend>()
+                            .map( |v| {
```
mat13mn, line 942: @bkontur not sure if that can be handled in a better way to still fit into that nice refactoring you just made ...

shell/tests/apply_and_verify_context_actions_from_file.rs
```diff
@@ -134,16 +134,20 @@ fn feed_tezedge_context_with_actions() -> Result<(), Error> {
     let header_stub = BlockHeaderWithHash::new(block_header_stub).unwrap();
 
     let cache = Cache::new_lru_cache(128 * 1024 * 1024).unwrap(); // 128 MB
-    let commit_log_db_path = PathBuf::from("/tmp/commit_log/");
-    let key_value_db_path = PathBuf::from("/tmp/key_value_store/");
 
+    let out_dir_env = env::var("OUT_DIR")
```
bkontur, line 138: @mat13mn 
this is unrelated, but could you please all merkle_storage test change to create dirs to `env::var("OUT_DIR")` ?

bkontur (approved):

shell/tests/apply_and_verify_context_actions_from_file.rs
mat13mn: sure - i will check that

light_node/src/configuration.rs
bkontur: > 
> 
> @bkontur not sure if that can be handled in a better way to still fit into that nice refactoring you just made ...

I guess, this is the correct place to handle this

shell/tests/apply_and_verify_context_actions_from_file.rs
mat13mn: just run the tests and everything seems file - that has been addressed in GC PR i guess

PR #469, opened 2021-02-26
Add context information to reader/writer [TE-436]

akoptelov:
Now messages from both binary encoding and decoding are supplied with the context information, field/list element. I think that will be enough to locate problematic place in values/bytes.


bkontur (approved):
@akoptelov 
I guess, looks ok,

PR #471, opened 2021-02-27
New commit log

mambisi:


bkontur:
@mambisi 
I would like to move all the commit log stuff to directory, please:
`storage/src/commit_log/*`

this `storage/src/persistent/* ` was dedicated more to common/api/traits for storage

mambisi:
Ohk

storage/src/persistent/commit_log/mod.rs
```diff
@@ -16,23 +15,185 @@ use crate::persistent::schema::{CommitLogDescriptor, CommitLogSchema};
...
+        }
+        return buf;
+    }
+}
+
+pub struct CommitLog {
+    path: PathBuf,
+}
+
+impl CommitLog {
```
bkontur, line 141: @mambisi 
could you, please, provide also delete function, if possible?

storage/src/persistent/commit_log/mod.rs
```diff
@@ -16,23 +15,185 @@ use crate::persistent::schema::{CommitLogDescriptor, CommitLogSchema};
...
+    #[inline]
+    pub fn append_msg<B: AsRef<[u8]>>(&mut self, payload: B) -> Result<u64, TezedgeCommitLogError> {
+        let mut writer = Writer::new(&self.path)?;
+        let offset = writer.write(payload.as_ref())?;
+        Ok(offset)
+    }
+
+    #[inline]
+    pub fn read(&self, from: usize, limit: usize) -> Result<MessageSet, TezedgeCommitLogError> {
+        let mut reader = Reader::new(&self.path)?;
```
bkontur, line 156: @mambisi 
do we need to create new instances `Reader::new/Writer:new`, all the time? it does lots of things in `new`

mambisi: No it doesn't but the BufWriter need to be drop for changes to take effect, we can make on reader, i will try it out

tizoc: @mambisi isn't calling `flush()` on the writer enough? The drop just falls `flush_buf` which is also called by `flush` https://doc.rust-lang.org/src/std/io/buffered/bufwriter.rs.html#414-421

mambisi: I think it will work

storage/src/persistent/commit_log/mod.rs
```diff
@@ -16,23 +15,185 @@ use crate::persistent::schema::{CommitLogDescriptor, CommitLogSchema};
...
+
+impl Iterator for MessageSet {
+    type Item = Message;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let index = self.indexes.get(self.cursor)?;
+        let data = &self.buf[self.acc..(self.acc + index.uncompressed_data_length as usize)];
+        self.acc += index.uncompressed_data_length as usize;
+        self.cursor += 1;
+        Some(data.to_vec())
```
tizoc, line 54: I don't know how performance sensitive is this code, so maybe this doesn't matter: we are allocating a new vector each time here, is that necessary?

storage/src/persistent/commit_log/mod.rs
```diff
@@ -16,23 +15,185 @@ use crate::persistent::schema::{CommitLogDescriptor, CommitLogSchema};
...
+    fn new(position: u64, compressed_data_length: u64, uncompressed_data_length: u64) -> Self {
+        Self {
+            position,
+            compressed_data_length,
+            uncompressed_data_length,
+        }
+    }
+    fn to_vec(&self) -> Vec<u8> {
+        let mut buf = vec![];
+        for b in &self.position.to_be_bytes() {
```
tizoc, line 124: Shorter and probably faster:

```rust
 buf.extend_from_slice(&self.position.to_be_bytes());
```

mambisi: Yeah, completely forgot about that function

mambisi: Thanks

storage/src/persistent/commit_log/mod.rs
tizoc: Found https://docs.rs/streaming-iterator/ which would solve the issue.

But as I said, if performance is not an issue here, then it doesn't matter.

PR #475, opened 2021-03-03
[TE-441] Fix context hash tree hashing

tizoc:


tizoc:
The current implementation works and is clean enough, but I think things could be abstracted a bit more by adding a `Prehash` trait with a `prehash(self, writable)` method that pushes it's own bytes into the required format into any writable (of which the hasher is, but this would work for outputting the bytes to anything).

storage/src/hash/ocaml.rs
```diff
@@ -0,0 +1,80 @@
...
+    //   https://github.com/ocaml/ocaml/blob/a5f63ba65fb8df7b2fa52076a2763b74078e463e/runtime/hash.c#L41-L46
+    h ^= h >> 16;
+    h *= Wrapping(0x85ebca6b);
+    h ^= h >> 13;
+    h *= Wrapping(0xc2b2ae35);
+    h ^= h >> 16;
+
+    // Fold result to the range [0, 2^30-1] so that it is a nonnegative
+    // OCaml integer both on 32 and 64-bit platforms.
+    return h.0 & 0x3FFFFFFF;
```
bkontur, line 62: @tizoc
clippy say unnecessery return, just: `h.0 & 0x3FFFFFFF`

storage/src/hash/mod.rs
```diff
@@ -0,0 +1,444 @@
...
+mod ocaml;
+
+use std::{array::TryFromSliceError, convert::TryInto, io};
+
+use blake2::digest::{InvalidOutputSize, Update, VariableOutput};
+use blake2::VarBlake2b;
+use failure::Fail;
+use ocaml::ocaml_hash_string;
+
+use crate::merkle_storage::{Commit, ContextValue, Entry, Node, NodeKind, Tree};
```
bkontur, line 13: @tizoc 
we need to clean this little bit more later,
but for this PR, could you please move this hash dir to another subdir `context`, to look like:
```
storage/src/context/hash/mod.rs 
```

later with Mateusz PR with GC, we will move: `context.rs` and `merkle_storage.rs` also to `storage/src/context`,
and those common stuff: `use crate::merkle_storage::{Commit, ContextValue, Entry, Node, NodeKind, Tree};` we could move for example to: `storage/src/context/mod.rs`

and very later, my intention is to split the whole storage library to smaller dedicated libraries - I guess this will also simplify ffi context integration - we will see later

storage/src/hash/mod.rs
```diff
@@ -0,0 +1,444 @@
+// Copyright (c) SimpleStaking and Tezedge Contributors
+// SPDX-License-Identifier: MIT
```
bkontur, line 2: @tizoc 
can you please put here some short description at the begining with link https://hackmd.io/DnsQrkXCRLmHA3xzLiFgGw?view#Trees ?

storage/src/hash/mod.rs
```diff
@@ -0,0 +1,444 @@
...
+// uses BLAKE2 binary 256 length hash function
+// hash is calculated as:
+// <number of child nodes (8 bytes)><CHILD NODE>
+// where:
+// - CHILD NODE - <NODE TYPE><length of string (1 byte)><string/path bytes><length of hash (8bytes)><hash bytes>
+// - NODE TYPE - leaf node(0xff0000000000000000) or internal node (0x0000000000000000)
+pub(crate) fn hash_tree(tree: &Tree) -> Result<EntryHash, HashingError> {
+    // If there are >256 entries, we need to partition the tree and hash the resulting inode
+    if tree.len() > 256 {
+        let entries: Vec<(&String, &Node)> = tree.iter().collect();
```
bkontur, line 190: @tizoc 
Bruno, this worries me a lots, like you mentioned upper: `IMPORTANT: entries must be sorted in lexicographic order of the name`

what if we change signature: `tree: &Tree` to something like `Iterator + Ord`
in case somebody could change this
`type Tree = im::OrdMap<String, Node>;`

or at least for now please add `IMPORTANT:` comment to this `type Tree` about ordering for `partition_entries`


or explicit sorting here, will add a lots of overhead?
what do you thing any idea?

storage/src/hash/mod.rs
tizoc: Ok, but we will have to update that link later once they publish the actual document, since that one is a temporary one where they were doing the editing.

bkontur: @tizoc 
aha, ok, I thought that it is finally

storage/src/hash/mod.rs
tizoc: I would rather avoid any explicit sorting, will just add a comment to `Tree`. Note that the sorted requirement has always been there, it has nothing to do with the new code, it is just that one of the Inode variants has the same constraints as the case for which Inodes are not required.

bkontur: @tizoc 
ok, cool, so please jst "add a comment to Tree"

tizoc: We have this already, but it is too vague, will update it with a better description:

```
// Tree must be an ordered structure for consistent hash in hash_tree
// Currently immutable OrdMap is used to allow cloning trees without too much overhead
pub(crate) type Tree = im::OrdMap<String, Node>;
```

storage/src/hash/mod.rs
tizoc: Sounds good, something that I find very smelly right now is how `hash` depends on stuff from the upper level, which also depends on stuff from `hash`, makes me uncomfortable.

bkontur: exactly, looks like kind of circular dependency now, thats why I want to move common stuff to src/context/mod.rs later

tizoc:
@bkontur ok, just pushed, let me know if you see anything else that needs changing.

tizoc:
Once all is good I will squash into a single commit.

bkontur (approved):
@tizoc 
Bruno, very very good,
please merge it also to develop

PR #476, opened 2021-03-04
New commit log without compression

mambisi:


storage/src/persistent/commit_log.rs
```diff
@@ -105,65 +110,46 @@ impl<S: CommitLogSchema> CommitLogWithSchema<S> for CommitLogs {
...
-                location: *location,
-            })?;
-        let bytes = msg_buf.iter().next().ok_or(CommitLogError::ReadError {
-            error: ReadError::CorruptLog,
-            location: *location,
-        })?;
-        let value = S::Value::decode(bytes.payload())?;
+        let mut msg_buf =
+            cl.read(location.0 as usize, 1)
+                .map_err(|_| CommitLogError::ReadError {
```
bkontur, line 126: @mambisi 
why did you removed error?
if something is wrong, we need to know, at least from log
```
 ReadError {
        error: ReadError,
```

bkontur: @mambisi 
and please move this `commit_log.rs` to subdir `commit_log`

mambisi: Read error was the old commit log, custom error message

mambisi: read error wasnt removed i modified it

storage/src/commit_log/error.rs
```diff
@@ -0,0 +1,37 @@
+use failure::Fail;
+use std::io;
+
+#[derive(Debug, Fail)]
+pub enum TezedgeCommitLogError {
```
bkontur, line 5: @mambisi 
do we need both: TezedgeCommitLogError and CommitLogError?

storage/src/persistent/commit_log.rs
bkontur: @mambisi
but now error is supressed, and I would like to propagete it at least as a part of CommitLogError::ReadError
```
.map_err(|_|
```

storage/src/commit_log/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+
+impl Iterator for MessageSet {
+    type Item = Message;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let index = self.indexes.get(self.cursor)?;
+        let data = &self.buf[self.acc..(self.acc + index.data_length as usize)];
+        self.acc += index.data_length as usize;
+        self.cursor += 1;
+        Some(data.to_vec())
```
tizoc, line 48: Going to repeat my comment from the other PR here because it was not answered.

Is this MessageSet going to be big? is performance an issue here? because it is allocating a new (short-lived it seems) vector on each iteration, and I think thats something that could be fixed (with https://docs.rs/streaming-iterator/ for example).

storage/src/persistent/commit_log.rs
```diff
@@ -308,4 +290,23 @@ mod tests {
             ranges
         );
     }
+
+    #[test]
+    fn test_commit_log_storage() {
+        let path = env!("LOG_DIR");
```
bkontur, line 296: @mambisi 
better change it here something like:
```
    let out_dir = env::var("OUT_DIR").expect("OUT_DIR is not defined");
    Path::new(out_dir.as_str()).join("LOG_DIR")
```

or what is LOG_DIR anyway?

storage/src/commit_log/mod.rs
mambisi: the size will depend on the mount of data being fetched, i am reading the data in irregular sizes i dont know how the streaming iterator can help, but if you can provide me a example i will have a look at it

storage/src/persistent/commit_log.rs
mambisi: It was for debugging purposes, will be removed

storage/src/persistent/commit_log.rs
mambisi: Ok

storage/src/commit_log/error.rs
mambisi: We dont but we might want to change the implementation in the future

bkontur: @mambisi 
ok, cool, so I would simplify it now, and if there will come any needs, we will change it in future

storage/src/commit_log/mod.rs
tizoc: Let's say the `MessageSet` contains 100 entries, if we iterate it with the current implementation, we will perform 100 allocations to copy from the chunks of data we already have. Now, if we process 100 `MessageSet`s, that is 100*100 allocations.

With the streaming iterator we would instead just borrow those slices, avoiding the need to perform allocation here.

storage/src/commit_log/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+    }
+    #[inline]
+    pub fn append_msg<B: AsRef<[u8]>>(&mut self, payload: B) -> Result<u64, TezedgeCommitLogError> {
+        let offset = self.writer.write(payload.as_ref())?;
+        Ok(offset)
+    }
+
+    #[inline]
+    pub fn read(&self, from: usize, limit: usize) -> Result<MessageSet, TezedgeCommitLogError> {
+        let reader = Reader::new(&self.path)?;
```
tizoc, line 119: Do we need to instantiate a new reader for every read?

tizoc: We commented this same thing on the old version of the PR and it was fixed: https://github.com/simplestaking/tezedge/pull/471/files#diff-ba0b581e19257bfa979ccf3802d5e614776170841c9fd21f57b070ed9f33888bR113-R128

storage/src/commit_log/mod.rs
tizoc: cc @bkontur

tizoc: Another option (assuming the `MessageSet` is not used after it gets iterated over, because this would mutate it) could be to use Vec's `drain` method https://doc.rust-lang.org/std/vec/struct.Vec.html#method.drain

tizoc: ok, @bkontur said:

> we usually read by one item (read one BlockHeader’s data for RPC request), but I think there are some list rpcs for headers, I hope that means just “few”

If that is the case I guess we are good for now (I would still like to optimize this, but not important for now, so don't worry about it).

storage/src/commit_log/mod.rs
mambisi: Yes for some reason the `flush()` doesnt update the reader so, it doesn't read correctly

mambisi: The reader get out of sync with the writer, that is the biggest problem i have been facing

storage/src/commit_log/mod.rs
mambisi: we only read 50 headers at time

storage/src/commit_log/mod.rs
tizoc: @mambisi ok, makes sense, because the reader has an internal buffer.

PR #478, opened 2021-03-05
Async p2p read [TE 406]

akoptelov:
- In addition to `BinaryReader` there is also `BinaryAsyncReader` that uses `AsyncRead`as a source
- BinaryMessage has new method for reading from `AsyncRead`
- EncryptedMessageReader now implements `AsyncRead` by reading and decrypting chunks


akoptelov:
Closing this one as for now we have more concrete solution for the problem.

PR #484, opened 2021-03-09
New commit log without compression (rebased)

tizoc:
Was #476


tizoc:
@mambisi when you get a chance please take a look here to make sure everything is right (I will squash the commits too in a bit, just want to be sure I didn't mess up the diff)

tizoc:
@mambisi @bkontur I guess this one was replaced by (or combined with) some other PR, closing.

PR #485, opened 2021-03-09
Context hash dataset tests

tizoc:
Tests using the datasets from: https://github.com/tarides/tezos-context-hash


bkontur (approved):
@tizoc 
please merge to develop

PR #503, opened 2021-03-16
Cleaned simplified commit log with compression

mambisi:
Cleaned and rebased (#492)


.drone.yml
```diff
@@ -2018,8 +2018,8 @@ clone:
 environment:
   NETWORK: mainnet
   FROM_BLOCK_HEADER: 0
-  TO_BLOCK_HEADER: 130 # also check parameter --stop=
-  TO_BLOCK_HEADER_FOR_RPC: 3
+  TO_BLOCK_HEADER: 1000 # also check parameter --stop=
```
bkontur, line 2021: @mambisi 
here 1000 -> 10000

and here `--stop=131` -> `--stop=10001`
https://github.com/simplestaking/tezedge/pull/503/files#diff-b54b39f1afced2465e1f3641db9d5bbf4f3a7fcf890996dfedd3c197bcb7f8c7R2143

https://github.com/simplestaking/tezedge/pull/503/files#diff-b54b39f1afced2465e1f3641db9d5bbf4f3a7fcf890996dfedd3c197bcb7f8c7R2161

.drone.yml
```diff
@@ -2018,8 +2018,8 @@ clone:
 environment:
   NETWORK: mainnet
   FROM_BLOCK_HEADER: 0
-  TO_BLOCK_HEADER: 130 # also check parameter --stop=
-  TO_BLOCK_HEADER_FOR_RPC: 3
+  TO_BLOCK_HEADER: 10000 # also check parameter --stop=
```
bkontur, line 2021: @mambisi 
please, revert this values in drone.yml, it was just onetimer :)

Cargo.lock
```diff
@@ -2925,6 +2925,7 @@ dependencies = [
  "derive_builder",
  "failure",
  "flate2",
+ "fs_extra",
```
bkontur, line 2928: @mambisi 
please remove `"commitlog"` dependecy if not needed anymore

storage/Cargo.toml
```diff
@@ -24,6 +24,7 @@ sled = "0.34.6"
 snap = "1.0.4"
 strum = "0.20"
 strum_macros = "0.20"
+zstd = "0.6.1+zstd.1.4.9"
```
bkontur, line 27: @mambisi 
please remove "commitlog" dependecy if not needed anymore

storage/src/commit_log/mod.rs
```diff
@@ -0,0 +1,430 @@
+// Copyright (c) SimpleStaking and Tezedge Contributors
+// SPDX-License-Identifier: MIT
+
```
bkontur, line 3: @mambisi @jurajselep 
Mambisi, imagine, that you want to introduce commit log to someone, who have never heard about it,
so please, add here some simple documentation:
- what is it used for
- how it is working
- what and how is stored
- key points
- caveats, specialities like that compression


maybe the commit_log is also not a good name,
we will see, when you update documentation here maybe.

it does not have to be article-like style, just few simple bullets is ok,
or something like here:
https://github.com/simplestaking/tezedge/blob/master/storage/src/merkle_storage.rs#L4-L47

storage/src/commit_log/mod.rs
```diff
@@ -0,0 +1,430 @@
...
+    data_file_path: PathBuf,
+    use_compression: bool,
+}
+
+impl CommitLog {
+    pub fn new<P: AsRef<Path>>(dir: P, use_compression: bool) -> Result<Self, CommitLogError> {
+        if !dir.as_ref().exists() {
+            std::fs::create_dir_all(dir.as_ref())?;
+        }
+        let mut data_file_path = PathBuf::new();
```
bkontur, line 37: @mambisi 
maybe, this could be simplified like: `dir.as_ref().join(DATA_FILE_NAME)`

storage/src/commit_log/mod.rs
```diff
@@ -0,0 +1,430 @@
...
+    /// Retrieve a stored record.
+    fn get(&self, location: &Location) -> Result<S::Value, CommitLogError>;
+}
+
+impl<S: CommitLogSchema> CommitLogWithSchema<S> for CommitLogs {
+    fn append(&self, value: &S::Value) -> Result<Location, CommitLogError> {
+        let cl = self
+            .cl_handle(S::name())
+            .ok_or(CommitLogError::MissingCommitLog { name: S::name() })?;
+        let mut cl = cl.write().expect("Write lock failed");
```
bkontur, line 169: @mambisi 
no expects/unwraps please

storage/src/commit_log/mod.rs
```diff
@@ -0,0 +1,430 @@
...
+            writer.write_all(payload.as_ref())?;
+            payload.as_ref().len()
+        };
+        writer.flush()?;
+        Ok((offset, buf_size))
+    }
+
+    pub fn read(&self, offset: u64, buf_size: usize) -> Result<Vec<u8>, CommitLogError> {
+        let mut buf = vec![0_u8; buf_size];
+        let mut reader = BufReader::new(File::open(self.data_file_path.as_path())?);
```
bkontur, line 72: @mambisi 
do we need to File::open here? couldnt we just use `data_file: File,` ?

and another thing: `self.data_file_path.as_path()` could it be retrieved from `data_file: File,` ? if so, then we dont need `self.data_file_path`

storage/src/commit_log/mod.rs
```diff
@@ -0,0 +1,430 @@
...
+        let out = cl.append_msg(&bytes)?;
+
+        Ok(Location(out.0, out.1))
+    }
+
+    fn get(&self, location: &Location) -> Result<S::Value, CommitLogError> {
+        let cl = self
+            .cl_handle(S::name())
+            .ok_or(CommitLogError::MissingCommitLog { name: S::name() })?;
+        let cl = cl.read().expect("Read lock failed");
```
bkontur, line 180: @mambisi
no expects/unwraps please

storage/src/commit_log/mod.rs
```diff
@@ -0,0 +1,430 @@
...
+        let value = S::Value::decode(&bytes)?;
+        Ok(value)
+    }
+}
+
+pub fn fold_consecutive_locations(locations: &[Location]) -> Vec<Range> {
+    if locations.is_empty() {
+        Vec::with_capacity(0)
+    } else {
+        let mut ranges = vec![];
```
bkontur, line 191: @mambisi 
initialize `vec![]` please with `Vec::with_capacity(locations.len)` then the `ranges.push(range);` does not need to reallocate all the time on every push

storage/src/commit_log/mod.rs
```diff
@@ -0,0 +1,430 @@
...
+        P: AsRef<Path>,
+        I: IntoIterator<Item = CommitLogDescriptor>,
+    {
+        let myself = Self {
+            base_path: path.as_ref().into(),
+            commit_log_map: RwLock::new(HashMap::new()),
+        };
+
+        for descriptor in cfs.into_iter() {
+            Self::register(&myself, descriptor.name())?;
```
bkontur, line 229: @mambisi 
could this be written as:
```
myself.register(descriptor.name())?;
```
```

File: storage/src/commit_log/mod.rs [hunk 12]
```diff
@@ -0,0 +1,430 @@
...
+
+    /// Register a new commit log.
+    fn register(&self, name: &str) -> Result<(), CommitLogError> {
+        let path = self.base_path.join(name);
+        if !Path::new(&path).exists() {
+            std::fs::create_dir_all(&path)?;
+        }
+        let log = CommitLog::new(path, true)?;
+
+        let mut commit_log_map = self.commit_log_map.write().unwrap();
```
bkontur, line 243: @mambisi 
please, no unwrap, mam Guard/PoisonError to CommitLogError

storage/src/commit_log/mod.rs
```diff
@@ -0,0 +1,430 @@
...
+        commit_log_map.insert(name.into(), Arc::new(RwLock::new(log)));
+
+        Ok(())
+    }
+
+    /// Retrieve handle to a registered commit log.
+    #[inline]
+    fn cl_handle(&self, name: &str) -> Option<CommitLogRef> {
+        let commit_log_map = self.commit_log_map.read().unwrap();
```
bkontur, line 252: @mambisi 
please, no unwraps, maybe better wrap ``Result<Option<CommitLogRef>>`

storage/src/commit_log/mod.rs
```diff
@@ -0,0 +1,430 @@
...
+        commit_log_map.get(name).cloned()
+    }
+
+    /// Flush all registered commit logs.
+    pub fn flush(&self) -> Result<(), CommitLogError> {
+        let commit_log_map = self.commit_log_map.read().unwrap();
```
bkontur, line 258: @mambisi 
please no unwraps

storage/src/commit_log/mod.rs
```diff
@@ -0,0 +1,430 @@
...
+        for commit_log in commit_log_map.values() {
```
bkontur, line 259: @mambisi 
I would try to better flush all values, and not to stop on the first failure

storage/src/commit_log/mod.rs
```diff
@@ -0,0 +1,430 @@
...
+            commit_log.flush()?;
+        }
+
+        Ok(())
+    }
+}
+
+impl Drop for CommitLogs {
+    fn drop(&mut self) {
+        let _ = self.flush().expect("Failed to flush commit logs");
```
bkontur, line 270: @mambisi 
no expect, expect in drop is kind of undefined behavior
better:
```
let _ = self.flush();
```

or 

```
if let Err(e) = self.flush() {
    println!("Failed to flush commit logs, reason: {:?}", e);
}
```

storage/src/commit_log/mod.rs
```diff
@@ -0,0 +1,430 @@
...
+use std::collections::HashMap;
+use std::fs::{File, OpenOptions};
+use std::io::{BufReader, BufWriter, Read, Seek, SeekFrom, Write};
+
+pub type CommitLogRef = Arc<RwLock<CommitLog>>;
+
+const DATA_FILE_NAME: &str = "table.data";
+
+pub struct CommitLog {
+    data_file: File,
```
bkontur, line 27: @mambisi @jurajselep 
Mambisi, this one, I am not really sure, is it ok to have just one big file?
we dont remove from this file (we would like to also, but this commit_log stuff with fixed Location does not support it yet). I am just afraid, when this file will be too long, lets. say 1300k blocks of mainnet?

storage/src/commit_log/mod.rs
mambisi: Ok

storage/src/commit_log/mod.rs
mambisi: https://doc.rust-lang.org/std/fs/struct.File.html,
std::fs:File has not `as_path()` method

mambisi: File might be auto sync when you try to read it without opening it, for the testing this was best to get consistent results in reading a file

mambisi (commented):
we can chunk it in smaller files, all that need to be updated is the `struct Location` to include the file id that holds the data

storage/src/commit_log/mod.rs
mambisi: we can chunk it in smaller files, all that need to be updated is the` struct Location` to include the file id that holds the data

bkontur: @mambisi @tizoc 
what do you recommend? 

I am thinking something like this:
```
pub struct Location(pub DbOffset, pub Offset, pub ByteLimit);
```

DbOffset - could be some u64 and it will point to internal CommitLog's map: `Map<DbOffset, fileName>`,
not to save fileName all the time to rocksdb

bkontur (changes requested):
@mambisi 
please check file: `src/commit_log/mod.rs` there are still som unresolved comments:
https://github.com/simplestaking/tezedge/pull/503/files#diff-ba0b581e19257bfa979ccf3802d5e614776170841c9fd21f57b070ed9f33888b

and also clippy stuff - unused imports

storage/src/commit_log/mod.rs
```diff
@@ -0,0 +1,468 @@
...
+    #[fail(display = "Commit log I/O error {}", error)]
+    IOError { error: io::Error },
+    #[fail(display = "Commit log {} is missing", name)]
+    MissingCommitLog { name: &'static str },
+    #[fail(display = "Failed to read record at {}", location)]
+    ReadError { location: Location },
+    #[fail(display = "Failed to read record data corrupted")]
+    CorruptData,
+    #[fail(display = "RwLock Poison Error {}", error)]
+    RWLockPoisonError { error: String },
```
bkontur, line 115: @mambisi 
maybe better: `RWLockPoisonError` -> `RwLockPoisonError`

storage/src/commit_log/mod.rs
mambisi: Ok

storage/src/commit_log/mod.rs
mambisi: Map<FileId,PathBuf>, where the FileId is cycle offset (1024.log, 2048.log), (1024.log will contain blocks from 0..1024)

mambisi: I am not sure if commit log stores blocks in order

storage/src/commit_log/mod.rs
mambisi: Ok

PR #499, opened 2021-03-15
Upgrade code to ocaml-interop v0.7.0

tizoc:
This change is all about the upgrade to boxroots, not that interesting.

ocaml-interop [CHANGELOG](https://github.com/simplestaking/ocaml-interop/blob/master/CHANGELOG.md#060---2021-03-15) for the new release.

PR #505, opened 2021-03-17
Upgrade code to ocaml-interop v0.7.2

tizoc:
Alternative version of #499 with a `tezos-sys` crate for linking libtezos.so


storage/src/persistent/codec.rs
```diff
@@ -182,7 +182,7 @@ impl BincodeEncoded for () {}
 
 /// Create number from a bytes
 ///
-/// ```rust
+/// ```rust,no_run
```
tizoc, line 185: Had to disable this one because it tries to load a library that is not available and fails.

tizoc:
Well, finally. Will squash and reorder the commits and merge tomorrow. @bkontur marked you for review so that you can see the beautiful hacks.

tezos/sys/src/lib.rs
```diff
@@ -0,0 +1,22 @@
...
+}
+
+/// This function does nothing. It exists to force cargo to link libtezos to crates
+/// that depend on tezos-sys.
+/// To do so, define a `pub` function in the crate that requires libtezos linking that
+/// calls this function.
+/// It doesn't seem to be required under normal compilation conditions, but when
+/// running the tests with the address sanitizer enabled, linking fails without
+/// this extra step.
+pub fn force_libtezos_linking() {}
```
tizoc, line 22: I have to expand this comment a bit to explain why it works. The problem is that cargo seems to not link in stuff that doesn't get called (but I have only experienced these issues when running the tests with ASAN enabled, and not always). What I do with this, is to add a call from the crates that require libtezos to be linked (the call does nothing), that way cargo see that the crate is used which forces it to be linked.

.drone.yml
```diff
@@ -1320,7 +1319,8 @@ environment:
   WRK_TEST_BLOCK_HEADER: 100
   WRK_TEST_DURATION: 10  # in seconds
   SODIUM_USE_PKG_CONFIG: 1
-  LD_LIBRARY_PATH: ./tezos/interop/lib_tezos/artifacts:./target/release
+  # Both tezos/sys and tezos/interop are used for compatibility with old and new versions
+  LD_LIBRARY_PATH: ./tezos/sys/lib_tezos/artifacts:./tezos/interop/lib_tezos/artifacts:./target/release
```
tizoc, line 1323: This is temporary. We will be able to remove this duplication after some time when this PR has been merged into develop and every other PR has been rebased against develop.

bkontur (approved):

PR #507, opened 2021-03-18
Safer handling of String encoding

akoptelov:
- Vector is not preallocated when decoding strings
- Remaining size is checked when decoding strings
- `NetworkVersion.chain_name` is constrained


bkontur (approved):
@akoptelov 

PR #508, opened 2021-03-18
Storage refactor

bkontur:
Original PR: https://github.com/simplestaking/tezedge/pull/463

- fixes data race in mark&move gc (https://simplestakingcom.atlassian.net/browse/TE-426)
- common memory usage statistics for all the backends (https://simplestakingcom.atlassian.net/browse/TE-431)
- single abstraction over all merkle storage backends (https://simplestakingcom.atlassian.net/browse/TE-432)
- GC implementations `mark&sweep` and `mark&move` with interface shared with other backends
- tests for MerkleStorageBackend for all the backends implementations

- [x]  StorageType and createion ccheck
- [x] // TODO: move to src/context/mod.rs
- [x] shell/scripts/storage_stats.rs move to storage module + reactor context_listner


light_node/src/configuration.rs
```diff
@@ -1259,4 +1188,63 @@ impl Environment {
...
+                })
+                .collect::<Vec<_>>()
+                .join(", ");
+            return Err(InvalidRecorderConfigurationError(format!(
+                "errors: {:?}",
+                errors
+            )));
+        }
+
+        // return just oks
```
tizoc, line 1238: Are errors silently ignored? no logging?

tizoc: Oh duh, missed the `return` above.

tizoc:
The fix for the GC data race is this one? https://github.com/simplestaking/tezedge/pull/508/commits/dec360ac2444c57a5b77d3f026f93ae65f36945f

From the description of the commit it looks like it is not final, right? Trying to understand if this should be considered solved or not for updating the Jira issue accordingly.

bkontur:
> 
> 
> The fix for the GC data race is this one? [dec360a](https://github.com/simplestaking/tezedge/commit/dec360ac2444c57a5b77d3f026f93ae65f36945f)
> 
> From the description of the commit it looks like it is not final, right? Trying to understand if this should be considered solved or not for updating the Jira issue accordingly.

@tizoc 
gc is turned off, we will need to integrate it properly later, that block_applied/cycle counting is not good, it was just temporary for testing

tizoc:
@bkontur ok, I see. Btw I don't really have much else to comment on this, and it is probably better to merge this now than before any optimization is implemented (unless you have something quick and small in mind, what I am thinking about deserves it's own PR)

PR #513, opened 2021-03-25
opt - unnecessesery clone for self.current_stage_tree.0

bkontur:


tizoc:
@bkontur I don't think avoiding those clones will help, the `im` data structures are CoW, so cloning is cheap. But you have one thing there that is worth exploring as an extension of #515, avoiding the unnecessary recalculation of hashes. Once we get numbers from that branch, if there is an improvement, I will look into that too (extending entries with a slot to store a reference to their own hash may be enough).

PR #515, opened 2021-03-25
Use Arc<EntryHash> for entry hashes to avoid unnecessary duplication

tizoc:
Just an experiment for now.


storage/src/context/merkle/hash/mod.rs
```diff
@@ -150,7 +150,7 @@ fn hash_long_inode(inode: &Inode) -> Result<EntryHash, HashingError> {
                     NodeKind::Leaf => hasher.update(&[1u8]),
                     NodeKind::NonLeaf => hasher.update(&[0u8]),
                 };
-                hasher.update(node.entry_hash);
+                hasher.update(node.entry_hash.as_ref());
```
tizoc, line 153: @bkontur the part I like least about this PR is the changes I had to do to the hashing functions as a result of the Arc wrappings. Fortunately it is not much and can be reverted easily. I think the rest is quite safe to integrate, but let me know what you think tomorrow.

bkontur: looks ok, I am not sure about this cloned before hex::ecnode https://github.com/simplestaking/tezedge/pull/515/files#diff-4723e0d4bd9c6f2bdb715752a2ca56ebcaa8796ec6dc57a2170fe555741bcf84R418
could this cause new additional cloning?

tizoc: @bkontur good catch, that is a leftover from something I was doing before, it is not required. Removing it now.

tizoc:
replaced by #517

PR #517, opened 2021-03-26
Staging arc entry hash rebase

bkontur:


tizoc:
@bkontur one note about this one, I used Arc, because with Rc I got some complaints from the compilation of the chain_head_manager, but I didn't investigate much what was the reason.

PR #523, opened 2021-03-30
Add POW checking [TE-414]

akoptelov:
- `Peer` actor now checks proof of work right after receiving `ConnectionMessage`
- Two new tests are added to the `p2p_test`, one with non-zero POW target and another one with a peer with zero POW connecting to a node with non-zero one.

@bkontur I have to add a timeout at the end of the test, otherwise it looks like infrastructure initialization is not complete by the time it is dropped, and that cause consequent tests to fail.

### DIFF
--- a/networking/src/p2p/peer.rs
+++ b/networking/src/p2p/peer.rs
@@ -16,14 +16,18 @@ use tokio::net::TcpStream;
 use tokio::runtime::Handle;
 use tokio::time::timeout;
 
-use crypto::nonce::{self, Nonce, NoncePair};
 use crypto::{
     blake2b::Blake2bError,
     crypto_box::{CryptoKey, PrecomputedKey, PublicKey},
+    proof_of_work::PowError,
 };
 use crypto::{
     crypto_box::PublicKeyError,
     hash::{CryptoboxPublicKeyHash, Hash},
+};
+use crypto::{
+    nonce::{self, Nonce, NoncePair},
+    proof_of_work::check_proof_of_work,
 };
 use tezos_encoding::{
     binary_reader::{BinaryReaderError, BinaryReaderErrorKind},
@@ -67,6 +71,8 @@ pub enum PeerError {
     CryptoError { error: crypto::CryptoError },
     #[fail(display = "Public key error: {}", _0)]
     PublicKeyError(PublicKeyError),
+    #[fail(display = "Not enough proof of work: {}", _0)]
+    PowError(PowError),
 }
 
 impl From<BinaryWriterError> for PeerError {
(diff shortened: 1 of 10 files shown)


networking/src/p2p/peer.rs
```diff
@@ -438,6 +445,24 @@ pub async fn bootstrap(
...
+    if connecting_to_self {
+        debug!(log, "Detected self connection");
+        // treat as if nack was received
+        return Err(PeerError::NackWithMotiveReceived {
+            nack_info: NackInfo::new(NackMotive::AlreadyConnected, &[]),
+        });
+    }
+
+    // make sure the peer performed enough crypto calculations
+    match check_proof_of_work(&received_connection_message_bytes.raw()[4..60], pow_target) {
```
bkontur, line 461: @akoptelov 
I think, this will be enought:
```
if let Err(e) = check_proof_of_work(...) {
   Err(e) => return Err(PeerError::PowError(e)),
}
```

networking/src/p2p/peer.rs
```diff
@@ -438,6 +445,24 @@ pub async fn bootstrap(
     let connection_message =
         ConnectionMessage::from_bytes(received_connection_message_bytes.content())?;
 
+    // create PublicKey from received bytes from remote peer
+    let peer_public_key = PublicKey::from_bytes(connection_message.public_key())?;
+
+    let connecting_to_self = peer_public_key == info.identity.public_key;
+    if connecting_to_self {
+        debug!(log, "Detected self connection");
```
bkontur, line 453: @akoptelov 
change this debug to warn please,
if this happens a lot, then we need to fix it not to happen

shell/src/peer_manager.rs
```diff
@@ -405,6 +413,7 @@ impl
             discovery_last: None,
             check_peer_count_last: None,
             shutting_down: false,
+            pow_target,
```
bkontur, line 416: @akoptelov 
I would add pow_target to the local_node_info: 
```
local_node_info: Arc::new(LocalPeerInfo::new(
                p2p_config.listener_port,
                identity,
                shell_compatibility_version,
            )),
```

this struct is passed already to bootstrap fn, so we dont need to add extra parameter

shell/tests/common/test_node_peer.rs
```diff
@@ -1,10 +1,10 @@
-/// Test node peer, which simulates p2p remote peer, communicates through real p2p socket
-use std::collections::HashSet;
 use std::net::SocketAddr;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, RwLock};
 use std::thread;
 use std::time::{Duration, SystemTime};
+/// Test node peer, which simulates p2p remote peer, communicates through real p2p socket
```
bkontur, line 6: @akoptelov 
please check all files here in common, it should start with:
```
```

and in infra.rs there are unused import commented, please remove them

bkontur (changes requested):
@akoptelov 
please see, comments,
but very very good that you automatically added a tests to p2p_test :),

I would like to better have unit tests for peer.rs to run all the time, I know it is related to TpcStream, but ok, step-by-step

akoptelov:
> I would like to better have unit tests for peer.rs to run all the time, I know it is related to TpcStream, but ok, step-by-step

So would I :)

To make it possible we need the following:
- Refactor `peer::bootstrap()` into two separate functions, like `handshake()` that exchanges `ConnectionMessage`s (and probably `Metadata`), and, ok, `bootstrap()` itself, that does the rest.
- Create simple infrastructure for two nodes, like in `shell` tests, but without storage etc.

Nothing impossible, but surely will take time.

bkontur:
@akoptelov 
Alex, could you, please, rebase this on actual develop?

akoptelov:
@bkontur  Dpne, please review.

bkontur (approved):
@akoptelov 
I guess, looks ok, 
please rebase/squash and merge to develop

bkontur:
@akoptelov 
I need to merge it now, so I moved your branch to tezedge repo,
and this is PR to be merged: https://github.com/tezedge/tezedge/pull/559

PR #524, opened 2021-03-31
Use consistent max logging level

akoptelov:
Debug builds will have all logging levels enabled.
Release builds will have debug as a maximum possible level.


bkontur (approved):
@akoptelov 
please rebase/merge to develop

PR #527, opened 2021-04-05
Optimize the Staging Area Tree part of Context Storage

sebastiencs:
This is a draft of my work on [TE-484], it's not finish yet.

This PR removes the hashmap `staged` from `MerkleStorage`.
Each `Node` now keeps the entry itself and its hash, the hash is computed only when necessary (when serializing).

It uses interior mutability to write the hash on the `Node`, let me know if interior mutability is not desirable.

I will update the PR with benchmarks results and a better description when I'm done with the code.

[TE-484]: https://simplestakingcom.atlassian.net/browse/TE-484

### DIFF
--- a/storage/src/context/merkle/merkle_storage.rs
+++ b/storage/src/context/merkle/merkle_storage.rs
@@ -45,8 +45,8 @@
 //! ``
 //!
 //! Reference: https://git-scm.com/book/en/v2/Git-Internals-Git-Objects
-use std::collections::HashMap;
 use std::{array::TryFromSliceError, sync::Arc};
+use std::{cell::RefCell, collections::HashMap};
 
 use failure::{Error, Fail};
 use serde::Deserialize;
@@ -56,7 +56,7 @@ use crypto::hash::{FromBytesError, HashType};
 
 use crate::context::gc::GarbageCollectionError;
 use crate::context::merkle::hash::EntryHash;
-use crate::context::merkle::hash::{hash_blob, hash_commit, hash_entry, hash_tree, HashingError};
+use crate::context::merkle::hash::{hash_commit, hash_entry, hash_tree, HashingError};
 use crate::context::merkle::merkle_storage_stats::{
     MerkleStorageAction, MerkleStoragePerfReport, MerkleStorageStatistics, StatUpdater,
 };
@@ -92,19 +92,19 @@ enum Action {
 }
 
 pub struct MerkleStorage {
-    /// tree with current staging area (currently checked out context)
-    current_stage_tree: (Tree, TreeId),
+    /// Current working tree (currently checked out context)
+    working_tree: (Tree, TreeId),
     /// key value storage backend
     db: Box<ContextKeyValueStore>,
-    /// all entries in current staging area
-    staged: HashMap<Arc<EntryHash>, Entry>,
-    /// all different versions of the staging tree
+    /// all different versions of the working tree
     trees: HashMap<TreeId, Tree>,
-    /// HashMap for looking up entry index in self.staged by hash
+    /// Last commit hash
     last_commit_hash: Option<EntryHash>,
     /// storage latency statistics
     stats: MerkleStorageStatistics,
 }
+
+unsafe impl Send for MerkleStorage {}
 
 #[derive(Debug, Fail)]
 pub enum MerkleError {
@@ -142,8 +142,8 @@ pub enum MerkleError {
     FoundUnexpectedStructure { sought: String, found: String },
     #[fail(display = "Entry not found! Hash={}", hash)]
     EntryNotFound { hash: String },
-    #[fail(display = "Tree not in staging area! TreeId={}", tree_id)]
-    TreeNotFoundInStaging { tree_id: TreeId },
+    #[fail(display = "Tree not in working tree! TreeId={}", tree_id)]
+    TreeNotFoundInWorkingTree { tree_id: TreeId },
 
     /// Wrong user input errors
     #[fail(display = "No value under key {:?}.", key)]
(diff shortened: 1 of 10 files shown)


tizoc:
> It uses interior mutability to write the hash on the Node, let me know if interior mutability is not desirable.

It is totally fine, and I am not even sure what the alternative would be, so don't worry about that. I'm introducing interior mutability in many places too because of the transition to the "immutable" API too (required for caches).

storage/src/context/merkle/mod.rs
```diff
@@ -15,7 +15,7 @@ pub mod merkle_storage_stats;
 // Tree must be an ordered structure for consistent hash in hash_tree.
 // The entry names *must* be in lexicographical order, as required by the hashing algorithm.
 // Currently immutable OrdMap is used to allow cloning trees without too much overhead.
-pub type Tree = im::OrdMap<String, Arc<Node>>;
+pub type Tree = im::OrdMap<Rc<String>, Rc<Node>>;
```
tizoc, line 18: For now you will need to use `Arc` instead of `Rc` (see: http://ci.tezedge.com/simplestaking/tezedge/3294/1/3), but that will get fixed soon, and then we can go back to `Rc`.

bkontur:
> 
> 
> > It uses interior mutability to write the hash on the Node, let me know if interior mutability is not desirable.
> 
> It is totally fine, and I am not even sure what the alternative would be, so don't worry about that. I'm introducing interior mutability in many places too because of the transition to the "immutable" API too (required for caches).

@sebastiencs @tizoc 
I guess ok, we can use it, but we need to handle it very correctly, it means: no panics, no unwraps, no expects.
Just keep in mind borrow_mut can cause panic!, so maybe better use try_borrow_mut  and so on

tizoc:
@sebastiencs forgot to ask, but with the new representation you are working on, the idea is to introduce it along with the changes here? if not, what else is missing on this one? (it doesn't compile right now).

sebastiencs:
@tizoc The new representation I'm working on would actually replace this current PR.
I was thinking to work on the new representation instead of fixing this PR.
Should I fix it first ?

tizoc:
@sebastiencs if you are replacing this with the representation there is no need.

tizoc:
@sebastiencs small but important detail: just noticed that this is set against the `master` branch, should be `develop`.

tizoc:
@sebastiencs is this one ready for review?

sebastiencs:
@tizoc Not yet, I'm making a few small changes, I will mark it as ready

sebastiencs:
It is ready for review.
I marked `MerkleStorage` as `Send` and used a `Mutex` on `TezedgeContext` instead of a `RwLock` to make it `Send + Sync`

tizoc:
Ok, I have to finish some things first and then I will review this.

storage/src/context/merkle/merkle_storage.rs
```diff
@@ -1556,8 +1461,8 @@ mod tests {
             .commit(0, "author".to_string(), "message".to_string())
             .unwrap();
 
-        assert_eq!(storage.staged.is_empty(), false);
-        assert_eq!(storage.stage_checkout(1).is_err(), false);
+        // assert_eq!(storage.staged.is_empty(), false);
```
tizoc, line 1464: Since this one doesn't make sense anymore, just remove it.

storage/src/lib.rs
```diff
@@ -1,12 +1,11 @@
 // Copyright (c) SimpleStaking and Tezedge Contributors
 // SPDX-License-Identifier: MIT
-#![forbid(unsafe_code)]
```
tizoc, line 3: Was this because of the `Send` trait? Please leave it commented out (instead of deleted) with a `TODO - TE-261: uncomment once storage has moved to protocol runner`. We really want to forbid any `unsafe` code.

tizoc (approved):
Looks good. Left two comments, please address those, and if possible squash your commits (combined as they make sense or as a single commit when merging, whatever you feel is better).

tizoc:
@bkontur left my review, not sure if you have any feedback. Once TE-261 is solved this code will go away, so it just has to work and be "good enough" for now (it is already a big improvement over the version in `develop` in terms of speed and memory usage).

@sebastiencs once this is merged please port it over the new context in https://github.com/tezedge/tezedge/pull/529 (branch from that branch and set it as the target). The new code is in `tezos/new_context`, and `WorkingTree` is what her was `MerkleStorage`. It already follows the (kinda) immutable API, and so does the context API. It is not meant to be used by more than one thread so you should be able to get rid of all the synchronization stuff. You will probably find things to improve there too (that were not as easy here), feel free to tweak it as you see fit. And then you can start working on the subtree API.

storage/src/lib.rs
sebastiencs: Yes it's because of the Send trait, I added a TODO

storage/src/context/merkle/merkle_storage.rs
```diff
@@ -495,13 +472,11 @@ impl MerkleStorage {
...
 
         for (key, child_node) in prefixed_tree.iter() {
-            let entry = self.get_entry(&child_node.entry_hash)?;
-            let delimiter: &str;
-            if prefix.is_empty() {
-                delimiter = "";
-            } else {
-                delimiter = "/";
-            }
+            // let entry = self.get_entry(&child_node.entry_hash)?;
```
tizoc, line 475: @sebastiencs Remove this commented out code please.

tizoc (approved):

tizoc:
@sebastiencs one last thing I just noticed, we use signed commits on this repo: https://docs.github.com/en/github/authenticating-to-github/managing-commit-signature-verification

sebastiencs:
@tizoc The commit is now signed

PR #529, opened 2021-04-09
New context (Rust-side)

tizoc:
Still WIP (there is a bunch of things that I need to cleanup, and I have to finish the interface exposed through the FFI that OCaml will call). Once done, this will be renamed to `tezos/context` and the current `tezos/context` will be gone (along with all the context related stuff in the shell).

The code is base on `storage/context` but with many modifications. The most important changes:

- Introduced split between the "shell" context API (operations on the repository) and "protocol" context API (operations on the working tree)
- API is now immutable-ish, each operation that manipulates the context returns a new context
- Removed backtracking logic.
- Modified (on a copy) the context actions replay program so that it works with this new context, it handles backtracking by itself.
- `MerkleStorage` has been replaced by `WorkingTree`, which just contains the logic for manipulating the tree that the protocol is working with, the "repository" logic (commit, etc) has been moved to `TezedgeContext`.
- RocksDB support removed.

Pending

- [x] Cleanup
- [x] Cleanup more (better initialization of the context from the Rust side)
- [ ] Update what used to be the documentation for the "merkle storage" module (it is kind of outdated now)
- [x] Finish FFI interface
- [x] ~Re-introduce stats~ (handled as Rust hooks called from the OCaml side now)
- [ ] What were the `MerkleStorage` tests need to be updated for what now is the `WorkingTree`.
- [x] Make TezEdge buildable again without the need for a working OCaml environment (#551)
- [x] Fix RPC router (by adding support for OkChunk responses)
- [x] Make (non-routed) RPCs work with the new context (needs some remote access to the context storage from the writable protocol runner)
- [x] Add support for a remote context reader for the readonly protocol runners.
- [x] ~Use `BoxRoot` to keep OCaml values and skip the conversion at the FFI boundaries.~ (tested, didn't make much of a difference for now, and complicates things, will try again once things have stabilized)
- [x] See if it is worth it to use OCaml strings (without conversion to Rust) when performing lookups, to avoid the conversion. (TE-513 and #543 )
- [x] Enable garbage collection for the InMem storage
- [x] Implement `fold` (may need more testing, but seems to be working fine so far)
- [ ] Implement snapshot loading
- [x] Fix problem that causes node hashes to be lost because of clones related to partitioning when hashing big (>256 entries) directories.
- [ ] Restore/reimplement ContextActions storage (extend the timings recording into this?)
- [x] make path for the timings database configurable (#557)
- [x] use one transaction per block when storing timings
- [x] Verify that GC is behaving correctly (working well so far, we have to make sure there are no corner cases not being considered and stress it more)
- [ ] Make sure all the docker files, config files, drone, etc are updated
- [x] remove `--one-context` and all context that assumed an in-shell context

Aiming for just "good enough", since merging this will not break anything because it is not used yet (rest of the shell remains unchanged).


tizoc:
Btw, I copied the test data for the context hash tests, should probably change that so that it is just a link to the stuff in `storage`, and then once `storage` goes away we can move those with `git mv`. Will re-arrange the commits and force-push tomorrow.

tizoc:
@sebastiencs on this branch you will note that the staging stuff is still there, you have to remove it when porting your new version. The backtracking logic is gone from the `WorkingTree`, and has been moved to the context actions replayer which still needs it.

tizoc:
@bkontur I will probably need some help with these, because I am not sure about where to look.

## Current drone issues when using the new context

### Various RPC test fail because of `.../votes/listing`

Testing locally:

1) `/chains/main/blocks/head/votes/listings` works

2) `/chains/main/blocks/36180/votes/listings` works too: 
```
[{"pkh":"tz1fyYJwgV1ozj6RyjtU1hLTBeoqQvQmRjVv","rolls":68},{"pkh":"tz1fj3tzFejSmPyZZ2xsqehBxQE9GGr3rK8d","rolls":54},{"pkh":"tz1cXeGHP8Urj2pQRwpAkCdPGbCdqFUPsQwU","rolls":440},{"pkh":"tz1aWXP237BLwNHJcCD4b3DutCevhqq2T1Z9","rolls":891},{"pkh":"tz1aCnK45NP7CG31WMkUTP3d3V486mXwzZZt","rolls":1},{"pkh":"tz1YSzTPwEUpMrRxeTTHNo6RVxfo2TMN633b","rolls":412},{"pkh":"tz1VpvtSaSxKvykrqajFJTZqCXgoVJ5cKaM1","rolls":497},{"pkh":"tz1VbCypAKf8pnfSgjrtwdFEXN1g6RjQkvtn","rolls":15},{"pkh":"tz1VWasoyFGAWZt5K2qZRzP3cWzv3z7MMhP8","rolls":359},{"pkh":"tz1T8UYSbVuRm6CdhjvwCfXsKXb4yL9ai9Q3","rolls":141},{"pkh":"tz1RR6wETy9BeXG3Fjk25YmkSMGHxTtKkhpX","rolls":55},{"pkh":"tz1R55a2HQbXUAzWKJYE5bJp3UvvawwCm9Pr","rolls":404},{"pkh":"tz1PirboZKFVqkfE45hVLpkpXaZtLk3mqC17","rolls":444}]
```

3) `/chains/main/blocks/1/votes/listings` fails with:
```
ServiceError { reason: ProtocolServiceError { error: ProtocolError { reason: ContextGetKeyFromHistoryError { reason: "UnknownContextHashError { context_hash: \"CoWB1jTJ9rG9uhhtp21HWq916LfYrny591tPsL2DeKoeEtJ72jeH\" }" } } }
```

### `current-branch/head-shell-test` fails

I don't know if it is because it launches the node twice reopening the context (which doesn't work currently for the tezedge context), but the failure is caused by checkouts not finding the commit on TezEdge (only on Irmin).

tizoc:
Another issue is that `../context/raw_bytes` takes a lot of time to build, causing the IPC call to timeout, which then causes responses to be mis-ordered and everything else to fail. That endpoint will probably require some rethinking.

PR #535, opened 2021-04-20
Compute entry hash when necessary

sebastiencs:
This apply the changes from https://github.com/tezedge/tezedge/pull/527 into https://github.com/tezedge/tezedge/pull/529.

There is no difference except that it uses `Rc` instead of `Arc`.

Note that the signature of `ProtocolContextApi::get` and `ProtocolContextApi::dirmem` can be changed to:
```rust
fn get(&self, key: &ContextKey) -> ContextValue;
fn dirmem(&self, key: &ContextKey) -> bool;
```

But I am not sure how to return those types to ocaml here:
https://github.com/tezedge/tezedge/blob/protocol-runner-storage/tezos/new_context/src/ffi.rs#L178

### DIFF
--- a/tezos/new_context/src/working_tree/working_tree.rs
+++ b/tezos/new_context/src/working_tree/working_tree.rs
@@ -47,8 +47,7 @@
 //! ``
 //!
 //! Reference: https://git-scm.com/book/en/v2/Git-Internals-Git-Objects
-use std::{array::TryFromSliceError, cell::RefCell, sync::Arc};
-use std::{collections::HashMap, rc::Rc};
+use std::{array::TryFromSliceError, cell::RefCell, rc::Rc};
 
 use failure::Fail;
 use serde::Deserialize;
@@ -57,11 +56,9 @@ use serde::Serialize;
 use crypto::hash::{FromBytesError, HashType};
 
 use crate::hash::EntryHash;
-use crate::hash::{hash_blob, hash_commit, hash_entry, hash_tree, HashingError};
+use crate::hash::{hash_commit, hash_entry, hash_tree, HashingError};
 use crate::persistent;
-use crate::working_tree::working_tree_stats::{
-    MerkleStorageAction, MerkleStoragePerfReport, StatUpdater, TezedgeContextStatistics,
-};
+use crate::working_tree::working_tree_stats::{MerkleStoragePerfReport, TezedgeContextStatistics};
 use crate::working_tree::{Commit, Entry, Node, NodeKind, Tree};
 use crate::{gc::GarbageCollectionError, tezedge_context::TezedgeIndex};
 use crate::{ContextKey, ContextValue, StringTreeEntry, StringTreeMap};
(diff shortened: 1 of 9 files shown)


tizoc:
@sebastiencs for this API we want to get rid of `Result` whenever possible, because the OCaml-side API assumes no errors happen. Please add comments about `Result` not being required in any place in the API where you find it to be the case and we will take care of those later.

tezos/new_context/src/working_tree/working_tree.rs
```diff
@@ -57,7 +56,7 @@ use serde::Serialize;
 use crypto::hash::{FromBytesError, HashType};
 
 use crate::hash::EntryHash;
-use crate::hash::{hash_blob, hash_commit, hash_entry, hash_tree, HashingError};
+use crate::hash::{hash_commit, hash_entry, hash_tree, HashingError};
 use crate::persistent;
 use crate::working_tree::working_tree_stats::{
     MerkleStorageAction, MerkleStoragePerfReport, StatUpdater, TezedgeContextStatistics,
```
tizoc, line 62: Clippy is complaining about some unused imports here.

tezos/new_context/src/working_tree/working_tree.rs
```diff
@@ -90,43 +89,10 @@ enum Action {
     Remove(RemoveAction),
 }
 
-pub struct StagedCache {
```
tizoc, line 93: @sebastiencs not to be solved now (because 1) not a big deal for the in-mem store, 2) this StagedCache code was horrible), but what happens if after backtracking, I try to read a value that was already read before when working on the discarded tree? does it get read from the underlying storage again?

tezos/new_context/src/working_tree/working_tree.rs
```diff
@@ -558,56 +488,46 @@ impl WorkingTree {
...
+        self.compute_new_root_with_change(&key, Some(new_node))
     }
 
     /// Delete an item from the staging area.
     pub fn delete(&self, key: &ContextKey) -> Result<Self, MerkleError> {
         //let stat_updater = StatUpdater::new(MerkleStorageAction::Delete, Some(key));
-        let new_root_hash = &self._delete(&self.tree, key)?;
-        //self.set_stage_root(&self.get_tree(new_root_hash)?, new_tree_id);
-        let tree = self.get_tree(new_root_hash)?;
+        let new_root_hash = &self._delete(key)?;
```
tizoc, line 520: `new_root_hash` -> `new_root_entry` because of the change in `_delete`'s return type.

tezos/new_context/src/working_tree/working_tree.rs
```diff
@@ -619,8 +539,8 @@ impl WorkingTree {
         to_key: &ContextKey,
     ) -> Result<Option<Self>, MerkleError> {
         //let stat_updater = StatUpdater::new(MerkleStorageAction::Copy, Some(from_key));
-        if let Some(new_root_hash) = &self._copy(&self.tree, from_key, to_key)? {
-            let tree = self.get_tree(new_root_hash)?;
+        if let Some(new_root_hash) = &self._copy(from_key, to_key)? {
```
tizoc, line 542: Same here. `new_root_entry`.

tizoc (changes requested):
Looks good. Left a few comments (trivial stuff).

tizoc:
@sebastiencs just made a few changes to the base branch that caused conflicts here (because I changed `get` to return `None` when the entry is not found to match the OCaml-side API).

tizoc:
One more thing, I [updated](https://github.com/tezedge/tezedge/pull/529/commits/6c4cd6ad48bdc2aba5dd3dd686468bfec2f01637) some method names to match the new OCaml API:

- `set` -> `add`
- `get` -> `find`
- `dirmem` -> `mem_tree`

tizoc:
Not sure why, but when running with the address sanitizer enabled, this assert failed: https://github.com/tezedge/tezedge/blob/b6dcc1d8ab061ab2542eb756a7dce26f7476a9f4/tezos/new_context/src/gc/mark_move_gced.rs#L568-L574

tezos/new_context/src/working_tree/working_tree.rs
sebastiencs: Yes it will read from the underlying storage. Is this something we want to prevent ?

sebastiencs:
Comments addressed.
The failed assertion seems to be resolved without any changes related to it 🤔

tezos/new_context/src/working_tree/working_tree.rs
tizoc: Not important now, but something that could be improved in the future. It is a property that the old implementation had, but I don't really know how much it helped.

PR #536, opened 2021-04-21
Add subtree API to context

sebastiencs:
https://viablesystems.atlassian.net/browse/TE-500

This add alll the methods from the subtree API to `WorkingTree` but `fold` and `list`

### DIFF
--- a/tezos/new_context/src/lib.rs
+++ b/tezos/new_context/src/lib.rs
@@ -35,7 +35,7 @@ pub use actions::ActionRecorder;
 pub use hash::EntryHash;
 pub use tezedge_context::TezedgeContext;
 pub use tezedge_context::TezedgeIndex;
-use working_tree::NodeKind;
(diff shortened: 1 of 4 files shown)


tezos/new_context/src/lib.rs
```diff
@@ -88,6 +89,15 @@ where
     fn mem(&self, key: &ContextKey) -> Result<bool, ContextError>;
     // mem_tree - check if directory exists
     fn mem_tree(&self, key: &ContextKey) -> bool;
+    fn find_tree(&self, key: &ContextKey) -> Result<Option<Self>, MerkleError>;
```
tizoc, line 92: `find_tree` and `add_tree` need to use `WorkingTree` instead of `Self`, because when this trait is implemented for `TezedgeContext`, we don't want to add (or extract) a context, but a tree.

PR #537, opened 2021-04-21
Make `WorkingTree` a value or a tree

sebastiencs:


tezos/new_context/src/working_tree/working_tree.rs
```diff
@@ -136,6 +143,8 @@ pub enum MerkleError {
     ValueExpected(&'static str),
     #[fail(display = "Invalid state: {}", _0)]
     InvalidState(&'static str),
+    #[fail(display = "The working tree is not a tree")]
+    NotATree,
```
tizoc, line 147: I think this can be removed now.

tezos/new_context/src/working_tree/working_tree.rs
```diff
@@ -275,29 +300,30 @@ impl WorkingTree {
...
-        let rv = self.value_exists(&root, key);
+        let rv = self.value_exists(root.as_ref(), key);
 
         //stat_updater.update_execution_stats(&mut self.stats);
         rv
     }
 
     /// Check if directory exists in current staged root
-    pub fn mem_tree(&self, key: &ContextKey) -> bool {
+    pub fn mem_tree(&self, key: &ContextKey) -> Result<bool, MerkleError> {
```
tizoc, line 310: Same here, no more need for `Result`.

tezos/new_context/src/working_tree/working_tree.rs
```diff
@@ -243,8 +267,9 @@ impl WorkingTree {
         Self::new(self.index.clone())
     }
 
-    pub fn is_empty(&self) -> bool {
-        self.tree.is_empty()
+    pub fn is_empty(&self) -> Result<bool, MerkleError> {
```
tizoc, line 270: No more need for `Result` here.

tezos/new_context/src/lib.rs
```diff
@@ -88,14 +88,14 @@ where
     // mem - check if value exists
     fn mem(&self, key: &ContextKey) -> Result<bool, ContextError>;
     // mem_tree - check if directory exists
-    fn mem_tree(&self, key: &ContextKey) -> bool;
+    fn mem_tree(&self, key: &ContextKey) -> Result<bool, ContextError>;
```
tizoc, line 91: These `Result` aren't needed with this new version.

tizoc (changes requested):
@sebastiencs looks good, just some `Result` related changes that need to be reverted because with the new value-as-empty-tree semantics they are not needed anymore

tezos/new_context/src/working_tree/working_tree.rs
sebastiencs: Ah yes I will remove thoses

tizoc (approved):
@sebastiencs perfect👍  once the test pass we can merge, with this I can fill most of the pending holes at the FFI boundaries and OCaml-side implementation.

PR #539, opened 2021-04-23
Avoid serializing a node twice

sebastiencs:
This improves slightly the performance, while the memory usage doesn't change much.

Data when the replayer reaches the block 10000
|  | without this PR | with this PR
| --- | --- | --- |
| time | 53s | 41s
| memory | 2115 MB | 2110 MB

I checked the number of entries in [this hashtable](https://github.com/tezedge/tezedge/blob/protocol-runner-storage/tezos/new_context/src/kv_store/in_memory_backend.rs#L18) after having processed 10000 blocks, and it's the same with and without this PR


tizoc:
@sebastiencs cool, that is a good speedup. Just one thing I would change `serialized` to something else that is more descriptive about what we really care about here (`persisted`, `saved`, `committed` or something like that). Also CI is complaining because of rustfmt, otherwise, it looks good👍

PR #541, opened 2021-04-29
Add `--release --workspace` flags to cargo test run on GitHub Actions

tizoc:
These are used in drone, but not here (maybe one of the reasons why the test suite takes so much longer).


tizoc:
Looks like it just makes things worse, closing.

PR #542, opened 2021-05-04
Implementation of the context timing

sebastiencs:
This exposes the context stats on the RPC endpoint `GET /stats/context`.

Data is stored in a sqlite database from the protocol runner process, where statistics are computed.
When the RPC endpoint is called, it reads from that sqlite file and formats it according to the example JSON I got.
Note that the endpoint expose only tezedge stats, even though irmin ones are in the db.

- I'm not sure what should be the path of the database file ?
- There is no tests yet on the RPC endpoint, should I include an already filled sqlite database in the repository ? So that the tests would just read from it


tezos/new_context/src/timings.rs
```diff
@@ -49,10 +70,621 @@ pub fn context_action(
...
+        } else {
+            action.key.join("/")
+        };
+
+        let query = format!(
+            "
+        INSERT INTO actions
+          (name, key, irmin_time, tezedge_time, block_id, operation_id, context_id)
+        VALUES
+          ('{name}', '{key}', {irmin_time}, {tezedge_time}, {block_id}, {operation_id}, {context_id});
```
tizoc, line 393: Instead of buildings these with `format!`, we want to use Sqlite's placeholders, it is safer and faster. See the third snippet here for an example (the one with `?` placeholders): https://docs.rs/sqlite/0.26.0/sqlite/

tizoc: Btw the API for binding arguments in the `sqlite` crate looks like a PITA, `rusqlite` seems to provide a far more convenient way of doing it: https://docs.rs/rusqlite/0.25.1/rusqlite/

tizoc (changes requested):
@sebastiencs overall it looks good, I left some comments.

Also, I think it would be better to use the `rusqlite` crage, seems to have a more convenient API, more users, support for transactions, etc. Some of the things may not be needed now but if we extend this to be a replacement for the current context actions storage, those are going to be come useful (because those features will help with optimizing things more).

tizoc:
Oh and about your questions

> I'm not sure what should be the path of the database file ?
Good question. I am not sure yet, add some default and we can figure it out later.

> There is no tests yet on the RPC endpoint, should I include an already filled sqlite database in the repository ? So that the tests would just read from it
For the test you can initialize the Sqlite store in-memory, and insert whatever values you want to it, no need for an on-disk file.

tizoc:
In-memory SQlite database: https://sqlite.org/inmemorydb.html

tezos/new_context/src/timings.rs
```diff
@@ -49,10 +69,545 @@ pub fn context_action(
...
+        let operation_id = self.current_operation.as_ref().map(|(id, _)| id.as_str());
+        let context_id = self.current_context.as_ref().map(|(id, _)| id.as_str());
+
+        let key = if action.key.is_empty() {
+            None
+        } else {
+            Some(action.key.join("/"))
+        };
+
+        self.sql.execute(
```
tizoc, line 373: Possible improvement here now that we use param placeholders instead of building the string by hand (because it is called very often, and we want it to be as fast as possible) is to use [prepare_cached](https://docs.rs/rusqlite/0.25.1/rusqlite/struct.Connection.html#method.prepare_cached). 

This will prepare the query the first time and then reuse the cached version on every subsequent call, avoiding the overhead of parsing the SQL query.

tezos/new_context/src/timings.rs
```diff
@@ -49,10 +69,545 @@ pub fn context_action(
     // TODO - Bruno: it is possible to avoid these conversions by borrowing the internal
     // &str directly. Since we want this function to add as little overhead as possible
     // investigate doing that later once everything is working properly.
-    let action_name: String = action_name.to_rust(rt);
+    let name: String = action_name.to_rust(rt);
     let key: Vec<String> = key.to_rust(rt);
-    let _ = (action_name, key, irmin_time, tezedge_time);
+    // let _ = (action_name, key, irmin_time, tezedge_time);
```
tizoc, line 74: You can remove this leftover comment here.

About the comment above, we cannot use the OCaml `str` now because we are sending it to another thread, but what can be done is to just compare the bytes (`action_name.as_str()`) and have an enum for the action type to which it is mapped (`"set" => ContextAction::Set`), that way we avoid the String allocation here.

tezos/new_context/src/timings.rs
```diff
@@ -49,10 +69,545 @@ pub fn context_action(
...
+        self.times_in_current_block_by_name
+            .entry(action.name.clone())
+            .or_default()
+            .add(action.irmin_time, action.tezedge_time);
+
+        Ok(())
+    }
+
+    // Compute stats for the current block and global ones
+    fn update_global_stats(&mut self) -> Result<(), SQLError> {
```
tizoc, line 405: Instead of having to keep a vector of all timings in `*_times`, we can use an SQL query to calculate these values (will probably be slightly slower, but avoids having to keep those vectors) using the aggregate functions:

https://www.sqlite.com/lang_aggfunc.html#sumunc (`total`, `max` and `count`).

Btw, not a big deal, just something that can be improved a bit, but not a priority.

rpc/src/services/context.rs
```diff
@@ -39,245 +41,219 @@ pub(crate) struct BlockStats {
...
+                max_time: 0.0,
+                total_time: 0.0,
+                actions_count: 0,
+            },
+            mem: 0.0,
+            find: 0.0,
+            find_tree: 0.0,
+            add: 0.0,
+            add_tree: 0.0,
+            equal: 0.0,
```
tizoc, line 112: @sebastiencs  equal/hash/kind/empty/is_empty are not needed (I don't think I even send those)

rpc/src/services/context.rs
```diff
@@ -0,0 +1,244 @@
...
+pub(crate) struct BlockStats {
+    actions_count: usize,
+    checkout_context_time: f64,
+    commit_context_time: f64,
+    operations_context: Vec<Operation>,
+}
+
+#[derive(Debug, Serialize, Default)]
+#[serde(rename_all = "camelCase")]
+struct OperationData {
```
tizoc, line 19: UI says "operation" but actually these are "actions"

tizoc (approved):

PR #543, opened 2021-05-04
Avoid key-value allocations when performing lookups on the tree

sebastiencs:


tezos/new_context/bin/context_action_file_replayer.rs
```diff
@@ -340,6 +341,7 @@ fn resolve_context_kv_store(
     match context_kv_store_configuration {
         ContextKvStoreConfiguration::Sled { path } => ("sled".to_string(), Some(path.clone())),
         ContextKvStoreConfiguration::InMem => ("inmem".to_string(), None),
+        ContextKvStoreConfiguration::InMemGC => ("inmem-gc".to_string(), None),
```
bkontur, line 344: @sebastiencs @tizoc 
this is not a best way, I did not want to mix K-V with Garbage collection algorithm,
if it is just temporary fix, ok, but final solution should be:
- I choose k-v store on startup
- I choose GC algorithm on startup

tizoc: @bkontur I know, and it is very likely that we will end having a few more settings for the storage, not just the GC.

tezos/new_context/src/working_tree/working_tree.rs
```diff
@@ -233,6 +236,17 @@ impl WorkingTree {
         Ok(Some(Self::new_with_tree(self.index.clone(), tree)))
     }
 
+    pub fn find_tree_ocaml(&self, key: OCaml<OCamlList<String>>) -> Result<Option<Self>, MerkleError> {
```
tizoc, line 239: @sebastiencs why are all the `_ocaml` variants needed? isn't converting the `OCamlList<String>` into a `&[&str]` to use that as keys enough?

tizoc: Maybe you didn't change this yet and what I am seeing is just the push from the rebase, if that is the case just ignore me.

sebastiencs: Yes it's just a push after the rebase

tezos/new_context/bin/context_action_file_replayer.rs
```diff
@@ -525,22 +534,36 @@ fn perform_context_action(
             value,
             new_tree_id,
             ..
-        } => (current_context.add(&key, value)?, Some(new_tree_id)),
+        } => {
+            let key: Vec<_> = key.iter().map(String::as_str).collect();
```
tizoc, line 538: @sebastiencs this conversion between key representation repeats a lot, can it be abstracted?

tezos/new_context/src/hash/mod.rs
```diff
@@ -90,23 +90,23 @@ fn index(depth: u32, name: &str) -> u32 {
...
+) -> Result<Inode<'a>, HashingError> {
     if entries.is_empty() {
         Ok(Inode::Empty)
     } else if entries.len() <= 32 {
         Ok(Inode::Value(
-            entries
-                .iter()
-                .map(|(s, n)| (Rc::clone(*s), *n))
-                .collect(),
+            entries.iter().map(|(s, n)| ((*s).clone(), *n)).collect(),
```
tizoc, line 101: @sebastiencs isn't `Rc::clone` the idiomatic way to handle these cases where you clone an `Rc` ?

tizoc: Oh, duh, sorry, just noticed that it is now a `KeyFragment`, ignore me.

sebastiencs: It is, but here we're cloning a `KeyFragment`

tizoc (approved):
@sebastiencs ok, looks good to me. If you have time now, please abstract that key mapping logic in the context actions replayer, otherwise you can do it later, it is not that important.

sebastiencs:
Great, I'm making the changes in the context actions replayer

PR #544, opened 2021-05-04
Update tezos/sys build to download headers

tizoc:
Still a draft, right now requires the environment vaiable `OCAML_WHERE_PATH=$(pwd)/tezos/sys/lib_tezos/artifacts/include` to be set when running cargo commands so that the OCaml headers can be found, waiting for cargo 1.52 to be release which includes a way to set environment variables that get passed down to the build process:

https://github.com/rust-lang/cargo/pull/9175

@bkontur btw, I am thinking that it may be better to remove the libtezos download/copy logic from the build script and handle it on a separate standalone script, doing so will simplify things.


bkontur:
> 
> 
> Still a draft, right now requires the environment vaiable `OCAML_WHERE_PATH=$(pwd)/tezos/sys/lib_tezos/artifacts/include` to be set when running cargo commands so that the OCaml headers can be found, waiting for cargo 1.52 to be release which includes a way to set environment variables that get passed down to the build process:
> 
> [rust-lang/cargo#9175](https://github.com/rust-lang/cargo/pull/9175)
> 
> @bkontur btw, I am thinking that it may be better to remove the libtezos download/copy logic from the build script and handle it on a separate standalone script, doing so will simplify things.

@tizoc 
could you, please, write more about "separate standalone scipt" solution?
Do you mean it like manual step before running `cargo`?
We need to think about to have possibility to build/run it, just with one command `cargo run` / `cargo test` because of CI, if anybody wants to add it to theirs CI and so on

tizoc:
@bkontur I am still thinking about it, but the basic idea is to move the current build script to it's own `cargo` subcommand (`cargo prepare-tezos` or something like that, I have to investigate how that works -- I just want to avoid having to rewrite it in bash, I like how the Rust one works). 

So it would be the same code, but will have to be called manually once before the actual build/test/run commands. Then the build script will just verify that the hashes match and fail if they don't.

There are two reasons for this:
- It all happening as part of the build is kind of a PITA because sometimes it inadvertently overrides things (not an issue most of the time, but kind of annoying when you are working on changes to libtezos)
- More importantly, this script now needs to download some files (the headers) that are needed by it's dependencies, so by the time the build script runs, it is already too late (sometimes it doesn't even run because the dependency fails to build because the headers are still not there).

tizoc:
Replaced by #551 for now, closing.

PR #546, opened 2021-05-10
Compute context stats in the protocol runner process

sebastiencs:
Computing global stats when the RPC endpoint is called is inefficient.


rpc/Cargo.toml
```diff
@@ -26,6 +26,7 @@ shell = { path = "../shell" }
 storage = { path = "../storage" }
 tezos_api = { path = "../tezos/api" }
 tezos_context = { path = "../tezos/context" }
+tezos_new_context = { path = "../tezos/new_context" }
```
tizoc, line 29: @sebastiencs a quick comment on this, nothing other than the protocol runner should depend on the tezos context, because that will require us to link libtezos in. Instead move the shared stuff somewhere else (`tezos/timings` maybe, whatever you think is good) and depend on that instead.

sebastiencs:
This computes both global and block statistics in the protocol runner process

tizoc:
@sebastiencs I see `tezos/timing/context_stats.db` was included in the last commit, please remove it.

sebastiencs:
@tizoc Thanks, the file is removed

tizoc:
@sebastiencs great! will do the full review + merge after the meeting.

tezos/timing/src/lib.rs
```diff
@@ -0,0 +1,870 @@
...
+        )?;
+
+        query.execute(named_params! {
+            ":root": root,
+            ":action_name": action_name,
+        })?;
+
+        let mut query = self.sql.prepare_cached(
+            "
+        UPDATE
```
tizoc, line 632: My idea was to keep the calculation on SQlite and make them faster by adding a few indexes (what made them slow was that a full linear scan needed to be made, but indexes fix that). But now that you have written all the calculations in Rust I think it is worth it to keep it that way, because if at some point we switch from SQlite to something else, all that code will keep working.

tizoc (approved):
@sebastiencs looks good, just one thing, please add the copyright notice that we have on every source code file in this repo to the timings source files, because it is missing there.

PR #548, opened 2021-05-12
[WIP] Arbitrary trait for crypto & p2p messages

pventuzelo:
[structural fuzzing] implement arbitrary trait for fuzzing p2p message

### DIFF
--- a/crypto/Cargo.toml
+++ b/crypto/Cargo.toml
@@ -14,3 +14,9 @@ num-traits = "0.2.8"
 rand = "0.7.3"
 sodiumoxide = "0.2.5"
 serde = { version = "1.0", features = ["derive"] }
+arbitrary = { version = "1", features = ["derive"], optional = true }
+
+[features]
+arbitrary-fuzz = [
+  "arbitrary",
+]
--- a/crypto/src/hash.rs
+++ b/crypto/src/hash.rs
@@ -9,6 +9,9 @@ use crate::{
 };
 use serde::{Deserialize, Serialize};
 
+#[cfg(feature = "arbitrary-fuzz")]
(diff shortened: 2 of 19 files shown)


crypto/Cargo.toml
```diff
@@ -14,3 +14,9 @@ num-traits = "0.2.8"
 rand = "0.7.3"
 sodiumoxide = "0.2.5"
 serde = { version = "1.0", features = ["derive"] }
+arbitrary = { version = "1", features = ["derive"], optional = true }
```
bkontur, line 17: @pventuzelo 
Patrick, could you, please, tell me more context or what is this about?

@akoptelov 
will this work also with Nom decoding?

akoptelov: > will this work also with Nom decoding?
@bkontur Yes, it should.

Patrick should be able to describe you in more details, but to my understanding this is done to allow fuzzing "from the other end" -- instead of giving fuzzed bytes to a parser we will be able to get fuzzed Rust structures and then feed them to an encoder (or other functions). So this is mostly for encoding. For decoding with nom we can use existing fuzz targets.

pventuzelo: @bkontur Sure, this is a feature to activate the arbitrary trait for p2p message. It's still WIP but the idea is to use that for structural fuzzing. More details about arbitrary trait [here](https://github.com/rust-fuzz/arbitrary)

@akoptelov yes exactly.

bkontur: @akoptelov @pventuzelo 
once https://github.com/tezedge/tezedge/pull/552 is merged,
you can rebase and continue on this also

PR #549, opened 2021-05-13
Compute total times in global stats

sebastiencs:
This avoid to compute the values on the front-end


sebastiencs:
The PR is ready for review:
- Compute total times in global stats, to avoid computing the values on the front-end
- Add irmin data to both global & block statistics
- Add read & write times in global statistics

The endpoint `GET /stats/context` takes a new query parameter `context_name` to select between irmin and tezedge

tizoc (approved):

PR #550, opened 2021-05-13
Multi backend for maindb 

mambisi:


light_node/src/configuration.rs
```diff
@@ -177,8 +177,8 @@ impl Storage {
     const LRU_CACHE_SIZE_64MB: usize = 64 * 1024 * 1024;
     const LRU_CACHE_SIZE_16MB: usize = 16 * 1024 * 1024;
 
-    const DEFAULT_CONTEXT_KV_STORE_BACKEND: &'static str = storage::context::kv_store::ROCKSDB;
-    const DEFAULT_CONTEXT_ACTIONS_RECORDER: &'static str = storage::context::actions::ROCKSDB;
+    const DEFAULT_CONTEXT_KV_STORE_BACKEND: &'static str = storage::context::kv_store::SLEDDB;
+    const DEFAULT_CONTEXT_ACTIONS_RECORDER: &'static str = storage::context::actions::FILE;
```
bkontur, line 181: @mambisi 
please leave it as SLEDDB (if possible), because default cfg supports searching context actions now,
it is reimplemented in Bruno/Sebastian branches

light_node/src/main.rs
```diff
@@ -46,10 +47,10 @@ mod configuration;
 mod identity;
 mod system;
 
-extern crate jemallocator;
+//extern crate jemallocator;
```
bkontur, line 50: @mambisi 
what about this jemalloc?
If everything is better without it, just remove the lines and depencencies please,
If it was just for purpose of memory-profiling and we want jemalloc, plese uncomment these lines

tizoc: @bkontur it is probably a good idea to have this enabled/disabled through a feature flag

rpc/src/server/dev_handler.rs
```diff
@@ -175,6 +175,21 @@ pub async fn dev_stats_memory(
...
+pub async fn dev_stats_rw(
+    _: Request<Body>,
+    _: Params,
+    _: Query,
+    env: RpcServiceEnvironment,
+) -> ServiceResult {
+    match dev_services::get_storage_rw_stats(&env.persistent_storage) {
+        Ok(resp) => make_json_response(&resp),
+        Err(e) => {
+            warn!(env.log(), "GetStatsRW: {}", e);
```
bkontur, line 187: @mambisi 
please change better description " GetStatsRW:":
```
warn!(env.log(), "Failed to get storage stats"; "reason" => e);
or 
warn!(env.log(), "Failed to get storage stats"; "reason" => format!("{}, e));
```

light_node/src/main.rs
```diff
@@ -511,17 +512,22 @@ fn main() {
...
 
     // initialize dbs
+    let maindb = initialize_maindb(
+        &log,
+        &env.storage.db_path,
+        env.storage.db.expected_db_version,
+        &main_chain,
+    )
+    .expect("Failed to initialize main db (sled) storage");
     let kv_cache = RocksDbCache::new_lru_cache(env.storage.db.cache_size)
```
bkontur, line 522: @mambisi 
this kv_cache should be just configured if ROCKSDB is selected,
so I would just add `caches`  parameter to the `initialize_maindb`

rpc/src/server/router.rs
```diff
@@ -255,6 +255,11 @@ pub(crate) fn create_routes(is_sandbox: bool, one_context: bool) -> PathTree<Met
         "/stats/memory",
         dev_handler::dev_stats_memory,
     );
+    routes.handle(
+        hash_set![Method::GET],
+        "/dev/stats/rw",
```
bkontur, line 260: @mambisi 
other stats starts  with `/stats/`
so I would change `/dev/stats/rw` to `/stats/storage/rw` or `/stats/maindb/rw`

storage/src/database/sled_backend.rs
```diff
@@ -0,0 +1,199 @@
...
+
+pub struct SledDBBackend {
+    db: sled::Db,
+}
+
+impl SledDBBackend {
+    pub fn new<P: AsRef<Path>>(path: P, cache: u64) -> Result<Self, Error> {
+        let db = Config::default()
+            .path(path)
+            .temporary(true)
```
bkontur, line 21: @mambisi 
what is temporary?
is it persitent?

storage/src/database/tezedge_database.rs
```diff
@@ -0,0 +1,355 @@
...
+use crate::block_storage::{BlockPrimaryIndex, BlockByContextHashIndex, BlockByLevelIndex};
+use crate::persistent::sequence::Sequences;
+use crate::database::notus_backend::NotusDBBackend;
+
+pub trait KVStoreKeyValueSchema: KeyValueSchema {
+    fn column_name() -> &'static str;
+}
+
+/// Custom trait to unify any kv-store schema access
+pub trait KVStore<S: KeyValueSchema> {
```
bkontur, line 23: @mambisi 
I would name this `KvStore` insteaed of `KVStore`

storage/src/persistent/mod.rs
```diff
@@ -39,6 +42,17 @@ where
     CommitLogs::new(path, cfs)
 }
 
+/// Open commit log at a given path.
+pub fn open_main_db<P>(path: P) -> Result<TezedgeDatabase, DatabaseError>
+where
+    P: AsRef<Path>,
+{
+    let path = path.as_ref().to_path_buf();
+    let backend = TezedgeDatabaseBackendOptions::NotusDB(NotusDBBackend::new(path.join("database"))?);
```
bkontur, line 51: @mambisi 
why is NotusDb here hard-coded?

bkontur (changes requested):
@mambisi 
please see comments, small cosmetics,
but overall it looks ok to me,
waiting for cfg and we are good to go

light_node/src/main.rs
bkontur: > n

good idea :)

storage/src/database/tezedge_database.rs
mambisi: Ok

light_node/src/main.rs
mambisi: Yeah it was for memory profiling, i will enable it

light_node/src/main.rs
mambisi: RockDB was removed from possible backend because of its `Iterator` lifetime constraints

storage/src/database/sled_backend.rs
mambisi: Persistent, i will change it

light_node/src/main.rs
bkontur: @mambisi 
if we dont need rocksdb anymore, just remove it and also from dependencies, right?

light_node/src/configuration.rs
mambisi: Leave it as ROCKDB or sled?

bkontur: if maindb is sled by default, so context actions store should be the same

storage/Cargo.toml
```diff
@@ -20,10 +20,13 @@ num_cpus = "1.13"
 rocksdb = {version = "0.15", features = ["snappy", "lz4", "zstd", "zlib"], default-features = false }
 serde = { version = "1.0", features = ["derive", "rc"] }
 slog = { version = "2.7", features = ["max_level_trace", "release_max_level_debug"] }
-sled = "0.34.6"
+sled = {version = "0.34.6", features=["compression"]}
 snap = "1.0.4"
 strum = "0.20"
 strum_macros = "0.20"
+bloomfilter = "1.0.5"
+notus = { git = "https://github.com/mambisi/notus.git"}
```
bkontur, line 28: @mambisi @jurajselep 
what about this?
if we use git reference, it should better point to some fixed value, like tag or commit_hash,
anyway, `notus` should be migrated under tezege, right?
`https://github.com/tezedge/notus.git`

storage/src/context/actions/mod.rs
```diff
@@ -20,6 +20,7 @@ pub mod action_file_storage;
 pub mod context_action_storage;
 
 pub const ROCKSDB: &str = "rocksdb";
```
bkontur, line 22: @mambisi 
if we dont use this constant `ROCKSDB`, please remove it

storage/src/database/backend.rs
```diff
@@ -0,0 +1,38 @@
...
+use crate::Direction;
+#[derive(Clone)]
+pub enum BackendIteratorMode {
+    Start,
+    End,
+    From(Vec<u8>, Direction),
+}
+
+pub trait TezedgeDatabaseBackendStore {
+    fn put(&self, column: &'static str, key: Vec<u8>, value: Vec<u8>) -> Result<(), Error>;
```
bkontur, line 11: @mambisi @tizoc 
I would rather introduce here some alias types maybe, what do you thing? something like:
```
pub type Key = Vec<u8>;
pub type Value = Vec<u8>;

...
fn put(&self, column: &'static str, key: Key, value: Value) -> Result<(), Error>;

```

light_node/src/main.rs
mambisi: no we will need it for context_actions

mambisi: it still has configuration for rockdb as a backend

storage/src/lib.rs
```diff
@@ -496,33 +508,36 @@ pub mod initializer {
     }
 
     pub fn initialize_rocksdb<Factory: RocksDbColumnFactory>(
```
bkontur, line 510: @mambisi 
if we dont use/need this rocksdb stuff, please, remove all of it

light_node/src/configuration.rs
```diff
@@ -177,8 +178,8 @@ impl Storage {
...
```
bkontur, line 182: @mambisi 
please remove DEFAULT_CONTEXT_ACTIONS_RECORDER

light_node/src/configuration.rs
```diff
@@ -971,6 +980,12 @@ impl Environment {
                         .collect(),
                 };
 
+                let maindb_backend: TezedgeDatabaseBackendConfiguration = args
+                    .value_of("maindb-backend")
+                    .unwrap_or("rocksdb")
```
bkontur, line 985: @mambisi 
here is a default "rocksdb" but then it means that DEFAULT_CONTEXT_KV_STORE_BACKEND is not used

PR #551, opened 2021-05-14
Override ocaml-boxroot-sys dependency to use one that includes the OCaml 4.10 headers

tizoc:
Different (temporal) approach for fixing the build on the new context branch.


tizoc:
Will probably just create a fork to keep the vendored stuff, copied all into the repo just to test it until I am sure it compiles fine on CI.

tizoc:
Merged directly locally, closing the PR.

PR #553, opened 2021-05-17
Use in-memory database during timing tests

sebastiencs:
This should fix tests in CI builds

### DIFF
--- a/tezos/timing/src/schema_stats.sql
+++ b/tezos/timing/src/schema_stats.sql
@@ -22,16 +22,22 @@ CREATE TABLE IF NOT EXISTS contexts (
   hash TEXT UNIQUE
 );
 
+CREATE TABLE IF NOT EXISTS keys (
+  id INTEGER PRIMARY KEY AUTOINCREMENT,
+  key TEXT NOT NULL UNIQUE
+);
+
 CREATE TABLE IF NOT EXISTS actions (
   id INTEGER PRIMARY KEY AUTOINCREMENT,
   name TEXT,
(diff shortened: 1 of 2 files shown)


tizoc:
@sebastiencs ok, cool. I see this one includes the key deduplication, is that part ready too?

sebastiencs:
@tizoc Yes this is ready

tizoc:
Ok, cool, waiting for the tests to finish and then I will merge.

tezos/timing/src/schema_stats.sql
```diff
@@ -22,16 +22,22 @@ CREATE TABLE IF NOT EXISTS contexts (
   hash TEXT UNIQUE
 );
 
+CREATE TABLE IF NOT EXISTS keys (
+  id INTEGER PRIMARY KEY AUTOINCREMENT,
+  key TEXT NOT NULL UNIQUE
```
tizoc, line 27: @sebastiencs did you notice any performance difference with this implementation? We are missing an index on this one, without it I think things will start to slow down after a bunch of keys have been added

sebastiencs: My understanding is that `UNIQUE` automatically makes an index on that column.
I tried to add a `CREATE INDEX key_idx ON keys(key);` but it makes a error saying that the index already exist.

tizoc: @sebastiencs oh, in sqlite I think `UNIQUE` already creates an index, let me check.

PR #557, opened 2021-05-18
Make the timing database path configurable

sebastiencs:
- Add a new parameter to the node: `timing-db-path`, it is the directory of the timing database.
- A new function `tezedge_timing_init` should be called from ocaml, before any other hooks, to set the db path
- RPC: Return a 404 HTTP response when block is not found

### DIFF
--- a/light_node/etc/tezedge/tezedge.config
+++ b/light_node/etc/tezedge/tezedge.config
@@ -30,12 +30,12 @@
 # --bootstrap-db-path <PATH>
 --bootstrap-db-path=bootstrap_db
 
-# Path to timing database directory
+# Path to context-stats database directory
 # In case it starts with "./" or "../", it is relative path to the current dir, otherwise to the --tezos-data-dir
 # If directory does not exists, it will be created. If directory already exists, and contains valid database, node
 # will erase it and create a new database
-# --timing-db-path <PATH>
---timing-db-path=timing_db
+# --context-stats-db-path <PATH>
+--context-stats-db-path=context-stats_db
 
 #Max number of threads used by database configuration. If not specified, then number of threads equal to CPU cores.
 #--db-cfg-max-threads <NUM>
--- a/light_node/etc/tezedge/tezedge.config
+++ b/light_node/etc/tezedge/tezedge.config
@@ -30,6 +30,13 @@
 # --bootstrap-db-path <PATH>
 --bootstrap-db-path=bootstrap_db
 
+# Path to timing database directory
+# In case it starts with "./" or "../", it is relative path to the current dir, otherwise to the --tezos-data-dir
+# If directory does not exists, it will be created. If directory already exists, and contains valid database, node
+# will erase it and create a new database
+# --bootstrap-db-path <PATH>
+--timing-db-path=timing_db
+
 #Max number of threads used by database configuration. If not specified, then number of threads equal to CPU cores.
 #--db-cfg-max-threads <NUM>
 #--db-context-cfg-max-threads <NUM>
@@ -149,4 +156,3 @@
 
 # Enable or disable private node. Use --peers to set IP addresses of the peers you want to connect to.
 # --private-node=false
-
--- a/light_node/etc/tezedge/tezedge.config
+++ b/light_node/etc/tezedge/tezedge.config
@@ -34,7 +34,7 @@
 # In case it starts with "./" or "../", it is relative path to the current dir, otherwise to the --tezos-data-dir
 # If directory does not exists, it will be created. If directory already exists, and contains valid database, node
 # will erase it and create a new database
-# --bootstrap-db-path <PATH>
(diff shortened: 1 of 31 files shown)


light_node/etc/tezedge/tezedge.config
```diff
@@ -30,6 +30,13 @@
 # --bootstrap-db-path <PATH>
 --bootstrap-db-path=bootstrap_db
 
+# Path to timing database directory
+# In case it starts with "./" or "../", it is relative path to the current dir, otherwise to the --tezos-data-dir
+# If directory does not exists, it will be created. If directory already exists, and contains valid database, node
+# will erase it and create a new database
+# --bootstrap-db-path <PATH>
+--timing-db-path=timing_db
```
tizoc, line 38: This comment doesn't match the one from `light_node/etc/tezedge/tezedge_drone.config`, and it mentions a different flag.

tizoc:
@sebastiencs I have a meeting now but will continue reviewing later, one thing I see is that the CI is failing because of a missing argument (the compilation error can also be seen at the end of the "files changed" tab).

tizoc:
@sebastiencs there are still some failures when compiling the tests, search for `TezosContextConfiguration`. Btw, it would be good to support an in-memory option for tests.

tizoc:
Or well, also optional, because I guess most tests don't really require the timings db (maybe you already added a way to make it optional, in that case the tests should use that).

tezos/api/src/ffi.rs
```diff
@@ -133,6 +133,7 @@ pub struct TezosContextConfiguration {
     pub enable_testchain: bool,
     pub readonly: bool,
     pub sandbox_json_patch_context: Option<PatchContext>,
+    pub timing_db_path: Option<PathBuf>,
```
tizoc, line 136: @sebastiencs I see it is optional here, but not elsewhere, why the mismatch? (it being optional is what we want)

tizoc: @sebastiencs btw, I am thinking that something like "context_stats_db" is a better name than "timings_db", we will probably use it for more stuff than just timings.

tezos/new_context/src/ffi.rs
```diff
@@ -649,6 +649,14 @@ ocaml_export! {
         timings::context_action(rt, action_name, key, irmin_time, tezedge_time);
         OCaml::unit()
     }
+
+    fn tezedge_timing_init(
+        rt,
+        db_path: OCamlRef<String>,
+    ) {
+        timings::init_timing(rt, db_path);
```
tizoc, line 657: Convert the string here instead and pass it directly, that way you don't have to pass the runtime handle around.

tezos/timing/src/lib.rs
```diff
@@ -912,6 +921,9 @@ mod tests {
         let block_hash = BlockHash::try_from_bytes(&vec![1; 32]).unwrap();
         let context_hash = ContextHash::try_from_bytes(&vec![2; 32]).unwrap();
 
+        TIMING_CHANNEL
```
tizoc, line 924: What happens when the other messages are received but the `InitTiming` one wasn't? Can it be made so that everything is ignored until this message is received?

tezos/timing/src/lib.rs
```diff
@@ -259,15 +262,22 @@ pub static TIMING_CHANNEL: Lazy<Sender<TimingMessage>> = Lazy::new(|| {
 });
 
 fn start_timing(recv: Receiver<TimingMessage>) {
-    #[cfg(not(test))]
-    let db_path = Some(DB_PATH.into());
+    let mut msgs = Vec::new();
```
tizoc, line 265: @sebastiencs I am confused about the logic here. I we are not timing, we want to discard everything, not collecting it right? if timings are never initiated, we will have a memory leak here.

tizoc: I would consider such situation a bug, and we can probably error out in that case (or log a warning). But lets revise it later once this has been merged and I have updated the OCaml side. For now lets just discard the messages.

sebastiencs:
@tizoc Branch updated with your comments addressed

tizoc:
@sebastiencs looks good, will wait until I am done with the GC stuff to merge this so that I can also update the OCaml side to make it call the initialization function.

light_node/src/configuration.rs
```diff
@@ -683,6 +690,7 @@ fn validate_required_args(args: &clap::ArgMatches) {
         )),
     );
     validate_required_arg(args, "bootstrap-db-path", None);
+    validate_required_arg(args, "context-stats-db-path", None);
```
tizoc, line 693: @sebastiencs just rechecking, and this should be an optional argument. We don't always want to measure this things (*we* do, but regular users probably don't).

tizoc: Great, thank you, will merge in a bit.

PR #563, opened 2021-05-25
Test - apply protocol switch blocks

adonagy:
WIP: test protocol switch block applications


.drone.yml
```diff
@@ -187,43 +226,92 @@ trigger:
     - develop
 
 ---
-##############################################################################################################
 kind: pipeline
-name: unit-tests
+name: test-protocol-switch-1-to-2
```
bkontur, line 230: @adonagy 
just a cosmetics, I would name it like: `mainnet-protocol-switch-1-to-2-test` :)

PR #565, opened 2021-05-29
Compute block times and fix single context

sebastiencs:


tezos/timing/src/schema_stats.sql
```diff
@@ -9,7 +9,9 @@ CREATE TABLE IF NOT EXISTS blocks (
   checkout_time_irmin REAL,
   checkout_time_tezedge REAL,
   commit_time_irmin REAL,
-  commit_time_tezedge REAL
+  commit_time_tezedge REAL,
+  start_time INTEGER DEFAULT CURRENT_TIMESTAMP NOT NULL,
```
tizoc, line 13: Not sure about this default here, I think it will store the value as a string formatted date and time (sqlite doesn't care much about what the type of the column is). I would rather set it as `REAL` without default and pass the time directly from Rust as a floating point number of the current time (we want better resolution than just 1 second).

tizoc: Another issue that I am not thinking about is that if the recording thread lags behind this value is going to be incorrect. I think the best thing to do is to get the timestamp inside the hook function, and send it as part of the message sent to the recording thread.

rpc/src/services/context.rs
```diff
@@ -309,7 +323,8 @@ fn make_block_stats_impl(
         let action_stats = ActionStats {
             data: ActionData {
                 root: root.clone(),
-                actions_count: row.get(1)?,
+                tezedge_count: row.get(1)?,
+                irmin_count: row.get(20)?,
```
tizoc, line 327: @sebastiencs is the count ever different between the two?

tizoc: (other than when a single context is being used)

sebastiencs: @tizoc It's only different when a single context is used (one of the two will be zero)

tizoc:
@sebastiencs ok, looks good, do I merge or were you going to make other changes?

sebastiencs:
@tizoc It's good for merging

PR #568, opened 2021-06-01
Add memory profiler

vlad9486:


apps/deploy_monitoring/src/image.rs
```diff
@@ -76,3 +76,9 @@ pub struct Sandbox;
 impl DeployMonitoringContainer for Sandbox {
     const NAME: &'static str = "deploy-monitoring-tezedge-sandbox-launcher";
 }
+
+pub struct TezedgeMemprof;
```
bkontur, line 80: @vlad9486 @adonagy @jurajselep 
do we want this just for TezEdge node or also for Ocaml node?

apps/deploy_monitoring/docker-compose.deploy.latest.yml
```diff
@@ -1,6 +1,18 @@
 version: "3"
 
 services:
+  tezedge-memprof:
```
bkontur, line 4: @vlad9486 @adonagy @jurajselep 
do we want to run memprof also on master.dev.tezedge.com?
if so, we need to add this also to `docker-compose.deploy.latest-release.yml`  right?

apps/deploy_monitoring/src/image.rs
vlad9486: @bkontur Now it work only for tezedge node. The name 'light-node' hardcoded into bpf module.

apps/deploy_monitoring/docker-compose.deploy.latest.yml
vlad9486: @bkontur You are right we will run it on master. I'll add.

bkontur:
@vlad9486 
could you please remove `libflate` dependency from deploy_monitoring/Cargo.toml,
it looks like it is unused

bkontur (approved):
@vlad9486 
for me ok

@adonagy 
Ado, what about you?

@vlad9486 
if Ado confirms, then rebase/squash/merge to develop, please

adonagy (approved):
Looks good to me

PR #570, opened 2021-06-01
Perf test failure

akoptelov:
This is for `wrk-performance-test-tezedge-new-vs-ocaml-vs-tezedge-old` failure investigations.


bkontur:
@akoptelov 
you dont need to remove all pipelines,
if you want to skip pipelines, just edit this:
```
trigger:
  branch:
    - masterX
    - developX
```

bkontur:
@akoptelov @jurajselep @adonagy 
if that wrk split works, lets split it for combinations:
- new vs ocaml
- new vs old

I guess, we dont need to compare old vs ocaml, because it was compared with previously merge, right?

akoptelov:
@bkontur  I don't think splitting will work... Currently we compare throughput/latency for all combinations, but fail if difference is too big only between old/new tezedge, To me it looks like the environment we run nodes is not suitable for performance measurements. 

@adonagy do you know details on how DroneCI runs these steps? Are there any limits on the whole pipeline? Let's say only single CPU is allocated to the pipeline. In this case scheduling can do tricks with latency, am I right? And if another pipeline is running on the same time, can it interfere with this test? So we can try and isolate environment somehow, and probably make it less loaded (e.g. can we stop Tezos nodes when we doing new/old comparison?).

bkontur:
> 
> 
> @bkontur I don't think splitting will work... Currently we compare throughput/latency for all combinations, but fail if difference is too big only between old/new tezedge, To me it looks like the environment we run nodes is not suitable for performance measurements.
> 
> @adonagy do you know details on how DroneCI runs these steps? Are there any limits on the whole pipeline? Let's say only single CPU is allocated to the pipeline. In this case scheduling can do tricks with latency, am I right? And if another pipeline is running on the same time, can it interfere with this test? So we can try and isolate environment somehow, and probably make it less loaded (e.g. can we stop Tezos nodes when we doing new/old comparison?).



@akoptelov 
I agree with "To me it looks like the environment we run nodes is not suitable for performance measurements.",
I said it before, that these comparision test should work in kind of isolation and with the "same binary"
like now:
- old tezege runs as cargo run
- new tezedge runs as light-node binary
and so on

PR #571, opened 2021-06-01
Store apply block results as bytes (WIP)

tizoc:


tezos/interop/src/ffi.rs
```diff
@@ -403,3 +420,79 @@ pub fn assert_encoding_for_protocol_data(
         })
     })
 }
+
+pub fn apply_block_result_metadata(
```
tizoc, line 424: @bkontur this function and the one immediately after are the ones that should be called to produce the JSON

tizoc:
@bkontur ok, I added more code and fixed more stuff. The part that is still obviously broken is the mismatch of the data format for the genesis result, which I guess should use bytes too.

When you get a chance take a look at the diff to see what is missing/wrong. If you point me into the right direction I can complete what is missing/fix what is wrong tomorrow.

tizoc:
Hmmm actually, there are more places that expect json values from the stored block data, those need to be updated too (or maybe methods like `block_header_proto_metadata_json()` updated to do the encoding on-the-fly).

PR #572, opened 2021-06-02
Deduplicate hashes and add new Garbage Collector

sebastiencs:


tezos/new_context/src/working_tree/working_tree.rs
```diff
@@ -1055,20 +1065,20 @@ impl WorkingTree {
...
-                found: "tree".to_string(),
-            }),
-            Entry::Blob(_) => Err(MerkleError::FoundUnexpectedStructure {
-                sought: "commit".to_string(),
-                found: "blob".to_string(),
-            }),
-        }
-    }
+    // // TODO: part of the outer layer, this should not be on the working tree
+    // fn get_commit(&self, hash: &EntryHash) -> Result<Commit, MerkleError> {
```
tizoc, line 1069: These methods that just read from the index are still on `WorkingTree` but in reality are methods on the index. On the branch where I am fixing the RPCs I am planning to remove these from here.

tezos/new_context/src/working_tree/working_tree.rs
```diff
@@ -636,85 +639,85 @@ impl WorkingTree {
...
-            self.get_key_values_from_tree_recursively(&fullpath, &entry, &mut keyvalues)?;
-        }
-
-        if keyvalues.is_empty() {
-            Ok(None)
-        } else {
-            Ok(Some(keyvalues))
-        }
-    }
+    // /// Get context tree under given prefix in string form (for JSON)
```
tizoc, line 642: Same for all these, not really part of `WorkingTree`

tezos/new_context/src/working_tree/working_tree.rs
```diff
@@ -501,30 +504,30 @@ impl WorkingTree {
...
-        // let stat_updater = StatUpdater::new(MerkleStorageAction::GetHistory, Some(key));
-
-        let commit = self.get_commit(commit_hash)?;
-        let entry = self.get_entry_from_hash(&commit.root_hash)?;
-        let rv = self.get_from_tree(self.get_tree(&entry)?, key);
-
-        // stat_updater.update_execution_stats(&mut self.stats);
-        rv
-    }
+    // /// Get value. Staging area is checked first, then last (checked out) commit.
```
tizoc, line 507: Not part of `WorkingTree` either

tezos/new_context/src/gc/new_gc.rs
```diff
@@ -0,0 +1,289 @@
...
+    }
+
+    fn remove(&mut self, free_id: HashId) {
+        let index = free_id.0;
+        self.values[index] = None;
+    }
+}
+
+pub struct NewGC {
+    current: BTreeMap<HashId, Arc<[u8]>>,
```
tizoc, line 91: What is the reason for the values being `Arc`? No two hashes will map to the same value.

tizoc: Another question, can we switch later to using the two in-memory backends we already here instead of `BTreeMap` directly?

tizoc: Oh wait, the `Arc` here is the hash, and not the value right?

sebastiencs: @tizoc The `Arc<[u8]>` here is the serialized value, not the hash.
It uses `Arc` instead of `Box` so that the value is readable from both the GC thread and the main thread.
So both threads can deserialize the value.

When the main thread wants to read a serialized value, it reads it from `MapValues` https://github.com/sebastiencs/tezedge/blob/deduplicate-hashes/tezos/new_context/src/gc/new_gc.rs#L63.

`current: BTreeMap<HashId, Arc<[u8]>>` contains the `HashId -> value` used in the current cycle.
`MapValues` contains `HashId -> value` of all cycles.

> Another question, can we switch later to using the two in-memory backends we already here instead of BTreeMap directly?

Yes, though the 2 memory backends (BTreeMap & HashMap) will replace the `MapValues`, though it will requires some changes

tezos/new_context/src/hash/mod.rs
```diff
@@ -281,286 +303,292 @@ pub(crate) fn hash_commit(commit: &Commit) -> Result<EntryHash, HashingError> {
...
-        let path = Path::new(&env::var("CARGO_MANIFEST_DIR").unwrap())
-            .join("tests")
-            .join("resources")
-            .join(file_name);
-        let file = File::open(path)
-            .unwrap_or_else(|_| panic!("Couldn't open file: tests/resources/{}", file_name));
-        GzDecoder::new(file)
-    }
-}
+// #[cfg(test)]
```
tizoc, line 323: @sebastiencs what prevents these from working? it is important to have this tested

tizoc:
@sebastiencs I merged some changes that will cause conflicts here, mostly the methods that read from the repository, which got moved from `WorkingTree` to `TezedgeIndex`.

tezos/new_context/src/hash/mod.rs
sebastiencs: Ah, that's an unintended change. I will re-enable those tests

rpc/src/server/mod.rs
```diff
@@ -128,11 +128,11 @@ pub type HResult = Result<Response<Body>, Box<dyn std::error::Error + Sync + Sen
 
 pub type Handler = Arc<
     dyn Fn(
-        Request<Body>,
-        Params,
-        Query,
-        RpcServiceEnvironment,
-    ) -> Box<dyn Future<Output = HResult> + Send>
+            Request<Body>,
```
tizoc, line 131: @sebastiencs about this fmt, I don't know why, but locally my `cargo fmt` does this here, but then the checker in GitHub actions complains that it is wrong (and the old version here is the correct one). Please revert it.

I don't know why it behaves differently, but it is super annoying.

tezos/new_context/src/gc/worker.rs
```diff
@@ -0,0 +1,177 @@
...
+
+pub(crate) enum Command {
+    StartNewCycle {
+        values_in_cycle: BTreeMap<HashId, Option<Arc<[u8]>>>,
+        new_ids: Vec<HashId>,
+    },
+    MarkReused {
+        reused: Vec<HashId>,
+    },
+    Exit,
```
tizoc, line 29: @sebastiencs if this is not going to be used please remove it.

tezos/new_context/src/tezedge_context.rs
```diff
@@ -10,32 +10,41 @@ use std::{
...
     working_tree::{
-        working_tree_stats::MerkleStoragePerfReport, Commit, Entry, KeyFragment, Node, Tree,
+        working_tree::MerkleError, working_tree_stats::MerkleStoragePerfReport, Commit, Entry,
+        KeyFragment, Node, Tree,
     },
-    StringTreeMap,
+    ContextKeyValueStore, StringTreeMap,
 };
+use crate::{working_tree::working_tree::WorkingTree, IndexApi};
+// use crate::{
```
tizoc, line 27: Some commented out code here, if not needed just remove it.

tezos/new_context/src/tezedge_context.rs
```diff
@@ -625,26 +697,42 @@ impl ShellContextApi for TezedgeContext {
...
+        Ok(value.map(|v| v.to_vec()))
     }
 
     fn get_merkle_stats(&self) -> Result<MerkleStoragePerfReport, ContextError> {
         Ok(MerkleStoragePerfReport::default())
     }
 
     fn get_memory_usage(&self) -> Result<usize, ContextError> {
-        Ok(self.index.repository.read()?.total_get_mem_usage()?)
+        todo!()
```
tizoc, line 734: Make this return `Ok(0)` instead and leave a `TODO`, we should avoid anything that could kill the process.

tezos/new_context/src/tezedge_context.rs
```diff
@@ -625,26 +697,42 @@ impl ShellContextApi for TezedgeContext {
...
+
+        let (commit_hash_id, batch, _referenced_older_entries) = self.tree.prepare_commit(
+            date,
+            author,
+            message,
+            self.parent_commit_hash,
+            &mut *repository,
+        )?;
+
+        repository.write_batch(batch)?;
```
tizoc, line 711: This `hash` method shouldn't be writing anything to the repository, all it needs to do is to produce a `Commit` to calculate its hash

tizoc: If this is done so that you can get the hash, what can be done instead is to create a `Commit` manually (in the same way that `prepare_commit` does and pass that to `hash_commit`, or maybe even better yet, add a `hash_commit` to `WorkingTree` that just produces the hash.

tizoc: Ok, I see that `hash_commit` returns `HashId`, so it will have to be split into two functions, one that does the actual work (let's say `hash_commit_into`) and writes the resulting hash into a slice passed as an argument, and then `hash_commit` can use that. And `WorkingTree.hash_commit(author, message, date)` will just produce the hash directly.

sebastiencs:
@tizoc The PR is ready for review

> Ok, I see that hash_commit returns HashId, so it will have to be split into two functions, one that does the actual work (let's say hash_commit_into) and writes the resulting hash into a slice passed as an argument, and then hash_commit can use that. And WorkingTree.hash_commit(author, message, date) will just produce the hash directly.

About that, it would require to change also `hash_tree` and `hash_block`.
Instead it now just resets the `HashId` store at the end of `ShellContextApi.hash`

tezos/new_context/src/gc/worker.rs
```diff
@@ -57,15 +56,17 @@ impl Cycles {
...
 
         let value = value?;
-        self.list
-            .back_mut()
-            .unwrap()
-            .insert(hash_id, Some(Arc::clone(&value)));
+        if let Some(v) = self.list.back_mut() {
+            v.insert(hash_id, Some(Arc::clone(&value)));
+        } else {
+            eprintln!("GC: Failed to insert value in Cycles")
```
tizoc, line 62: @bkontur @adonagy do we log output from the protocol runners anywhere? trying to figure out what would be the best way to handle situations like this one (maybe more `Result` is needed here to get these bubbled up, but sometimes a warning is all we need)

tezos/new_context/src/gc/worker.rs
```diff
@@ -0,0 +1,183 @@
+use std::{
```
tizoc, line 1: @sebastiencs one small thing, here and in all source code files we need the copyright notice comment that we have everywhere else.

tezos/new_context/src/gc/worker.rs
```diff
@@ -0,0 +1,183 @@
...
+        for id in unused.keys() {
+            vec.push(*id);
+        }
+        vec
+    }
+}
+
+impl GCThread {
+    pub(crate) fn run(mut self) {
+        while let Ok(msg) = self.recv.recv() {
```
tizoc, line 82: @sebastiencs what would be an error result here? thinking that `loop` + `match` with logging + break on `Err`, or maybe even have `run` return `Result<(), AnError>` would be better. This is one of those things that if it fails silently, would be super hard to figure out.

tezos/new_context/src/gc/worker.rs
```diff
@@ -0,0 +1,183 @@
...
+        let navailable = self.free_ids.available();
+
+        let (to_send, pending) = if navailable < unused_length {
+            unused.split_at(navailable)
+        } else {
+            (&unused[..], &[][..])
+        };
+
+        if self.free_ids.push_slice(&to_send).is_err() {
+            eprintln!("GC: Fail to send free ids");
```
tizoc, line 117: Not sure what the possible errors are here, but `if let Err(err) = ...` to include the error in the output would be helpful.

tezos/new_context/src/gc/worker.rs
```diff
@@ -0,0 +1,183 @@
...
+        let navailable = self.free_ids.available();
+        if navailable == 0 {
+            return;
+        }
+
+        let n_to_send = navailable.min(self.pending.len());
+        let start = self.pending.len() - n_to_send;
+        let to_send = &self.pending[start..];
+
+        if self.free_ids.push_slice(&to_send).is_err() {
```
tizoc, line 141: Same as above

tezos/new_context/src/hash/mod.rs
```diff
@@ -260,35 +297,49 @@ pub(crate) fn hash_blob(blob: &ContextValue) -> Result<EntryHash, HashingError>
...
+        .ok_or(HashingError::ValueExpected("root_hash"))?;
+    hasher.update(root_hash.as_ref());
 
     if commit.parent_commit_hash.is_none() {
         hasher.update(&(0_u64).to_be_bytes());
     } else {
-        let parent_commit_hash = commit
-            .parent_commit_hash
+        let parent_commit_hash = store
+            .get_hash(commit.parent_commit_hash.unwrap())?
```
tizoc, line 316: @sebastiencs a new `unwrap()` was introduced here, but it should `.ok_or(HashingError::ValueExpected("parent_commit_hash"))?` as before.

tezos/new_context/src/kv_store/entries.rs
```diff
@@ -0,0 +1,61 @@
+use std::{
```
tizoc, line 1: Missing copyright comment here too.

tezos/new_context/src/persistent/database.rs
```diff
@@ -44,6 +44,18 @@ pub enum DBError {
     MemoryStatisticsOverflow,
     #[fail(display = "IPC Context access error: {:?}", reason)]
     IpcAccessError { reason: ContextServiceError },
+    #[fail(display = "Missing entry: {:?}", hash_id)]
+    MissingEntry { hash_id: HashId },
+    #[fail(display = "Attempt to hash tree on a readonly protocol runner")]
+    AttemptToHashOnReadOnly,
```
tizoc, line 50: @sebastiencs looks like this one is not needed enymore

tizoc:
@sebastiencs overall it looks good. I left some comments about small stuff that needs to be addressed first.

tezos/new_context/src/hash/mod.rs
sebastiencs: @tizoc This statement is actually in an `else` block of the condition `commit.parent_commit_hash.is_none()`.
So `commit.parent_commit_hash` is never `None`.
I will change the code to make that more clear

tizoc (approved):
@sebastiencs ok, looks good. Please merge with the "squash" option.

PR #577, opened 2021-06-05
Fix duration of a block

sebastiencs:
- Call `Instant::now` at the hook call, not in the timing thread which
might be lagging behind
- Do not recompute hashes for the commit and root of the tree

### DIFF
--- a/tezos/new_context/src/working_tree/working_tree.rs
+++ b/tezos/new_context/src/working_tree/working_tree.rs
@@ -753,7 +753,7 @@ impl WorkingTree {
         let mut referenced_older_entries: HashSet<EntryHash> = HashSet::new();
         self.get_entries_recursively(
             &entry,
-            &commit_hash,
+            commit_hash,
             Some(root.as_ref()),
             &mut batch,
             &mut referenced_older_entries,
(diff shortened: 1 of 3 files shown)


tezos/new_context/src/working_tree/working_tree.rs
```diff
@@ -1004,7 +999,7 @@ impl WorkingTree {
                             None => Ok(()),
                             Some(entry) => self.get_entries_recursively(
                                 entry,
-                                child_node.entry_hash.borrow().clone(),
+                                &child_node.entry_hash.borrow().clone().unwrap(),
```
tizoc, line 1002: What guarantees that this entry hash will be present? why not `child_node.entry_hash()?` like above?

tizoc: The other thing is that it is being cloned, to be passed as a reference, that will then be cloned again by the inner call.

The `commit_hash` recalculation is fine, I am just not entirely sure about this part here.

tizoc (commented):
Left a few comments regarding the children node hashes.

tizoc:
Other than that it looks good, it is just that part that I am not sure about.

tezos/new_context/src/working_tree/working_tree.rs
sebastiencs: @tizoc The hashes for the whole tree are actually already computed, before the call to `get_entries_recursively` here:
https://github.com/tezedge/tezedge/blob/protocol-runner-storage/tezos/new_context/src/working_tree/working_tree.rs#L738

I will fix the double clone

tizoc: Ahhh, good point.

PR #580, opened 2021-06-07
Readonly access to context through IPC

tizoc:


rpc/src/server/tezedge_context_remote.rs
```diff
@@ -0,0 +1,92 @@
...
+
+use std::sync::Arc;
+
+use crypto::hash::ContextHash;
+use failure::Fail;
+use tezos_new_context::{ContextError, ContextKeyOwned, ContextValue, StringTreeEntry};
+use tezos_wrapper::TezosApiConnectionPool;
+
+#[derive(Clone)]
+pub struct TezedgeContextRemote {
```
bkontur, line 12: @tizoc 
I would put this maybe to `tezos/wrapper` module, that we could use it maybe in shell, if needed

tizoc: Done and renamed to `*Client`, much better name.

PR #581, opened 2021-06-07
Preserve frame pointer

vlad9486:
- Patch rocksdb to preserve frame pointers.
- Add flag cargo build in distroless docker image to preserve frame pointers in our code.


bkontur (approved):
@vlad9486 
rebase/squash/merge to develop

PR #584, opened 2021-06-08
Use sql transaction to insert actions in the timing database

sebastiencs:
I have tested with the replayer, and the timing thread doesn't lag behind the hooks anymore


tizoc:
Looks good, merging👍

PR #589, opened 2021-06-11
Other indexers support

akoptelov:
We have requirements to support two other indexers, Conseil and Tzkt. These new tests use Tezos and Tezedge nodes bootstrapped to 20000 blocks and run indexers for the whole chain. Also they measure time spent on reaching the last block.


adonagy (approved):
Ok, LGTM, please rebase and merge into develop.

PR #604, opened 2021-06-18
Fix switching to the target branch

akoptelov:
This is to fix a git failure in dhat test


tizoc (approved):

akoptelov:
@adonagy merging it without your approval.

PR #607, opened 2021-06-21
In  `ProtocolRpcError`, include the request when available.

tizoc:


tezos/wrapper/src/service.rs
```diff
@@ -783,7 +791,7 @@ impl ProtocolController {
         &self,
         request: ProtocolRpcRequest,
     ) -> Result<ProtocolRpcResponse, ProtocolServiceError> {
-        self.call_protocol_rpc_internal(ProtocolMessage::ProtocolRpcCall(request))
+        self.call_protocol_rpc_internal(request.clone(), ProtocolMessage::ProtocolRpcCall(request))
```
bkontur, line 794: @tizoc 
I am afraid of cloning the whole request,
why not just clone path/method instead of the whole request data?
I dont know, if to avoid any cloing is possible here

tizoc: @bkontur we only need the path, no need for the rest, it was just a quick hack to get some meaningful output, will fix in a bit.

tezos/api/src/ffi.rs
```diff
@@ -744,7 +744,7 @@ impl From<TezosErrorTrace> for FfiJsonEncoderError {
 
 pub type Json = String;
 
-#[derive(Serialize, Deserialize, Debug)]
+#[derive(Clone, Serialize, Deserialize, Debug)]
```
bkontur, line 747: @tizoc 
and please remove also the unneeded Clone

tizoc: Ok, all done.

bkontur (approved):

PR #608, opened 2021-06-21
Better error handling for timing

sebastiencs:
https://viablesystems.atlassian.net/browse/TE-527

Remove any `unwrap()` in the timing code.
This will log errors on stderr instead.

This also use a fixed capacity channel from the standard library instead of crossbeam.


tizoc (approved):
@sebastiencs looks good. One question, is it possible for the timings thread to "die", leaving a dangling channel?

tizoc:
@sebastiencs ok, good.

PR #611, opened 2021-06-21
Support for "not found" responses from RPC route handler

tizoc:


tizoc:
@bkontur tomorrow I will try to extend this to convert the results from the ocaml router to the correct responses in the Rust side: https://github.com/tezedge/tezedge/blob/master/tezos/api/src/ffi.rs#L673-L684

bkontur:
@tizoc Bruno, do we need this? if yes, please rebase to develop, if not just close and remove branch please

tizoc:
Yes, I think we do, rebasing it now.

tizoc:
Hmmmm I am not entirely sure now, because I added a generic handler for the OCaml side RPC responses, this one was meant to handle the generic `Not_found` that we now consider an error.

PR #613, opened 2021-06-22
Measure memory usage in tezedge context

sebastiencs:
This adds a new `Map` implementation, which replace [im_rc::OrdMap](https://docs.rs/im-rc/15.0.0/im_rc/struct.OrdMap.html).
Implements the method `get_memory_usage` on `ShellContextApi`


bkontur:
@sebastiencs Sebastien, do we need this? if yes, please rebase to develop, if not just close and remove branch please

PR #617, opened 2021-06-22
Make the context backend used by docker-compose runs configurable

tizoc:
It is set by the TEZOS_CONTEXT_BACKEND environment variable and defaults to Irmin.

Also enables the "resources" tab because it is needed to show the storage stats, even if monitoring is not enabled.

### DIFF
--- a/docker-compose.latest.yml
+++ b/docker-compose.latest.yml
@@ -4,7 +4,7 @@ services:
 
   node:
     image: simplestakingcom/tezedge:latest
-    command: ["--network=mainnet", "--p2p-port=9732", "--rpc-port=18732", "--websocket-address=0.0.0.0:4927", "--peer-thresh-low", "30", "--peer-thresh-high", "45"]
+    command: ["--network=mainnet", "--p2p-port=9732", "--rpc-port=18732", "--websocket-address=0.0.0.0:4927", "--peer-thresh-low", "30", "--peer-thresh-high", "45", "--tezos-context-storage=${TEZOS_CONTEXT_STORAGE:-irmin}"]
     hostname: node
     ports:
       - "4927:4927"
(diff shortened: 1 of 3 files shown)


tizoc:
@adonagy @bkontur this change is required so that we can add instructions at the end of the storage stat article for launching the docker images with both contexts. Something like this:

```
$ docker-compose -f docker-compose.latest.yml pull
$ env TEZOS_CONTEXT_BACKEND=both docker-compose -f docker-compose.latest.yml up
```

But if you have a better idea let me know, this is the simplest approach I could think of.

docker-compose.latest.yml
```diff
@@ -18,7 +20,7 @@ services:
     image: simplestakingcom/tezedge-explorer:latest
     hostname: explorer
     environment:
-      API: '[{"id":"localhost","name":"rust.dev.mainnet.localhost","http":"http://localhost:18732", "monitoring":"", "debugger":"", "ws":"ws://localhost:4927", "features":["MONITORING","MEMPOOL_ACTION","STORAGE_BLOCK"]}]'
+      API: '[{"id":"localhost","name":"rust.dev.mainnet.localhost","http":"http://localhost:18732", "monitoring":"", "debugger":"", "ws":"ws://localhost:4927", "features":["MONITORING","MEMPOOL_ACTION","STORAGE_BLOCK","RESOURCES"]}]'
```
bkontur, line 23: @tizoc 
does RESOURCES work without debugger?
because there is no debugger

bkontur (approved):

docker-compose.latest.yml
tizoc: @bkontur what is missing is the monitoring page, which will show up as default and empty. What we want to be able to access is `resources -> storage`. Ideally we would have a flag that lets us enable just that.

tizoc:
@bkontur FYI just saw this failure, but it didn't happen in the next run (this PR doesn't anything related to it): http://ci.tezedge.com/tezedge/tezedge/1067/5/5

I think it is the same test that was failing in some situations before because a lock related failure.

PR #625, opened 2021-06-24
Test TZ against Octez node

akoptelov:


akoptelov:
@adonagy Merging it without your review. Looks like it works )

PR #627, opened 2021-06-24
Add environment variable flag to disable the context GC for tests

tizoc:
Also re-enable both contexts in the drone config files, and disable the GC in florence's indexer pipeline.

### DIFF
--- a/light_node/src/configuration.rs
+++ b/light_node/src/configuration.rs
@@ -628,7 +628,8 @@ pub fn tezos_app() -> App<'static, 'static> {
             .takes_value(true)
             .value_name("STRING")
             .possible_values(&SupportedContextKeyValueStore::possible_values())
-            .help("Choose the merkle storege backend - supported backends: 'inmem', 'inmem-gc', 'btree'"))
+            .help("Choose the TezEdge context storage backend - supported backends: 'inmem'"))
+        // TODO - TE-261: right now this is obsolete, either reintegrate with the timings database or remove
         .arg(Arg::with_name("compute-context-action-tree-hashes")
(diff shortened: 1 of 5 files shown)


light_node/src/configuration.rs
```diff
@@ -1122,6 +1129,12 @@ impl Environment {
                         )
                     });
 
+                let context_kv_store_gc = args
+                    .value_of("context-kv-store-gc")
+                    .unwrap_or("true")
```
bkontur, line 1134: @tizoc @sebastiencs @jurajselep 
I would set false as default, to save the same behavior as we have now,
and if anybody wants GC or later for rolling mode, it has to be changed by startup args,
but default now is **irmin + full/archive mode**

bkontur: @tizoc 
and I would maybe add here some check/panic,
for combination: context-kv-store-gc=true with irmin-only,
just to be supersure

bkontur (approved):

light_node/src/configuration.rs
tizoc: @bkontur I wouldn't even document this flag, doesn't make sense to run the in-memory backend with GC disabled because memory will fill up very quickly, it is just a workaround for the tests and not something users should use, ever.

tizoc:
@bkontur @sebastiencs ok, new version reads environment variable directly from the `InMem` storage constructor, which is hacky, but the patch is much cleaner and will be easier to revert once history modes have been implemented. It doesn't require an updated libtezos-ffi either now.

sebastiencs (approved):

bkontur (approved):

PR #628, opened 2021-06-24
Ci: renaming

adonagy:


.drone.yml
```diff
@@ -1,5 +1,11 @@
 ##############################################################################################################
-# Ocaml stuff is based on image tezos/tezos:v8 (if released new one, fell free to upgrade it here)
+# Octez stuff is based on image tezos/tezos:v8 (if released new one, fell free to upgrade it here)
```
bkontur, line 2: @adonagy v8 is not actual :), just remove/change it tezos/tezos:v9 or tezos/tezos:vXYZ

bkontur (approved):

.drone.yml
adonagy: I did remove the line. On the delphinet tests, we use v8.2 and further down we use v9-release :)

PR #629, opened 2021-06-25
Change snapshot dir for tezedge

akoptelov:


adonagy (approved):
@akoptelov Looks like it works :) Enable the other pipelines and it's good to go

PR #630, opened 2021-06-25
For wrk tests, use the Irmin context

tizoc:
Change the wrk performance test so that the Irmin context is used. Using the TezEdge context without GC is not really a realistic scenario, and by using Irmin in both the OCaml and Rust nodes, we can better understand the difference.


bkontur:
@tizoc Bruno, do we need this? if yes, please rebase to develop, if not just close a remove branch please

tizoc:
Ok, I rebased it locally, will wait until the compressed artifacts things is done, rebase again over that and push.

It is not strictly necessary, because it is working right now (with GC disabled), but we probably want this setup instead, because it is how we are running things in production.

tizoc:
Ok, merging this now, also includes some changes to how the delphinet wrk tests run (waits for the full bootstrap to reduce noise) and an updated libtezos that avoids querying the context to get the `test_chain_status` value.

PR #633, opened 2021-06-29
Implement cache for RPC

vlad9486:


rpc/src/server/mod.rs
```diff
@@ -181,13 +183,26 @@ pub fn spawn_server(
...
+                                        }
                                         let params: Params = params.into_iter().map(|(param, value)| (param.to_string(), value.to_string())).collect();
                                         let query: Query = req.uri().query().map(parse_query_string).unwrap_or_else(HashMap::new);
 
                                         let handler = handler.clone();
                                         let fut = handler(req, params, query, env);
                                         match Pin::from(fut).await {
-                                            Ok(response) => Ok(response),
+                                            Ok(response) => {
+                                                // TODO: more conditions for cache
```
vlad9486, line 198: Maybe we should not cache certain requests

mambisi: Yes.We will also need time to live for endpoints some endpoints

rpc/src/server/mod.rs
```diff
@@ -181,13 +183,26 @@ pub fn spawn_server(
                                 }
                                 _ => {
                                     if allowed_methods.contains(request_method) {
+                                        let uri_string = req.uri().to_string();
```
bkontur, line 186: @vlad9486 @jurajselep @mambisi @tizoc 
this req.uri() caching here is too dangerous, like `/chains/main/blocks/head/*`  is changing all the time, so we could server a bad data,
I would maybe better move this json caching to shell_handler.rs layer:
between:
```
pub async fn chains_block_id_metadata(
    _: Request<Body>,
    params: Params,
    _: Query,
    env: RpcServiceEnvironment,
) -> ServiceResult {
    let chain_id = parse_chain_id(required_param!(params, "chain_id")?, &env)?;
    let block_hash = parse_block_hash(&chain_id, required_param!(params, "block_id")?, &env)?;

// TODO: add request cache here, where uri is transformed to chain_id and exact block_hash

    result_to_json_response(
        base_services::get_block_metadata(&chain_id, &block_hash, &env).await,
        env.log(),
    )
}    
```

bkontur: @vlad9486 
wow `CacheBody::Inner` is used in `result_to_json_response`, so I guess it is a good way,
so I would do that URL-cache matching here, because we need to transform Params/Queury to chain_id/block_hash (and/or we can cache also serveral other RPC uris which does not have chain/block_hash)

PR #634, opened 2021-06-29
Add cache to RPC

mambisi:


rpc/src/services/base_services.rs
```diff
@@ -23,9 +23,15 @@ use tezos_messages::ts_to_rfc3339;
...
+use cached::proc_macro::cached;
+use cached::TimedCache;
+use cached::TimedSizedCache;
+use cached::SizedCache;
+
 /// Retrieve blocks from database.
+#[cached( name="BLOCK_HASH_CACHE",type = "TimedCache<(ChainId,BlockHash), Vec<BlockHash>>", create = "{TimedCache::with_lifespan(20)}", convert = "{(chain_id.clone(),block_hash.clone())}", result = true)]
 pub(crate) fn get_block_hashes(
-    _chain_id: ChainId,
+    chain_id: ChainId,
```
bkontur, line 34: @mambisi 
to aviod this unused warning,
just leave it like before:
```_chain_id: ChainId```

and just in that annotation use:
```{(_chain_id.clone()```

File: rpc/src/rpc_actor.rs [hunk 2]
```diff
@@ -40,6 +41,7 @@ pub struct RpcCollectedState {
 pub struct RpcServer {
     shell_channel: ShellChannelRef,
```
bkontur, line 42: @mambisi 
I guess, we dont need shell_channel here, because it is in env, please remove it

mambisi: Ok

rpc/src/rpc_actor.rs
```diff
@@ -134,6 +141,11 @@ impl Receive<ShellChannelMsg> for RpcServer {
 
     fn receive(&mut self, _ctx: &Context<Self::Msg>, msg: ShellChannelMsg, _sender: Sender) {
         if let ShellChannelMsg::NewCurrentHead(_, block) = msg {
+            // prepare main chain_id
+            let chain_id = parse_chain_id(MAIN_CHAIN_ID, &self.env).unwrap();
+            // warm-up - calls where chain_id + block_hash
+            let _ = crate::services::base_services::get_block_metadata(&chain_id, &block.hash,&self.env);
```
bkontur, line 147: @mambisi 
we need to add here all cached calls,

later we could do those precache calls in parallel with `try_join!` or something

tezos/messages/src/p2p/encoding/version.rs
```diff
@@ -54,3 +55,11 @@ impl PartialEq for NetworkVersion {
             && self.p2p_version == other.p2p_version
     }
 }
+
+impl Hash for NetworkVersion {
```
bkontur, line 59: @mambisi 
does cached needs this?

rpc/src/services/protocol/mod.rs
```diff
@@ -461,7 +465,18 @@ fn handle_rpc_response(
...
+#[cached( name="CALL_PROTOCOL_RPC_CACHE",
+type = "TimedSizedCache<Vec<u8>, serde_json::value::Value>",
+create = "{TimedSizedCache::with_size_and_lifespan(250,20)}",
+convert = "{bincode::serialize(&(chain_param,&chain_id,&block_hash,&rpc_request)).unwrap()}",
+result = true)]
+ */
+#[cached( name="CALL_PROTOCOL_RPC_CACHE",
+type = "TimedSizedCache<(String,ChainId,BlockHash,Vec<u8>), serde_json::value::Value>",
+create = "{TimedSizedCache::with_size_and_lifespan(10,20)}",
+convert = "{(chain_param.to_owned(),chain_id.clone(),block_hash.clone(),serde_json::to_vec(&rpc_request).unwrap())}",
```
bkontur, line 478: @mambisi 
this is still no good, we cannot cache all calls here, 
just those with **http method** `GET`

mambisi: Ok

rpc/src/rpc_actor.rs
mambisi: Ok

rpc/src/rpc_actor.rs
mambisi: We need it

https://github.com/mambisi/tezedge/blob/c440246edbfca0e93954c04ad6a58597d082ad15/rpc/src/server/shell_handler.rs#L297

tezos/messages/src/p2p/encoding/version.rs
mambisi: The NetworkVersion is used as key for cache which requires keys to `impl` `Hash`

rpc/src/rpc_actor.rs
bkontur: @mambisi 
yes, we need it, but it is used from `env.shell`,
I want to just remove it from `pub struct RpcServer {` and where need it, just it it from `env`

tezos/messages/src/p2p/encoding/version.rs
bkontur: aha, ok, I see `get_node_version`, well I guess, this does not need to be cached, but ok

rpc/src/services/protocol/mod.rs
```diff
@@ -461,8 +463,19 @@ fn handle_rpc_response(
...
+#[cached( name="CALL_PROTOCOL_RPC_CACHE",
+type = "TimedSizedCache<Vec<u8>, serde_json::value::Value>",
+create = "{TimedSizedCache::with_size_and_lifespan(250,20)}",
+convert = "{bincode::serialize(&(chain_param,&chain_id,&block_hash,&rpc_request)).unwrap()}",
+result = true)]
+ */
+#[cached( name="CALL_PROTOCOL_RPC_CACHE",
+type = "TimedSizedCache<(String,ChainId,BlockHash,RpcRequest), serde_json::value::Value>",
+create = "{TimedSizedCache::with_size_and_lifespan(10,20)}",
+convert = "{(chain_param.to_owned(),chain_id.clone(),block_hash.clone(),rpc_request.clone())}",
```
bkontur, line 473: @mambisi 
well, this is still no good, rpc_request contains context_path, which could contains `head`,
maybe we should do here a little parsing for that context_path,
like the cache key will be: `chain_id+block_hash+fixed_context_path`

mambisi: ok I get it now

mambisi: I am not sure what the context_path is

tizoc: @mambisi take a look at the `parse_block_hash` function in `rpc/src/helpers.rs` for example, which parses the block fragment (to for example, get a hash from "head").

tizoc: What we need is for example:

- `"/chains/main/blocks/head/helpers/preapply/operations"` -> `"/chains/main/blocks/<some-hash>/helpers/preapply/operations"`
- `"/chains/main/blocks/head~3/helpers/preapply/operations"` -> `"/chains/main/blocks/<some-other-hash>/helpers/preapply/operations"`

etc

Because the meaning of "head" keeps changing all the time, but the meaning of the hash doesn't.

tizoc: Ufff, sorry, I am just making things more confusing. The block hash is already available. Workshop is starting in 3 minutes, we can discuss later. What is important is that the path needs to be normalized so that the `head` part doesn't affect anything.

tizoc: @mambisi so, as @bkontur said, what is needed here is a "fixed" or "normalized" version of the path, because both the chain id and the block has have been obtained already, so what is needed here is some kind of reverse resolution of the URL, maybe hyper already has that?

- `/chains/main/blocks/head/helpers/preapply/operation` -> `/chains/:chain_id/blocks/:block_id/helpers/preapply/operations`
- `/chains/main/blocks/head~3/operation_hashes` -> `/chains/:chain_id/blocks/:block_id/operation_hashes`

etc

@bkontur something I am thinking about is, what about the requests that go through the router? those probably require some pre-parsing too to be correctly cached, right?

tizoc: From looking at https://tezos.gitlab.io/shell/rpc.html and https://tezos.gitlab.io/active/rpc.html#rpc-index, as long as we take care of the `/chains/:chain_id/blocks/:block_id` it seems it should be mostly enough

bkontur: @tizoc @mambisi 
yes, exactly also like Bruno wrote,

- this discussion is just about caching rpc protocol router
- we need to cache just GET request
- all protocol router calls have fixed prefix: `/chains/:chain_id/blocks/:block_id/`  (because just from block we know protocol)
- so cache key should be instead of:
  ```
  convert = "{(chain_param.to_owned(),chain_id.clone(),block_hash.clone(),rpc_request.clone())}"
  ```
 something like this:
  ```
 convert = "{(chain_id.clone(),block_hash.clone(), cache_normalized_url(&rpc_request))}"
  ```
 or maybe just this:
  ```
  convert = "{cache_normalized_url(&rpc_request, &chain_id, &block_hash)}"  // not need to copy chain_id/block_hash
  ```

- that function `cache_normalized_url`  should just replace here `rpc_request.context.path` this two chain_id/block_id
-- maybe just some easy string replacement is enough

@tizoc 
btw. how do we handle rpc mapping in ocaml router for `/chains/main/blocks/head~3/context/constants`?
do you remove everything before `/context/constants`?

tizoc: @bkontur yes, it is ignored, I just use the values sent as part of the request https://gitlab.com/simplestaking/tezos/-/blob/v1.6.0/src/lib_ffi/ffi_rpc_router.ml#L40

tizoc: @bkontur I mean, the structure needs to match, but the values from the fragments are not used for anything.

tizoc: Btw, if every request we send to the OCaml-side contains that prefix already, for the cache key we can just remove it, and use only the remaining part of the path, no need to normalize.

tizoc:
Closing this one, replaced by #638

PR #638, opened 2021-07-01
RPC cache

tizoc:
Built on top of #634 with some tweaks


rpc/src/rpc_actor.rs
```diff
@@ -134,6 +150,50 @@ impl Receive<ShellChannelMsg> for RpcServer {
 
     fn receive(&mut self, _ctx: &Context<Self::Msg>, msg: ShellChannelMsg, _sender: Sender) {
         if let ShellChannelMsg::NewCurrentHead(_, block) = msg {
+            // prepare main chain_id
```
tizoc, line 153: @bkontur I am thinking that during bootstrap maybe we don't want this, what do you think?

bkontur: @tizoc yes, exactly, I was thinking to move `current_bootstrap_state: SynchronizationBootstrapStateRef,` to RpcEnvironment` there is a is_bootstrapped flag

rpc/src/rpc_actor.rs
```diff
@@ -134,6 +150,50 @@ impl Receive<ShellChannelMsg> for RpcServer {
...
+            let chain_id = parse_chain_id(MAIN_CHAIN_ID, &self.env).unwrap();
+            // warm-up - calls where chain_id + block_hash
```
bkontur, line 155: @tizoc 
and another thing is maybe we could trigger this warms in parallel with `futures::try_join!,`
I used `try_join!` in base_service e.g.

tezos/api/src/ffi.rs
```diff
@@ -751,6 +752,58 @@ pub struct RpcRequest {
     pub accept: Option<String>,
 }
 
+fn ffi_routed_subpath(path: &str) -> String {
```
tizoc, line 755: Btw, I don't like this name, suggestions welcome. Maybe `ffi_rpc_router_cache_key` or something like that, dunno.

rpc/src/rpc_actor.rs
tizoc: hmmmm I tried to do that now, but it requires `receive` to be `async` too:

```
error[E0728]: `await` is only allowed inside `async` functions and blocks
```

But that trait is defined by `riker`, and it doesn't use `async`.

@bkontur but many of those functions are `async` and we are not awaiting for their result. I am not familiar with how Rust async works, but if it is like in other similar things I have used,  the `async` ones will all be launched without waiting for the previous results since nothing is being `await`ed

rpc/src/rpc_actor.rs
tizoc: ok, I will add that tomorrow, I need to figure out how (and from where) this is called to add that argument.

rpc/src/rpc_actor.rs
tizoc: Thinking about this more, if those are not awaited, maybe they are not even being executed, so it is possible that this code here is not correct.

tizoc: Ok, just pushed a workaround. One thing is that not all those calls are `async`, maybe something needs to be done about that. Used `join!` instead of `try_join!` because the later seems to cancel everything else after any of the tasks fails.

tizoc: That `block_on` will make this `receive` method block btw, but I don't know the implications of that since I am not familiar with this side of the code, so please take a look tomorrow, maybe this is not the right place to do this.

.drone.yml
```diff
@@ -5187,8 +5187,11 @@ steps:
   image: tezedge/tezos-node-bootstrap:latest
   pull: if-not-exists
   user: root
+  volumes:
+    - name: perf
+      path: /perf
   commands:
-    - tezos-node-bootstrap performance-test --level=$${WRK_TEST_BLOCK_HEADER} --tezedge-old-node=$${TEZEDGE_NODE_TARGET_BRANCH_RPC_CONTEXT_ROOT} --tezedge-new-node=$${TEZEDGE_NODE_RPC_CONTEXT_ROOT} --ocaml-node=$${OCTEZ_NODE_RPC_CONTEXT_ROOT_1} --wrk-test-duration=$${WRK_TEST_DURATION} --max-latency-threshold=100 --throughput-threshold=10
+    - tezos-node-bootstrap performance-test --tezedge-old-node=$${TEZEDGE_NODE_TARGET_BRANCH_RPC_CONTEXT_ROOT} --tezedge-new-node=$${TEZEDGE_NODE_RPC_CONTEXT_ROOT} --ocaml-node=$${OCTEZ_NODE_RPC_CONTEXT_ROOT_1} --wrk-test-duration=$${WRK_TEST_DURATION} --max-latency-threshold=100 --throughput-threshold=10 --latency-no-fail --throughput-no-fail --url-file=/perf/wrk-rpc-list-delphi.txt
```
bkontur, line 5194: @akoptelov  @tizoc 
is this change ok?

tezos/api/src/ffi.rs
bkontur: @tizoc hehe, just add please some docs/comments for function

tezos/api/src/ffi.rs
```diff
@@ -751,6 +752,58 @@ pub struct RpcRequest {
...
+    let base = Url::parse("http://tezedge.com")
```
bkontur, line 756: @tizoc 
could we get better rid of that expect/unwraps here?

rpc/src/rpc_actor.rs
```diff
@@ -129,11 +145,69 @@ impl Actor for RpcServer {
...
+    let _ = futures::join!(
+        get_block_metadata,
+        get_block,
+        get_block_operations_metadata,
+        get_block_operation_hashes,
+        get_block_header,
+    );
+
+    // Sync calls
+    let _ = crate::services::base_services::get_additional_data(
```
bkontur, line 182: @tizoc 
Bruno, add this `base_services::get_additional_data`  at the begining, to be cached first,
because all other calls use this internally, so once this is cached, all other will be faster and get it from cache :)

bkontur: @tizoc 
and we should the the same for calls:

```
BlockStorage::new(persistent_storage).get_with_json_data(
``

File: rpc/src/rpc_actor.rs (hunk 2 above)
Comment (bkontur) [reply rpc/src/rpc_actor.rs:208]: @tizoc 
maybe instead: `futures::executor::block_on` we should move to rpc_actor and use: `tokio_executor: &Handle`

Comment (bkontur) [reply rpc/src/rpc_actor.rs:208]: @tizoc 
and I dont know, if it is possibility to set sometime here: `futures::executor::block_on`,
tokio has something like `tokio::time::timeout` and we use it somewhere for async/await

File: .drone.yml (hunk 4 above)
Comment (akoptelov) [reply .drone.yml:5194]: Yes, this makes the new `tezedge/tezos-node-bootstrap` work for this step. I forgot to update this line, now this is in synch with this one:
https://github.com/tezedge/tezedge/blob/26aeaa32cb2fdf53187d1e278b541799e233ee83/.drone.yml#L5170

File: tezos/api/src/ffi.rs [hunk 7]
```diff
@@ -751,6 +752,58 @@ pub struct RpcRequest {
...
+    // /chain/:chain_id/blocks/:block_id
+    let (chain, _, blocks, _) = (
+        segments.next(),
+        segments.next(),
+        segments.next(),
+        segments.next(),
+    );
+
+    match (chain, blocks) {
+        (Some("chain"), Some("blocks")) => (),
```
mambisi, line 778: @tizoc  the subpath is suppose to be `chains/<chain_id>/blocks/<block_id>`, so it should be 
```
(Some("chains"), Some("blocks"))
```

mambisi: https://tezos.gitlab.io/shell/rpc.html

tezos/api/src/ffi.rs
tizoc: It is in `RpcRequest::ffi_routed_subpath` this one here was just extracted out so that the tests don't need to construct the whole `RpcRequest` just to test this. Btw I am going to rename it to `ffi_rpc_router_cache_key` to make it very explicit that it is why this function exists.

tezos/api/src/ffi.rs
tizoc: @bkontur yes, I can just return the original and fail silently here, but if this call fails for any reason something has gone *terribly* wrong 👀

.drone.yml
tizoc: @akoptelov @bkontur other than what Alexander mentioned, there is one thing I changed here that may be controversial, I added `--latency-no-fail --throughput-no-fail` so that the test finishes to completion even when there is a latency or throughput regression (without those, the test will fail at the first regression and not continue).

tezos/api/src/ffi.rs
tizoc: fixed, good catch

rpc/src/rpc_actor.rs
tizoc: @bkontur ok, I made this change (the `is_bootsrapped` value was available on the context where this message is produced, so I just used that): https://github.com/tezedge/tezedge/pull/638/commits/d0ffc329df00047d64a24db85f0d82e71831b865

rpc/src/rpc_actor.rs
tizoc: Added a timeout here: https://github.com/tezedge/tezedge/pull/638/commits/12fdbe096848b93217a33e6ecff2ac532f6c31f0

I don't know if 3 seconds is too little or too much. Will wait for drone and see if the warning shows up on the logs or not.

tizoc: Ok, turns out none of this works. Python tests started failing, so I checked the logs.

I get this from the `block_on` call:

```
cannot execute `LocalPool` executor from within another executor: EnterError
```

and this when adding the timeout:

```
there is no reactor running, must be called from the context of a Tokio 1.x runtime
```

I don't know yet how to interpret this, but overall it looks like a different approach is needed to await for those futures.

tizoc: Oh, just found `tokio_runtime.block_on`, trying that.

tizoc: @bkontur something that would be need for this to work is https://docs.rs/tokio/1.7.1/tokio/runtime/struct.Handle.html#method.block_on

so that I can do:

```rust
            // warm-up - calls where chain_id + block_hash
            if is_bootstrapped {
                let tokio_runtime_handle = tokio::runtime::Handle::current();
                if let Err(err) = tokio_runtime_handle.block_on(tokio::time::timeout(
                    RPC_WARMUP_TIMEOUT,
                    warm_up_rpc_cache(&chain_id, &block, &self.env),
                )) {
                    warn!(
                        self.env.log(),
                        "RPC warmup timeout after {:?}: {:?}", RPC_WARMUP_TIMEOUT, err
                    );
                }
            }
```

But we are using 1.2.0, and that method was added to 1.4.0.

Is there anything that keeps us at 1.2.0?

tizoc: Well, I think that would not work either, I tried similar variations that are possible in 1.2.0, but the error is always something related to:

```
there is no reactor running, must be called from the context of a Tokio 1.x runtime
```

I don't fully understand how riker and tokio relate, but I guess riker actors don't run inside the tokio runtime context.

tizoc: Passing the handle from light_node's main and using that seems to work. I don't really like how unstructured it is.

```rust
                self.tokio_executor.spawn(async move {
                    tokio::time::timeout(
                        RPC_WARMUP_TIMEOUT,
                        warm_up_rpc_cache(chain_id, block, env),
                    )
                });
 ```
 
 Btw, @mambisi suggested I use `tokio::join!` instead of the version in `futures`, which I guess makes sense. Why do other parts of the code use the `futures` one?

**Comment by tizoc** [comment]
http://ci.tezedge.com/tezedge/tezedge/1209/7/9 failed, but not sure why, cannot reproduce it locally.

File: rpc/src/rpc_actor.rs (hunk 2 above)
Comment (bkontur) [reply rpc/src/rpc_actor.rs:223]: @tizoc 
wow, you just extended `ShellChannelMsg::NewCurrentHead` with `is_bootstrapped`, very good, very good :)

Comment (bkontur) [reply rpc/src/rpc_actor.rs:223]: @tizoc @mambisi 
yes, we should better use tokio::try_join/join, but we should move `tokio_executor: Handle` to RpcServiceEnvironment, and I am not sure if should also wrap it with Arc:
```
tokio_executor: Arc<Handle>
```

Comment (bkontur) [reply rpc/src/rpc_actor.rs:223]: @tizoc 
yes, I saw that tokio is 1.7 now, we should upgrade deps https://viablesystems.atlassian.net/browse/TE-492
but I want to do it better after release, hope fully today

Comment (bkontur) [reply rpc/src/rpc_actor.rs:223]: @tizoc @mambisi 
ok, I am moving now tokio_executor to env and replaceing the futures::try_join

File: rpc/src/services/base_services.rs [hunk 8]
```diff
@@ -227,8 +298,15 @@ pub(crate) fn get_block_protocols(
 }
 
 /// Extract the current_protocol and the next_protocol from the block metadata
+#[cached(
+    name = "BLOCK_ADDITIONAL_DATA_CACHE",
+    type = "TimedSizedCache<(ChainId, BlockHash), Option<BlockAdditionalData>>",
```
bkontur, line 303: @tizoc @mambisi @jurajselep 
I am not sure about cachcing Optional here,
lets say situation:
- we have current head on **5**
- we receive request /chains/main/blocks/**6**/additional_data
-- so we add to cache: None
- then we apply block **6** + we trigger warmup for 6
- we receive another call /chains/main/blocks/**6**/additional_data

what does that second call /chains/main/blocks/**6**/additional_data returns?
I guess None, instead of Some(data)

how does it cache Optional, I guess, we should not cache optional,
I am going to add some tests to drone CI, to cover something like this and see

bkontur: @tizoc @mambisi @jurajselep 
 there should be a cfg for optionals:

image

mambisi: Its using a Hashmap under the hood so 
```
let mut map = HashMap::new();
map.insert(("Nice".to_owned(), Some("Game".to_string())), 2);
map.insert(("Nice".to_owned(), None), 2);
```

Output:
```
{
    (
        "Nice",
        Some(
            "Game",
        ),
    ): 2,
    (
        "Nice",
        None,
    ): 2,
}
```

bkontur: @tizoc @mambisi  @jurajselep 
hmm, there is no possibility "not to cache Ok(None)", 
I have a simple test for that

rpc/src/services/base_services.rs
```diff
@@ -23,7 +23,22 @@ use tezos_messages::ts_to_rfc3339;
...
+use cached::proc_macro::cached;
+use cached::SizedCache;
+use cached::TimedCache;
+use cached::TimedSizedCache;
+
+pub const TIMED_SIZED_CACHE_SIZE: usize = 10;
+pub const TIMED_SIZED_CACHE_TTL_IN_SECS: u64 = 60;
+
 /// Retrieve blocks from database.
+#[cached(
```
bkontur, line 35: @tizoc @mambisi @jurajselep 
I am removing this from cache,
because, if the first call returns empty vec,
the next call will be cached for the empty, which is no good,
and also this is not even re-cached or warm-up-ed,


I dont know how this could be done by these annotations,
this will need to be cached (programatically) only if the result is ok and not empty vec,

I am collecting TODO stuff here

rpc/src/services/base_services.rs
tizoc: @bkontur since we can use any structure we want for cache, it should just be a matter of using one that ignores anything that results in `None`.

tizoc: @bkontur btw, if we receive a request for `6`, but `6` has not been processed yet, cannot that be considered an error?

I am not very clear on the logic of how this should work. After `6` gets applied, is the additional data for it fixed forever? or is there anything that could update that data?

I am thinking that maybe caching the `None` is ok, what is actually needed is to remove the value from the cache once block `6` is applied, and before the warmup function is called (that `None` should be a valid result until then anyway).

tizoc: Still trying to figure out the code that `cached` generates for these annotations, but what I have in mind is (roughly) something like this:

```
super::base_services::BLOCK_ADDITIONAL_DATA_CACHE.get_mut().unwrap().cache_remove(&(chain_id.clone(), block.hash.clone()));
```

bkontur: > 
> 
> Still trying to figure out the code that `cached` generates for these annotations, but what I have in mind is (roughly) something like this:
> 
> ```
> super::base_services::BLOCK_ADDITIONAL_DATA_CACHE.get_mut().unwrap().cache_remove(&(chain_id.clone(), block.hash.clone()));
> ```

@tizoc 
I was thinking exactly the same, to access that caches, but it is another coding,
so I refactored `get_adiition_data->Result<Optional`  to `get_additional_data_or_fail<Result<BLockAdditionData`
like you suggested before comment :)

tizoc: @bkontur yes that is my preferred solution, feels more robust than manually invalidating the cache.

bkontur: @tizoc 
and what about things we put into the cache,
shouldnt we wrap them with `Arc` to aviod cloning, or this is not a issue?

tizoc: @bkontur very good point. For small values I guess cloning is not that big of a deal, but some of the values we cache are huge. Unless you did already, I will try adding that.

tizoc: @bkontur actually, `get` methods borrow values, no cloning happens: https://github.com/jaemk/cached/blob/12858a8a4af7f278e32ae90383d7483a54f829c6/src/lib.rs#L475-L479

And caches are wrapped by a `Mutex`, so no extra synchronization is needed.

tizoc: Wait, actually, that is still wrong (some cloning needs to be happening to return the results from the cached version of the function), let me check what the macro does exactly.

tizoc: Ok, it says so here: https://github.com/jaemk/cached/blob/12858a8a4af7f278e32ae90383d7483a54f829c6/src/lib.rs#L248

Adding `Arc` now (for the big ones at least).

rpc/src/services/base_services.rs
```diff
@@ -146,8 +166,15 @@ pub(crate) async fn get_block_header(
 }
 
 /// Get information about block shell header
+#[cached(
+    name = "BLOCK_SHELL_HEADER_CACHE",
+    type = "TimedSizedCache<(ChainId, BlockHash), Option<BlockHeaderShellInfo>>",
```
tizoc, line 171: @bkontur same situation here regarding the `Option` value, right?

bkontur: @tizoc 
uff, thx, I missed that

rpc/src/services/base_services.rs
```diff
@@ -156,23 +183,30 @@ pub(crate) fn get_block_shell_header(
         .map(|header| BlockHeaderShellInfo::new(&header)))
 }
 
+#[cached(
+    name = "LIVE_BLOCKS_CACHE",
+    type = "TimedSizedCache<(ChainId, BlockHash), Vec<String>>",
```
tizoc, line 188: @bkontur about the ones that return `Vec`, is the situation the same for all of them? (empty vec meaning that we don't have data yet, so it shouldn't be cached).

rpc/src/services/base_services.rs
tizoc: Just pushed the `Arc` wrappers for most (but not all) of the cached functions.

tizoc: re-pushed, got eaten by your last push.

rpc/src/services/base_services.rs
bkontur: @tizoc @mambisi @jurajselep 
well, it depends, there are RPCs like `/mempool/pendings`  which retursn list, and it returns dynamic data, so this should never be cached,

and there are static RPC, which return applied block's data, which are once evaluated, they are unchanged,
but if we have RPC that return Vec and we dont check, if block is applied, but just tries to read some data: `some_storage.get_vec_data(block_hash: &BlockHash) -> Result<Vec<Data>>`,
then we can have problem and cache data, before they are applied, so after apply if we also miss warm-up, we can serve invalid data: empty vs not empty
but in this case, if we `some_storage.get_vec_data` find some Vec data for block, we assume, that have it all and can be cached now

.drone.yml
bkontur: @tizoc @akoptelov 
what about this? do we need to do something more?
if we have any regression, will this tests fail somehow or not?

bkontur (approved):

.drone.yml
tizoc: @bkontur something better is needed. Latencies are not sable enough for this to not randomly fail when somebody starts a branch that changes something completely unrelated. Also, we cannot improve everything all the time, sometimes tradeoffs are needed (and sometimes numbers may be good because they are producing the wrong answer fast, and fixing that will cause a regression).

I just don't like the current model that considers everything a failure, I think it is the wrong approach. What we need is some good visualization, and it will be up to an human to decide if the changes are acceptable or not.

bkontur: @tizoc @akoptelov 
ok, could you please create a jira under to tech.debt for that latencies, latencies are hot issue

tizoc: Missed this yesterday. Ok, will do.

PR #641, opened 2021-07-02
Custom serialization and inline small values

sebastiencs:
- Custom serialization.
  The functions `serialize_entry` and `deserialize_entry` have been fuzzed without issue.
- Uses indexes/ids instead of shared pointers
- Inline small blobs during serialization, their hash won't be store anywhere.
  Hashes for those need to be recomputed
~~- Adds statistics & memory usage.
  The method `TezedgeContext.get_memory_usage` will need to be called from ocaml and the data sent to the timing thread~~
- Adds statistics & memory usage, send it to the timing thread which will store them in the database


tezos/new_context/src/working_tree/storage.rs
```diff
@@ -0,0 +1,574 @@
...
+}
+
+#[derive(Debug, PartialEq, Eq)]
+enum BlobRef {
+    Inline { length: u8, value: [u8; 7] },
+    Ref { start: usize, end: usize },
+}
+
+impl BlobStorageId {
+    pub fn new_inline(value: &[u8]) -> Result<Self, StorageIdError> {
```
tizoc, line 147: @sebastiencs these two would be better named `try_new_inline` and `try_new` to make it explicit that they can fail.

tezos/new_context/src/lib.rs
```diff
@@ -347,3 +324,37 @@ mod tests {
...
+    }
+    fn write_u32(&mut self, n: u32) {
+        self.0 = n as u64
+    }
+    fn write_u16(&mut self, n: u16) {
+        self.0 = n as u64
+    }
+}
+
+/// A map, without hashing
```
tizoc, line 359: Please add some comments explaining why this exists.

tezos/new_context/src/tezedge_context.rs
```diff
@@ -42,7 +43,7 @@ pub struct PatchContextFunction {}
 pub struct TezedgeIndex {
     pub repository: Arc<RwLock<ContextKeyValueStore>>,
     pub patch_context: Rc<Option<BoxRoot<PatchContextFunction>>>,
-    pub strings: Rc<RefCell<StringInterner>>,
+    pub storage: Rc<RefCell<Storage>>,
```
tizoc, line 46: Adding comments for this struct briefly explaining repository and storage would be good.

tezos/new_context/src/tezedge_context.rs
```diff
@@ -57,25 +58,24 @@ impl TezedgeIndex {
...
     pub fn find_entry_bytes(&self, hash: HashId) -> Result<Option<Vec<u8>>, DBError> {
         let repo = self.repository.read()?;
         Ok(repo.get_value(hash)?.map(|v| v.to_vec()))
     }
 
-    pub fn find_entry(&self, hash: HashId) -> Result<Option<Entry>, DBError> {
+    pub fn find_entry(
+        &self,
+        hash: HashId,
+        storage: &mut Storage,
```
tizoc, line 74: Can you explain why this `&mut Storage` argument is needed? it is part of `self`, so I guess the reason is that it needs to be `&mut`, but why is that?

tezos/new_context/src/working_tree/serializer.rs
```diff
@@ -0,0 +1,591 @@
...
+                        let kind = descriptor.kind();
+                        let blob_inline_length = descriptor.blob_inline_length() as usize;
+
+                        let node = if blob_inline_length > 0 {
+                            // The blob is inlined
+
+                            let blob = data
+                                .get(pos..pos + blob_inline_length)
+                                .ok_or(UnexpectedEOF)?;
+                            let blob_id = storage.add_blob_by_ref(blob)?;
```
tizoc, line 330: Ok, I see this is why `&mut Storage` is needed here (and elsewhere). But why is deserialization adding new values?

tezos/new_context/src/tezedge_context.rs
```diff
@@ -661,21 +699,35 @@ impl ShellContextApi for TezedgeContext {
...
+        } = self.tree.prepare_commit(
             date,
             author,
             message,
             self.parent_commit_hash,
             &mut *repository,
             true,
         )?;
+
+        // TODO: Store serialize_stats somewhere
```
tizoc, line 717: For this one open a Jira issue so that it doesn't get forgotten, and update the comment to `TODO - TE-NNN: ...`

tezos/new_context/src/tezedge_context.rs
```diff
@@ -717,12 +769,28 @@ impl ShellContextApi for TezedgeContext {
...
+    fn get_memory_usage(&self) -> Result<ContextMemoryUsage, ContextError> {
+        let repository = self.index.repository.read()?;
+        let storage = (&*self.index.storage).borrow();
+
+        let usage = ContextMemoryUsage {
+            repo: repository.memory_usage(),
+            storage: storage.memory_usage(),
+        };
+
+        println!("usage={:?}", usage);
```
tizoc, line 781: Remove these prints that were used for collecting data on the storage deploy please.

tizoc: Btw, I am thinking that it may be useful to just log this data regularly (but using `slog` like the rest of the application) like we do for other things.

tezos/new_context/src/working_tree/mod.rs
```diff
@@ -86,66 +57,169 @@ pub struct Commit {
...
+            inner: Cell::new(
+                NodeInner::new()
+                    .with_commited(false)
+                    .with_node_kind(node_kind)
+                    .with_entry_hash_id(0)
+                    .with_entry_available(true)
+                    .with_entry_id(match entry {
+                        Entry::Tree(tree_id) => tree_id.into(),
+                        Entry::Blob(blob_id) => blob_id.into(),
+                        Entry::Commit(_) => panic!(), // A Node never contains a commit
```
tizoc, line 149: I think this should be `unreachable!()` instead.

tezos/new_context/src/working_tree/mod.rs
```diff
@@ -86,66 +57,169 @@ pub struct Commit {
...
+            Entry::Blob(blob_id) => {
+                let blob_id: u64 = (*blob_id).into();
+                if blob_id != inner.entry_id() || !inner.entry_available() {
+                    inner.set_entry_available(true);
+                    inner.set_node_kind(NodeKind::Leaf);
+                    inner.set_entry_id(blob_id);
+                }
+            }
+            Entry::Commit(_) => {
+                panic!()
```
tizoc, line 203: Same here. But are we 100% sure this can never be reached under any circumstance?

tezos/new_context/src/working_tree/serializer.rs
```diff
@@ -0,0 +1,591 @@
+use std::{
```
tizoc, line 1: This one is missing the copyright comment.

tezos/new_context/src/working_tree/storage.rs
```diff
@@ -0,0 +1,574 @@
+use std::{
```
tizoc, line 1: Copyright comment.

tezos/new_context/src/working_tree/storage.rs
```diff
@@ -0,0 +1,574 @@
...
+#[derive(Debug)]
+pub struct StorageMemoryUsage {
+    node_cap: usize,
+    trees_cap: usize,
+    temp_tree_cap: usize,
+    blobs_cap: usize,
+    strings: StringsMemoryUsage,
+}
+
+pub struct Storage {
```
tizoc, line 254: This one needs a comment explaining its purpose and where it is used. I now have the impression that it is used for keeping the working tree data (because it is cleared after the commit), but I am still not sure I fully understand it.

tizoc (changes requested):
@sebastiencs I still need to review some other parts when I have time, and give the whole thing a few extra review passes. Overall it is looking very good, but I left a few comments. Will come back to it once I am free from other things.

tezos/new_context/src/gc/worker.rs
```diff
@@ -161,29 +161,8 @@ impl GCThread {
...
-                            Some(hash) => hash,
-                            None => continue,
-                        });
-                    }
-                }
-                Entry::Commit(commit) => {
-                    // Push the root tree for this commit
-                    reused.push(commit.root_hash);
-                }
+            for hash_id in iter_hash_ids(&value) {
```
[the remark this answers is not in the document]
sebastiencs: Yes, it will not allocate anything. Previously it was allocating for the keys `Rc<str>` and nodes `Rc<Node>`

tezos/new_context/src/kv_store/mod.rs
```diff
@@ -23,38 +20,48 @@ pub mod stats;
 pub const INMEM: &str = "inmem";
 
 #[derive(Debug, Clone, Copy, Hash, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize)]
-pub struct HashId(NonZeroUsize); // NonZeroUsize so that `Option<HashId>` is 8 bytes
+pub struct HashId(NonZeroU32); // NonZeroU32 so that `Option<HashId>` is 4 bytes
```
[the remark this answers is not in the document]
sebastiencs: I see. I don't know what is the archive mode, will it require to have a unique `HashId` without the possibility to reuse them ?

tezos/new_context/src/lib.rs
sebastiencs: Ok

tezos/new_context/src/working_tree/serializer.rs
sebastiencs: deserialization requires `&mut Storage` because it extracts values from the serialized bytes, such as the strings, `Node`, blobs, etc..
The working tree is able to work with values from `Storage` only, not from serialized bytes directly.
I will add some comments

tezos/new_context/src/tezedge_context.rs
sebastiencs: Using `slog` would be great, right now every logs from the working tree is using `println`

tezos/new_context/src/working_tree/mod.rs
sebastiencs: As far as I know, a `Node` contains only a `Tree` or a `Blob`.
The enum `Entry` has a `Commit` member only to be able to (de)serialize easily between those 3 types.
I could remove `Commit` from `Entry` but then we would need 2 new functions: `serialize_commit` and `deserialize_commit`

sebastiencs:
@tizoc Your comments were addressed, I added some documentation, let me know if something is not clear.

> The method TezedgeContext.get_memory_usage will need to be called from ocaml and the data sent to the timing thread

About that, actually I think we could use the channel `TIMING_CHANNEL` directly from the working tree, to send the data to the timing thread. The data doesn't need to go to ocaml.
`TIMING_CHANNEL` can also be used to send the serializing statistics after every commits

sebastiencs:
@tizoc [`d4eda05` (#641)](https://github.com/tezedge/tezedge/pull/641/commits/d4eda05693814de2fad30b56c4a8ab8daff6e8cf) will send the memory usage to the timing thread and store it in the database on every commits

tezos/new_context/src/working_tree/storage.rs
```diff
@@ -251,11 +254,33 @@ pub struct StorageMemoryUsage {
     strings: StringsMemoryUsage,
 }
 
+/// `Storage` contains all the data from the working tree.
+///
+/// This is where all trees/blobs/strings are allocated.
+/// The working tree only has access to ids which refer to data inside `Storage`.
+///
+/// Because `Storage` is for the working tree only, it is cleared before
+/// every checkout.
```
tizoc, line 256: Excellent, this along the comments for each field make it much easier to understand👍

tezos/new_context/src/ffi.rs
```diff
@@ -540,14 +538,23 @@ ocaml_export! {
         rt,
         context: OCamlRef<DynBox<TezedgeContextFFI>>,
         value: OCamlRef<OCamlBytes>
-    ) -> OCaml<DynBox<WorkingTreeFFI>> {
+    ) -> OCaml<Result<DynBox<WorkingTreeFFI>, String>> {
```
tizoc, line 541: Ok, before merging I have to update this on the OCaml side, almost forgot.

tezos/new_context/src/tezedge_context.rs
```diff
@@ -387,24 +432,31 @@ impl TezedgeIndex {
...
-            Entry::Blob(blob) => {
+            Entry::Blob(blob_id) => {
                 // push key-value pair
-                entries.push((self.string_to_key(path), blob.clone()));
+                let blob = storage.get_blob(*blob_id)?;
+                entries.push((self.string_to_key(path), blob.to_vec()));
                 Ok(())
             }
             Entry::Tree(tree) => {
                 // Go through all descendants and gather errors. Remap error if there is a failure
```
tizoc, line 445: @sebastiencs lets remove this comment from here, I have no idea what it is about and when it was written, just makes things more confusing.

tezos/new_context/src/ffi.rs
sebastiencs: Ah yes, thanks

tezos/new_context/src/working_tree/storage.rs
```diff
@@ -0,0 +1,607 @@
...
+    string_interner::{StringId, StringInterner},
+    Node,
+};
+
+#[bitfield]
+#[derive(Clone, Copy, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]
+pub struct TreeStorageId {
+    /// Note: Must fit in NodeInner.entry_id (61 bits)
+    #[skip]
+    __: B14,
```
tizoc, line 26: @sebastiencs what is this? can you add a comment? or give it a meaningful name, prefixed with `__` to make it explicit that it is not used (at least that is how I am interpreting it now)

tezos/new_context/src/working_tree/working_tree.rs
```diff
@@ -139,7 +148,31 @@ impl TreeWalkerLevel {
         let children_iter = if should_continue {
             if let WorkingTreeValue::Tree(tree) = &root.value {
                 // TODO: can this clone be avoided?
```
tizoc, line 150: outdated comment

tezos/new_context/src/working_tree/working_tree.rs
```diff
@@ -139,7 +148,31 @@ impl TreeWalkerLevel {
...
-                Some(tree.clone().into_iter())
+
+                let storage = root.index.storage.borrow();
+                let tree = match storage.get_tree(*tree) {
+                    Ok(tree) => tree,
+                    _ => {
+                        // TODO: Handle this error in a better way
+                        eprintln!("TreeWalkerLevel Error: TreeNotFound");
```
tizoc, line 157: Can we at least print the path here? as-is this error would not help much

tezos/new_context/src/working_tree/working_tree.rs
```diff
@@ -139,7 +148,31 @@ impl TreeWalkerLevel {
...
+                    }
+                };
+
+                let mut tree_vec = Vec::with_capacity(tree.len());
+                for (key_id, node_id) in tree {
+                    let key = match storage.get_str(*key_id) {
+                        Ok(key) => key.to_string(),
+                        Err(e) => {
+                            // TODO: Handle this error in a better way
+                            eprintln!("TreeWalkerLevel Error: {:?}", e);
```
tizoc, line 168: Path on this one too.

tezos/new_context/src/working_tree/working_tree.rs
```diff
@@ -194,7 +227,7 @@ impl Iterator for TreeWalker {
 
                     if let Some((k, node)) = iter.next() {
                         // TODO: what to do with errors here?
```
tizoc, line 229: Like you did above, please print an error here, now that we capture all output from the protocol runners it will at least show up in the logs.

tizoc:
@sebastiencs ok, I don't think I have anything else for add. Please squash into fewer commits combined in a way that makes sense (or even a single commit works). Once I am done updating the OCaml side and a new binary has been produced, then we can merge it.

pventuzelo:
@sebastiencs I'm still getting the same leak on Valgrind with the new commit (049c7cda8cf031398d18c387eb6f040ce447a98d). Is it fixed?

tizoc (approved):

PR #644, opened 2021-07-02
Handle duplicate addresses [TE-490]

akoptelov:
Now `P2pPeers` internally maintains the set of peer addresses and report the error when trying to add a peer with already existing address.

### DIFF
--- a/shell/src/peer_manager.rs
+++ b/shell/src/peer_manager.rs
@@ -1417,6 +1417,7 @@ pub mod tests {
 
         let PeerState { peer_id, .. } =
             test_peer(&actor_system, network_channel.clone(), &tokio_runtime, 7777);
+
         // add peer 1 as incoming
         {
             // try permit for incoming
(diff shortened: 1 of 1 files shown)


akoptelov:
OK, not sending `Nack` yet.

shell/src/peer_manager.rs
```diff
@@ -1059,6 +1063,69 @@ fn resolve_dns_name_to_peer_address(
...
+        self.peers.len()
+    }
+
+    fn try_insert(
+        &mut self,
+        peer_ref: PeerRef,
+        peer_address: SocketAddr,
+        bootstrap_requested_last: Option<Instant>,
+    ) -> Result<(), PeerManagerError> {
+        if self.addr_set.contains(&peer_address) {
```
bkontur, line 1095: @akoptelov 
please check comment in jira TE-490 
we should better handle unique connections by public_key and not by socket_address, 
because we can run more nodes from one IP with different port settings lets say

so I would just extend `P2pPeerState` with public_key and check uniquess based on this attribute, so we probably dont need P2pPeersMap, but I like the impl. `try_insert`

bkontur (changes requested):
@akoptelov 
please check comments

PR #645, opened 2021-07-02
add -disable-blacklist command line

dkuehr:
Adds new `-disable-blacklist` command line option that allows misbehavior from remote-peers. For testing/fuzzing purposes.


light_node/src/configuration.rs
```diff
@@ -423,6 +423,10 @@ pub fn tezos_app() -> App<'static, 'static> {
             .long("disable-mempool")
             .global(true)
             .help("Enable or disable mempool"))
+        .arg(Arg::with_name("disable-blacklist")
```
tizoc, line 426: I vote for `disable-peer-blacklist` and "Disable peer blacklisting" as the description.

PR #649, opened 2021-07-05
Implement message throttling

akoptelov:
- Now we have separate quota for each message send and receive
- It is reset each period of time (5s by default)
- It is possible to configure quotas and completely disable it (but have it calculated, so we can see message rate)

### DIFF
--- a/networking/src/p2p/peer.rs
+++ b/networking/src/p2p/peer.rs
@@ -14,6 +14,7 @@ use slog::{debug, info, o, trace, warn, Logger};
 use tokio::io::AsyncWriteExt;
 use tokio::net::TcpStream;
 use tokio::runtime::Handle;
+use tokio::sync::Notify;
 use tokio::time::timeout;
 
 use crypto::{
@@ -35,7 +36,10 @@ use tezos_messages::p2p::encoding::ack::{NackInfo, NackMotive};
 use tezos_messages::p2p::encoding::prelude::*;
 
 use crate::p2p::network_channel::NetworkChannelMsg;
+use crate::p2p::peer::quota::get_reset_period;
 use crate::{LocalPeerInfo, PeerId};
+
+use self::quota::ThrottleQuota;
 
 use super::network_channel::{NetworkChannelRef, NetworkChannelTopic, PeerMessageReceived};
 use super::stream::{EncryptedMessageReader, EncryptedMessageWriter, MessageStream, StreamError};
@@ -206,6 +210,8 @@ struct Network {
     socket_address: SocketAddr,
 }
 
+mod quota;
+
 pub type PeerRef = ActorRef<PeerMsg>;
 
 /// Represents a single p2p peer.
@@ -222,6 +228,8 @@ pub struct Peer {
     peer_id_marker: String,
     peer_metadata: MetadataMessage,
     peer_compatible_network_version: NetworkVersion,
+    throttle_quota: Arc<std::sync::Mutex<quota::ThrottleQuota>>,
+    quota_update_stop: Arc<Notify>,
 }
 
 impl Peer {
@@ -232,17 +240,28 @@ impl Peer {
         network_channel: NetworkChannelRef,
         tokio_executor: Handle,
         info: BootstrapOutput,
+        log: &Logger,
     ) -> Result<PeerRef, CreateError> {
         sys.actor_of_props(
             peer_actor_name,
-            Props::new_args::<Peer, _>((network_channel, tokio_executor, info)),
+            Props::new_args::<Peer, _>((
+                network_channel,
+                tokio_executor,
+                info,
+                log.new(o!("peer_uri" => peer_actor_name.to_string())),
+            )),
         )
     }
 }
 
-impl ActorFactoryArgs<(NetworkChannelRef, Handle, BootstrapOutput)> for Peer {
+impl ActorFactoryArgs<(NetworkChannelRef, Handle, BootstrapOutput, Logger)> for Peer {
     fn create_args(
-        (event_channel, tokio_executor, info): (NetworkChannelRef, Handle, BootstrapOutput),
+        (event_channel, tokio_executor, info, log): (
+            NetworkChannelRef,
+            Handle,
+            BootstrapOutput,
+            Logger,
+        ),
     ) -> Self {
         Peer {
             network_channel: event_channel,
@@ -257,6 +276,8 @@ impl ActorFactoryArgs<(NetworkChannelRef, Handle, BootstrapOutput)> for Peer {
             peer_id_marker: info.3,
             peer_metadata: info.4,
             peer_compatible_network_version: info.5,
+            throttle_quota: Arc::new(std::sync::Mutex::new(ThrottleQuota::new(log))),
+            quota_update_stop: Arc::new(Notify::new()),
         }
     }
 }
@@ -266,9 +287,31 @@ impl Actor for Peer {
 
     fn post_stop(&mut self) {
         self.net.rx_run.store(false, Ordering::Release);
+        self.quota_update_stop.notify_one();
     }
 
     fn pre_start(&mut self, ctx: &Context<Self::Msg>) {
+        // quota replenish task
+
+        let throttle_quota = self.throttle_quota.clone();
+        let log = ctx.system.log().clone();
+        let stop = self.quota_update_stop.clone();
+        self.tokio_executor.spawn(async move {
+            loop {
+                tokio::select! {
+                    _ = tokio::time::sleep(Duration::from_millis(get_reset_period())) => {
+                        match throttle_quota.lock() {
+                            Ok(ref mut quota) => quota.reset_all(),
+                            Err(e) => warn!(log, "Failed to obtain a lock on throttling quota"; "reason" => format!("{:?}", e)),
+                        }
+                    }
+                    _ = stop.notified() => {
+                        return;
+                    }
+                }
+            }
+        });
+
         let myself = ctx.myself();
         let system = ctx.system.clone();
         let net = self.net.clone();
@@ -277,6 +320,7 @@ impl Actor for Peer {
         let peer_id_marker = self.peer_id_marker.clone();
         let peer_metadata = self.peer_metadata.clone();
         let peer_compatible_network_version = self.peer_compatible_network_version.clone();
+        let throttle_quota = self.throttle_quota.clone();
 
         self.tokio_executor.spawn(async move {
             // prepare PeerId
@@ -298,7 +342,7 @@ impl Actor for Peer {
             }, None);
 
             // begin to process incoming messages in a loop
-            begin_process_incoming(net, myself.clone(), network_channel, log).await;
+            begin_process_incoming(net, myself.clone(), network_channel, throttle_quota, log.clone()).await;
 
             // connection to peer was closed, stop this actor
             system.stop(myself);
@@ -315,6 +359,19 @@ impl Receive<SendMessage> for Peer {
     type Msg = PeerMsg;
 
     fn receive(&mut self, ctx: &Context<Self::Msg>, msg: SendMessage, _sender: Sender) {
+        match self.throttle_quota.lock() {
+            Ok(ref mut quota) => {
+                if !quota.can_send(msg.message.as_ref()) {
+                    return;
+                }
+            }
+            Err(e) => warn!(
+                ctx.system.log(),
+                "Failed to obtain a lock on throttling quota";
+                "reason" => format!("{:?}", e)
+            ),
+        }
+
         let system = ctx.system.clone();
         let myself = ctx.myself();
         let tx = self.net.tx.clone();
@@ -571,6 +628,7 @@ async fn begin_process_incoming(
     net: Network,
     myself: PeerRef,
     event_channel: NetworkChannelRef,
+    throttle_quota: Arc<std::sync::Mutex<ThrottleQuota>>,
     log: Logger,
 ) {
     info!(log, "Starting to accept messages");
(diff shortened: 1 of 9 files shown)


networking/src/p2p/peer.rs
```diff
@@ -298,7 +460,9 @@ impl Actor for Peer {
             }, None);
 
             // begin to process incoming messages in a loop
-            begin_process_incoming(net, myself.clone(), network_channel, log).await;
+            begin_process_incoming(net, myself.clone(), network_channel, throttle_quota, log.clone()).await;
+
+            notify.notify_one();
```
bkontur, line 465: @akoptelov 
I am not sure if this `notify.notify_one();` is trigger all the time when peer craches, thread fails...
maybe, I would add also another `notify.notify_one();`  to the `post_stop` just to be sure

akoptelov: I've just moved it there.

networking/src/p2p/peer.rs
```diff
@@ -315,6 +529,19 @@ impl Receive<SendMessage> for Peer {
     type Msg = PeerMsg;
 
     fn receive(&mut self, ctx: &Context<Self::Msg>, msg: SendMessage, _sender: Sender) {
+        match self.throttle_quota.lock() {
+            Ok(ref mut quota) => {
+                if !quota.can_send(msg.message.as_ref()) {
+                    warn!(ctx.system.log(), "Cannot send message because of send quota is exceeded"; "msg" => format!("{:?}", msg.message.as_ref()), "peer_uri" => ctx.myself().uri().to_string());
```
tizoc, line 535: Is it possible for this to flood the logs?

networking/src/p2p/peer.rs
```diff
@@ -582,23 +810,30 @@ async fn begin_process_incoming(
...
+                                            message: Arc::new(msg),
+                                        }
+                                        .into(),
+                                        topic: NetworkChannelTopic::NetworkEvents.into(),
+                                    },
+                                    None,
+                                );
+                            }
+                        } else {
+                            warn!(log, "Message is dropped because its receive quota is exceeded"; "msg" => format!("{:?}", msg), "peer_uri" => myself.uri().to_string());
```
tizoc, line 832: Same question as above (and same answer I guess). Will this flood the logs if we keep getting a lot of messages and we are over our quota?

networking/src/p2p/peer.rs
```diff
@@ -636,3 +871,150 @@ async fn begin_process_incoming(
...
+        );
+        let msg = create_test_mgs();
+        let idx = ThrottleQuota::msg_index(&msg);
+        for _ in 0..super::THROTTLING_QUOTA_MAX[idx].0 + 11 {
+            peer.tell(SendMessage::new(Arc::new(msg.clone())), None);
+        }
+        std::thread::sleep(Duration::from_millis(50));
+
+        // 11 warnings on dropped messages
+        assert_eq!(received_messages.load(Ordering::Relaxed), 11);
```
tizoc, line 1007: Ok, this is related to my comments above, I see that because of how it is tested the warning needs to be logged for every message.

I don't know what is the amount of values expected, but if it is thousands, do we want thousands warnings in the logs?

tizoc:
I left a few (probably not terribly useful) comments. I am not familiar with the underlying code, so I am not really equipped to properly review this.

tizoc:
Btw, there are some failures here (and I see warnings related to the quotas in the logs): http://ci.tezedge.com/tezedge/tezedge/1279/5/8

tizoc:
@akoptelov please check on the #security slack tomorrow the discussion I had with @dkuehr, he just happened to hit the logger flooding issue I asked about when testing his fuzzer.

networking/src/p2p/peer.rs
akoptelov: Thanks for noticing that. Indeed that might be an issue. I'll log the first occurrence and then the total number when doing quota reset.

akoptelov:
> I left a few (probably not terribly useful) comments. I am not familiar with the underlying code, so I am not really equipped to properly review this.

This is actually very nice catch. Now it should be much quieter.

akoptelov:
> Btw, there are some failures here (and I see warnings related to the quotas in the logs): http://ci.tezedge.com/tezedge/tezedge/1279/5/8

Yeah, this is `BlockHeader`, and we discussed that we should increase the limit for it. Otherwise we might do that individually for the test.

networking/src/p2p/peer.rs
```diff
@@ -812,7 +819,12 @@ async fn begin_process_incoming(
             Ok(res) => match res {
                 Ok(msg) => match throttle_quota.lock() {
                     Ok(ref mut quota) => {
-                        if quota.can_receive(&msg) {
+                        if let Err(quota) = quota.can_receive(&msg) {
+                            if quota == -1 {
+                                warn!(log, "Message is dropped because its receive quota is exceeded";
```
tizoc, line 824: much better👍

akoptelov:
@bkontur Let's merge it.

networking/src/p2p/peer.rs
```diff
@@ -206,6 +208,192 @@ struct Network {
...
+
+    /// Quota for tx/rx messages per [THROTTLING_QUOTA_RESET_MS]
+    static ref THROTTLING_QUOTA_MAX: [(isize, isize); THROTTLING_QUOTA_NUM] = {
+        let mut default = [
+            (1, 1), // Disconnect
+            (1, 1), // Advertise
+            (10, 10), // SwapRequest
+            (10, 10), // SwapAck
+            (1, 1), // Bootstrap
+            (10, 500), // GetCurrentBranch
```
binier, line 260: Here we accept 500 `GetCurrentBranch` messages. But below for `CurrentBranch` we have a quota of `200`. Maybe they should be equal? Here we will accept 500 get requests, but only respond to 200. Same for other kinds of messages.

akoptelov: According to my measurements, there is an asymmetry in messages rate in Tezos network (https://viable-systems.slack.com/archives/C01SGFZ5L5P/p1626424648216800).

networking/src/p2p/peer.rs
```diff
@@ -315,6 +541,22 @@ impl Receive<SendMessage> for Peer {
     type Msg = PeerMsg;
 
     fn receive(&mut self, ctx: &Context<Self::Msg>, msg: SendMessage, _sender: Sender) {
+        match self.throttle_quota.lock() {
+            Ok(ref mut quota) => {
+                if let Err(quota) = quota.can_send(msg.message.as_ref()) {
+                    if quota == -1 {
+                        warn!(ctx.system.log(), "Cannot send message because of send quota is exceeded";
+                              "msg" => format!("{:?}", msg.message.as_ref()), "peer_uri" => ctx.myself().uri().to_string());
```
bkontur, line 549: @akoptelov 
do we really need to log here message bytes?

networking/src/p2p/peer.rs
```diff
@@ -315,6 +541,22 @@ impl Receive<SendMessage> for Peer {
...
+                    }
+                    return;
+                }
+            }
+            Err(_) => error!(
```
bkontur, line 554: @akoptelov 
do not hide error please, 
```
            Err(e) => error!(
                ctx.system.log(),
                "Failed to obtain a lock on throttling quota";
                "reason" => format!("{:?}", e),
            ),
```

networking/src/p2p/peer.rs
```diff
@@ -582,23 +825,33 @@ async fn begin_process_incoming(
...
+                                        .into(),
+                                        topic: NetworkChannelTopic::NetworkEvents.into(),
+                                    },
+                                    None,
+                                );
+                            }
+                        }
                     }
-                }
+                    Err(_) => error!(log, "Failed to obtain a lock for throttle quota"),
```
bkontur, line 853: @akoptelov 
also please, log error at least

networking/src/p2p/peer.rs
bkontur: @akoptelov 
also change error to warn please

networking/src/p2p/peer.rs
```diff
@@ -206,6 +208,192 @@ struct Network {
...
+                        default[i] = (tx, rx);
+                    }
+                }
+            }
+        }
+        default
+    };
+}
+
+struct ThrottleQuota {
```
bkontur, line 289: @akoptelov 
I would move this all throttling stuff, to separate file:
`quota.rs` or `throttle_quota.rs`

networking/src/p2p/peer.rs
```diff
@@ -582,23 +825,33 @@ async fn begin_process_incoming(
...
-                                }
-                                .into(),
-                                topic: NetworkChannelTopic::NetworkEvents.into(),
-                            },
-                            None,
-                        );
+                Ok(msg) => match throttle_quota.lock() {
+                    Ok(ref mut quota) => {
+                        if let Err(quota) = quota.can_receive(&msg) {
+                            if quota == -1 {
```
bkontur, line 831: @akoptelov 
I dont like this second if `if quota == -1 {` , it is not that clear, 
if there is a problem with quota, then the first if `if let Err(quota)` should be enought and I would just log exceeeded quota number to log

networking/src/p2p/peer.rs
```diff
@@ -315,6 +541,22 @@ impl Receive<SendMessage> for Peer {
     type Msg = PeerMsg;
 
     fn receive(&mut self, ctx: &Context<Self::Msg>, msg: SendMessage, _sender: Sender) {
+        match self.throttle_quota.lock() {
+            Ok(ref mut quota) => {
+                if let Err(quota) = quota.can_send(msg.message.as_ref()) {
+                    if quota == -1 {
```
bkontur, line 547: @akoptelov 
also here, just one IF `if let Err(quota) =`  should be enought

networking/src/p2p/peer.rs
```diff
@@ -266,9 +469,31 @@ impl Actor for Peer {
...
+        let throttle_quota = self.throttle_quota.clone();
+        let log = ctx.system.log().clone();
+        let stop = self.quota_update_stop.clone();
+        self.tokio_executor.spawn(async move {
+            loop {
+                tokio::select! {
+                    _ = tokio::time::sleep(Duration::from_millis(*THROTTLING_QUOTA_RESET_MS)) => {
+                        match throttle_quota.lock() {
+                            Ok(ref mut quota) => quota.reset_all(),
+                            Err(_) => error!(log, "Failed to obtain a lock on throttling quota"),
```
bkontur, line 487: @akoptelov 
supressed erorr, please add to log, and change everywhere error! to warn! please

networking/src/p2p/peer.rs
```diff
@@ -206,6 +208,192 @@ struct Network {
...
+
+    fn index_to_str(index: usize) -> &'static str {
+        if index < THROTTLING_QUOTA_NUM {
+            THROTTLING_QUOTA_STRS[index]
+        } else {
+            "<invalid index>"
+        }
+    }
+
+    fn can_send(&mut self, msg: &PeerMessageResponse) -> Result<isize, isize> {
```
bkontur, line 336: @akoptelov 
I would simplify this a little bit:
`Result<isize, isize>`  -> `Result<(), isize>`
I dont see anywhere to handle `Ok(quota)`

also the same for `can_receive`

bkontur (changes requested):
@akoptelov 
please check PR comments

networking/src/p2p/peer.rs
bkontur: @akoptelov 
ok, so I would do it better reverse:
- the whole IF with log `if quota == -1 {`  just move inside to the `quota.can_receive`
- so here we just need simple if `if quota.can_receive(&msg).is_ok() {`
  -  or maybe easier, just return `bool` on `quota.can_receive`

networking/src/p2p/peer.rs
akoptelov: See above.

networking/src/p2p/peer.rs
bkontur: @akoptelov 
and what about that `error!` -> `warn!` ?

networking/src/p2p/peer/quota.rs
```diff
@@ -0,0 +1,202 @@
...
+
+    pub fn can_receive(&mut self, msg: &PeerMessageResponse) -> Result<(), isize> {
+        let index = Self::msg_index(msg);
+        if THROTTLING_QUOTA_MAX[index].1 <= 0 {
+            return Ok(());
+        }
+        self.quotas[index].1 = self.quotas[index].0.checked_sub(1).ok_or(self.quotas[index].1)?;
+        if self.quotas[index].1 >= 0 {
+            self.quotas[index].0 =
+                cmp::min(self.quotas[index].0 + 1, THROTTLING_QUOTA_MAX[index].0);
```
bkontur, line 168: @akoptelov 
what about this `self.quotas[index].0 + 1`  if anybody will set quota to i32::MAX it will trigger overflow in debug mode I guess, better use checked_add

and also later `"amount" => -self.quotas[index].0 + tx_max`

networking/src/p2p/peer/quota.rs
```diff
@@ -0,0 +1,199 @@
...
+                      "msg_kind" => Self::index_to_str(index));
+            }
+            false
+        }
+    }
+
+    pub fn reset_all(&mut self) {
+        for index in 0..THROTTLING_QUOTA_NUM {
+            let (tx_max, rx_max) = THROTTLING_QUOTA_MAX[index];
+            if self.quotas[index].0 < 0 {
```
bkontur, line 177: @akoptelov 
you access here 4 times `self.quotas[index]`,
wouldnt it be faster to access it just once or something like that?
```
let mut tuple = self.quotas[index];
```

networking/src/p2p/peer/quota.rs
```diff
@@ -0,0 +1,199 @@
...
+        let index = Self::msg_index(msg);
+        if THROTTLING_QUOTA_MAX[index].1 <= 0 {
+            return true;
+        }
+        decrease(&mut self.quotas[index].1);
+        if self.quota_disabled || self.quotas[index].1 >= 0 {
+            true
+        } else {
+            if self.quotas[index].1 == -1 {
+                warn!(self.log, "Cannot receive message because its receive quota is exceeded";
```
bkontur, line 167: @akoptelov 
I would also add it here peer_uri, that we know and analyze more

akoptelov: We have that already, it is in the parent log:
```
Jul 19 16:54:25.219 WARN Cannot send message because its send quota is exceeded, msg_kind: Advertise, peer_uri: test-peer
Jul 19 16:54:30.170 WARN Tx quota is exceeded, amount: 11, quota: 1, msg_kind: Advertise, peer_uri: test-peer
```

bkontur (approved):
@akoptelov @jurajselep 
please, rebase/squash/merge to develop,
and please dont forget for that test to p2p_test.rs

networking/src/p2p/peer.rs
akoptelov: @binier Anyway, that symmetry does make sense, so updated that.

PR #653, opened 2021-07-05
Before performing an RPC service lookup, normalize the URL path

### LINKED ISSUE #650: Double slash (``/'') should be equivalent to single slash in RPCs
For example:
```
:; curl -s http://127.0.0.1:8732/chains/main/blocks//head/header
Compat { error: ErrorMessage { msg: "Invalid block_id_param: , reason: missing checksum" } }
:;  curl -s http://127.0.0.1:8732/chains/main/blocks/head/header
{"hash":"BLockGenesisGenesisGenesisGenesisGenesisf79b5d1CoW2","chain_id":"NetXdQprcVkpaWU","level":0,"proto":0,"predecessor":"BLockGenesisGenesisGenesisGenesisGenesisf79b5d1CoW2","timestamp":"2018-06-30T16:07:32Z","validation_pass":0,"operations_hash":"LLoZS2LW3rEi7KYU4ouBQtorua37aWWCtpDmv1n2x3xoKi6sVXLWp","fitness":[],"context":"CoVNUgnfVocNYM2HNkHxMMcZdzqFDkCQ5vEGa3gsdpkidoA1QPZo","protocol":"PrihK96nBAFSxVL1GLJTVhu9YnzkMFiBeuJRPA8NwuZVZCE1L6i"}
```
The former command should likely return the same as the latter command. This would match the behaviour of Nomadic Labs `tezos-node` which would help make sure that scripts that work against NL `tezos-node` would work without update with `tezedge`.

tizoc:
Fixes #650

### DIFF
--- a/rpc/src/server/mod.rs
+++ b/rpc/src/server/mod.rs
@@ -25,6 +25,7 @@ use tezos_api::environment::TezosEnvironmentConfiguration;
 use tezos_messages::p2p::encoding::version::NetworkVersion;
 use tezos_wrapper::TezedgeContextClient;
 use tezos_wrapper::TezosApiConnectionPool;
+use url::Url;
 
 use crate::rpc_actor::RpcCollectedStateRef;
 use crate::{error_with_message, not_found, options};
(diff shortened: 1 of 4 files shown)


rpc/src/server/mod.rs
```diff
@@ -210,6 +214,19 @@ pub fn spawn_server(
         }))
 }
 
+/// Normalizes the request path
+fn normalize_path(original_path: &str) -> Option<String> {
```
bkontur, line 218: @tizoc 
Could you, please, add here some easy unit-test with some corner cases?
Because from description "Normalizes the request path" it is not very obvious, what it is doing :)

bkontur:
@tizoc 
I just forced push rpc-cache-tweaks here, so closing this

rpc/src/server/mod.rs
tizoc: @bkontur yes, but first I have to understand exactly what hyper and `PathTree` do (and don't do) with the path to understand what needs to be done here (and what doesn't). The current fix is a quick hack to fix multiple `/`s in the path. It is also important that the normalization behaves the same as what Octez does.

tizoc: Or to put it another way, the actual solution is probably much more complicated than this and requires some research, this one here is not the final solution.

PR #657, opened 2021-07-06
Add docker compose files for florencenet and mainnet with both contex…

tizoc:
…ts enabled.


bkontur:
ok, thx, commit cherry-picke to `origin/dockers`

PR #658, opened 2021-07-07
Rewrite P2p to single threaded deterministic state machine

binier:
Replaces riker(actor system) implementation for peer management and handshaking, with deterministic state machine.

Here are some docs useful before reviewing: https://github.com/binier/tezedge/blob/p2p-simulator/tezedge_state/README.md


tezedge_state/Cargo.toml
```diff
@@ -0,0 +1,34 @@
...
+
+[dependencies]
+getset = "0.1"
+bytes = "1"
+slab = "0.4.3"
+rand = "0.7.3"
+failure = "0.1"
+futures = "0.3"
+hex = "0.4"
+riker = "0.4"
```
tizoc, line 17: is riker used for anything in this implementation?

binier: No its not. I need to clear dependencies in other places too.

binier: removed unused dependencies from tezedge-state.

shell/src/chain_manager.rs
```diff
@@ -324,36 +327,38 @@ impl ChainManager {
...
+                    let peer_id = peer_state.peer_id.clone();
                     // clear innner state (not needed, it will be drop)
                     peer_state.clear();
+                    if let Some(peer_branch_bootstrapper) = self.chain_state.peer_branch_bootstrapper()
+                    {
+                        peer_branch_bootstrapper.tell(CleanPeerData(peer_id), None);
+                    }
                 }
                 // tell bootstrapper to clean potential data
-                if let Some(peer_branch_bootstrapper) = self.chain_state.peer_branch_bootstrapper()
```
bkontur, line 347: @binier 
I would better leave this as it was, outside that if,
for example, if `self.peers.remove` was called before in `impl Receive<SystemEvent>` 
then you possibly dont clean peer_branch_bootstrapper

shell/src/state/peer_state.rs
```diff
@@ -332,6 +343,13 @@ impl UpdateIsBootstrapped for PeerState {
     }
 }
 
-pub fn tell_peer(msg: Arc<PeerMessageResponse>, peer: &PeerState) {
-    peer.peer_id.peer_ref.tell(SendMessage::new(msg), None);
+pub fn tell_peer(
+    network_channel: NetworkChannelRef,
```
bkontur, line 347: @binier 
I guess, it is enought: `network_channel: &NetworkChannelRef` then you dont need to `network_channel.clone()` all the time

shell/src/chain_manager.rs
binier: I removed that because bootstrapper shouldn't be deciding whom to disconnect. Could cause lots of issues like what if state machine didnt receive event that the peer was disconnected, we might end up bootstrapper having 0 peers and state machine having maxed out connections. Then the bootstrap process would be stuck until someone disconnects and new peer gets connected.

Instead state machine is notified about a stalled peer and it disconnects him and issues an event that the peer was disconnected. Only then we can remove that peer from bootstrapper's state.

shell/src/state/peer_state.rs
binier: Fixed.

binier:
> @binier
> where is that patch?

RocksDB source code:
https://github.com/tezedge/rocksdb/commit/94753e5675989206aa558310ff0609a617f528a4

RocksDB Rust bindings:
https://github.com/tezedge/rust-rocksdb/tree/tezedge

shell/src/chain_manager.rs
```diff
@@ -226,7 +226,7 @@ pub struct ChainManager {
     identity_peer_id: CryptoboxPublicKeyHash,
 
     /// Holds the state of all peers
-    peers: HashMap<ActorUri, PeerState>,
+    peers: HashMap<PeerAddress, PeerState>,
```
bkontur, line 229: @binier 
I am not sure about this, if SocketAddress is enought, I could still have more peers with the same identity?

shell/src/chain_manager.rs
```diff
@@ -1713,7 +1706,7 @@ impl Receive<DisconnectStalledPeers> for ChainManager {
                         Ok(result) => result,
                         Err(_) => {
                             warn!(ctx.system.log(), "Failed to collect current local head";
-                                            "peer_id" => state.peer_id.peer_id_marker.clone(), "peer_ip" => state.peer_id.peer_address.to_string(), "peer" => state.peer_id.peer_ref.name(), "peer_uri" => uri.to_string());
+                                            "peer_ip" => state.peer_id.address.to_string());
```
bkontur, line 1709: @binier 
I tried to unify logs for peers a little  but address is not enought here, 
peer_id_marker there was a CryptoboxPublicKeyHash.to_base58_check, so I would like to have at least that "idX123456.." with socket_address in logs for later investigations, we needed/used that already several times

shell/src/chain_manager.rs
binier: `TezedgeState` makes sure to not accept any peers with same identity:

https://github.com/binier/tezedge/blob/534a29e8c2742855e49b0da61621c98146b82011/tezedge_state/src/acceptors/peer_handshake_message_acceptor.rs#L58-L74

shell/src/chain_manager.rs
binier: Hm... You are right, especially in this case we won't be able to identify same peer across multiple connection/disconnections.

tezedge_state/src/proposer/mio_manager.rs
```diff
@@ -0,0 +1,279 @@
...
+
+    address_to_token: HashMap<PeerAddress, usize>,
+    peers: Slab<NetPeer>,
+}
+
+impl MioManager {
+    const DEFAULT_BACKLOG_SIZE: u32 = 255;
+
+    pub fn new(listen_addr: SocketAddr) -> Self {
+        let poll = mio::Poll::new().expect("failed to create mio::Poll");
```
bkontur, line 112: @binier 
I guess, we should not use any expect/unwrap here

tezedge_state/src/proposer/mio_manager.rs
```diff
@@ -0,0 +1,279 @@
...
+        let waker = Arc::new(
+            mio::Waker::new(poll.registry(), MIO_WAKE_TOKEN).expect("failed to create mio::Waker"),
```
bkontur, line 114: @binier 
I guess, we should not use any expect/unwrap here

binier: We probably should. I dont think we can handle those errors anyhow.

shell/src/tezedge_state_manager.rs
```diff
@@ -0,0 +1,446 @@
...
+// SPDX-License-Identifier: MIT
+
+//! Module takes care about (initialization, thread starting, ...) one singletons [TezedgeState] and [TezedgeProposer]
+
+use std::collections::HashSet;
+use std::iter::FromIterator;
+use std::net::{IpAddr, SocketAddr};
+use std::sync::mpsc::Receiver;
+use std::sync::{mpsc, Arc};
+use std::time::{Duration, Instant, SystemTime};
```
bkontur, line 11: @binier @jurajselep 
do we need SystemTime here?
I remember Juraj, wanted to remove all usage of SystemTime, but I dont remember exactly why

binier: `SystemTime` is used only for initial time. Then it is updated using monotonic clock.

If you'll check proposals, now they include `Duration` instead of `Instant`, using that that system time gets updated.

I know why we cant rely on systemtime. In fact it would break my state machine easily because of how not reliable it is...

But I only call `SystemTime::now()` only once, and update it with durations, which are measured by comparing current time in proposer and event time.

shell/src/chain_manager.rs
bkontur: @binier 
we need to be able identify peer by address and identity from logs,
if dont have time to do this now, we will do it later: https://viablesystems.atlassian.net/browse/TE-683

simulator/tests/simulate_disconnects.rs
```diff
@@ -0,0 +1,305 @@
+// use std::time::{Duration, Instant};
```
bkontur, line 1: @binier 
do we need this or why is it commented out?

simulator/tests/simulate_many_incoming_connections.rs
```diff
@@ -0,0 +1,65 @@
+use std::time::{Duration, Instant, SystemTime};
```
bkontur, line 1: @binier 
lots of files miss:
```
// Copyright (c) SimpleStaking, Viable Systems and Tezedge Contributors

```

simulator/tests/simulate_disconnects.rs
binier: That test needs fixing, it's partially adjusted to new simulator, but still needs some work.

storage/src/lib.rs
```diff
@@ -12,6 +12,7 @@ use std::{
 };
 
 use failure::Fail;
+pub use rocksdb;
```
bkontur, line 15: @binier 
where is this need?
I dont see any usage of rocksdb on tezedge_state or other new module

tezedge_state/Cargo.toml
```diff
@@ -0,0 +1,37 @@
+[package]
+name = "tezedge-state"
+version = "0.1.0"
```
bkontur, line 3: @binier 
this will be 1.7.0,
but I have fixed it in consequent PR:
https://github.com/tezedge/tezedge/pull/752/files

simulator/tests/simulate_many_incoming_connections.rs
binier: done,

storage/src/lib.rs
binier: I think @vlad9486 needed it for debugger, to use same version.

bkontur: @vlad9486 
Vlad, is it needed for debugger?

tezedge_state/src/chunking/chunk_read_buffer.rs
```diff
@@ -0,0 +1,140 @@
...
+
+        Some(chunk)
+    }
+
+    /// Consume buffer and replace it with new empty one.
+    pub fn take_if_ready(&mut self) -> Option<BinaryChunk> {
+        if !self.is_finished() {
+            return None;
+        }
+        // TODO: use reference instead
```
bkontur, line 128: @binier 
what about this?
is it needed or please add here some continue jira not to forget it

tezedge_state/src/chunking/message_read_buffer.rs
```diff
@@ -0,0 +1,122 @@
...
+
+    /// Returns if more might be available to be read.
+    fn _read_from<R: Read>(
+        &mut self,
+        reader: &mut R,
+        crypto: &mut PeerCrypto,
+    ) -> Result<PeerMessageResponse, ReadMessageError> {
+        loop {
+            self.chunk_reader.read_from(reader)?;
+            // TODO: stop reading if message_len < message_buf.len() + chunk_expected_len
```
bkontur, line 51: @binier 
still relevant comment?

tezedge_state/src/effects/effects_recorder.rs
```diff
@@ -0,0 +1,146 @@
...
+    pub fn new() -> Self {
+        Default::default()
+    }
+}
+
+impl Effects for RecordedEffects {
+    fn get_nonce(&mut self, _: &PeerAddress) -> Nonce {
+        self.nonces
+            .pop_front()
+            .expect("no more recorded nonces avaiable")
```
bkontur, line 31: @binier 
we cannot use expect in state machine, this is unrecoverable

tezedge_state/src/effects/effects_recorder.rs
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    fn choose_peers_to_connect_to(
+        &mut self,
+        _: &BTreeSet<PeerListenerAddress>,
+        _: usize,
+    ) -> Vec<PeerListenerAddress> {
+        self.chosen_peers_to_connect_to
+            .pop_front()
+            .expect("no more recorded chosen peers to connect to avaiable")
```
bkontur, line 41: @binier 
we cannot use expect in state machine, this is unrecoverable

tezedge_state/src/effects/effects_recorder.rs
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    fn choose_potential_peers_for_advertise(
+        &mut self,
+        _: &BTreeSet<PeerListenerAddress>,
+    ) -> Vec<PeerListenerAddress> {
+        self.chosen_potential_peers_for_advertise
+            .pop_front()
+            .expect("no more recorded chosen potential peers for advertise avaiable")
```
bkontur, line 50: @binier 
we cannot use expect in state machine, this is unrecoverable

tezedge_state/src/effects/effects_recorder.rs
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    fn choose_potential_peers_for_nack(
+        &mut self,
+        _: &BTreeSet<PeerListenerAddress>,
+    ) -> Vec<PeerListenerAddress> {
+        self.chosen_potential_peers_for_nack
+            .pop_front()
+            .expect("no more recorded chosen potential peers for nack avaiable")
```
bkontur, line 59: @binier 
we cannot use expect in state machine, this is unrecoverable

tezedge_state/src/lib.rs
```diff
@@ -0,0 +1,31 @@
...
+
+mod tezedge_state_wrapper;
+pub use tezedge_state_wrapper::*;
+
+pub mod proposer;
+
+pub mod sample_tezedge_state;
+
+#[derive(Debug, Eq, PartialEq)]
+pub enum InvalidProposalError {}
```
bkontur, line 31: @binier 
wouldnt it be better to place it to `src/proposals/mod.rs` ?

tezedge_state/src/peer_address.rs
```diff
@@ -0,0 +1,230 @@
...
+
+impl DefaultRecorder for PeerAddress {
+    type Recorder = CloneRecorder<PeerAddress>;
+
+    fn default_recorder(self) -> Self::Recorder {
+        CloneRecorder::new(self)
+    }
+}
+
+// impl Hash for PeerAddress {
```
bkontur, line 89: @binier 
do we still need this commented code?

tezedge_state/src/peer_address.rs
```diff
@@ -0,0 +1,230 @@
...
+    /// Caller must make sure that the set port is the port that the
+    /// remote peer is listening on.
+    pub fn as_listener_address(&self) -> PeerListenerAddress {
+        PeerListenerAddress(self.0)
+    }
+
+    /// For testing purposes.
+    ///
+    /// Get unique ipv4 address based on index.
+    pub fn ipv4_from_index(index: u64) -> Self {
```
bkontur, line 48: @binier 
I am not very big fan of mixing test stuff to "pub" stuff

tezedge_state/src/chunking/chunk_read_buffer.rs
binier: Yeah that comment no longer applies. `take_ref_if_ready` method exists which does what that comment says.

tezedge_state/src/chunking/message_read_buffer.rs
binier: That's an optimization maybe good to have (unless peer's implementation is broken, that optimization should never apply), for now I don't want to break anything so it can be left as is.

tezedge_state/src/effects/effects_recorder.rs
binier: That is while replaying recording. We should panic if state machine requested nonce more times at current run than while recording. That means non-determinism and we should panic to detect it.

tezedge_state/src/effects/effects_recorder.rs
binier: same as for `get_nonce()`.

tezedge_state/src/effects/effects_recorder.rs
binier: same as for `get_nonce()`.

tezedge_state/src/effects/effects_recorder.rs
binier: same as for `get_nonce()`.

tezedge_state/src/lib.rs
binier: Yeah maybe, but I might just remove that as you can see, no error can occur just based on proposal.

Before I was validating time passed in proposal and erroring if proposal was outdated, but now each proposal contains `time_passed` instead, so we don't need that check anymore (and it is removed).

tezedge_state/src/peer_address.rs
binier: I guess no longer. I'll remove it.

tezedge_state/src/proposer/mod.rs
```diff
@@ -0,0 +1,948 @@
...
+        if event.is_server_event() {
+            // we received event for the server (client opened tcp stream to us).
+            loop {
+                // as an optimization, execute requests only after 100
+                // accepted new connections. We need to execute those
+                // requests as they might include command to stop
+                // listening for new connections or disconnect new peer,
+                // if for example they are blacklisted.
+                for _ in 0..100 {
+                    match self.manager.accept_connection(&event) {
```
vlad9486, line 303: It will return after each accepted connection, and newer reach 100.

binier: Why? If there are 100+ peers in the backlog, we will reach 100, and if not and if we return, that's not a problem. That code is just an optimization.

binier: It's documented above those lines.

tezedge_state/src/chunking/chunk_read_buffer.rs
bkontur: @binier 
ok, cool, so, please, just remove that TODO

tezedge_state/src/chunking/message_read_buffer.rs
bkontur: @binier 
ok, cool, so please, just update that TODO with this comment, you just wrote here,
that we now what to do with that later

tezedge_state/src/chunking/chunk_read_buffer.rs
binier: Yeah I will, I have already in local, but I'll group those changes in 1 commit at the end of review :)

tezedge_state/src/peer_address.rs
binier: I agree, we should have that under feature flag.

tezedge_state/src/effects/effects_recorder.rs
bkontur: @binier 
well, I am just afraid if anybody can still use it accidentally for something other for running node,
I know it will mess a nice trait Effects api without Result :),
at least put there, please, some comments at the begining
```
//! Never use, dedicated just for replay...
```

maybe, it should be aslo some feature/cfg (never used it before)

tezedge_state/src/lib.rs
bkontur: @binier 
I see, so even better if we can remove that :)

tezedge_state/src/proposer/mod.rs
```diff
@@ -0,0 +1,948 @@
...
+    Efs: Effects + Debug,
+    M: Manager<Stream = S, NetworkEvent = NetE, Events = Es>,
+{
+    #[inline]
+    fn wait_for_events(&mut self, events: &mut Es) {
+        self.manager
+            .wait_for_events(events, self.config.wait_for_events_timeout)
+    }
+
+    // fn handle_event(&mut self, event: Event<NetE>) {
```
bkontur, line 256: @binier 
do we need this commented code?

tezedge_state/src/proposer/mod.rs
```diff
@@ -0,0 +1,948 @@
...
+    }
+
+    fn handle_readiness_event(&mut self, event: &NetE) {
+        self.time.update(event.time());
+        let peer = match self.manager.get_peer_for_event_mut(&event) {
+            Some(peer) => peer,
+            None => {
+                // Should be impossible! If we receive an event for
+                // the peer, that peer must exist in manager.
+                // TODO: write error log.
```
bkontur, line 333: @binier 
please, better do that log (error/warn) now, we could see if that really happens and how often

tezedge_state/src/proposer/mod.rs
bkontur: @binier @vlad9486 
I was also staring at the `0..100` I added note to TE-652 for that `check for _ in 0..100 {`

I dont think it is a good idea to have hard-coded constants like this, it should be a part of configuration at least.

is it releated to low/high/expectected_peers count?

tezedge_state/src/proposer/proposal_loader.rs
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    pub fn next(&mut self) -> Option<Result<RecordedProposal, serde_json::Error>> {
+        self.reader.next()
+    }
+}
+
+impl Clone for ProposalLoader {
+    fn clone(&self) -> Self {
+        // TODO
```
bkontur, line 30: @binier 
do we still need that?

tezedge_state/src/proposer/mio_manager.rs
bkontur: @binier 
I got your point, it is done just on startup, it start or it fails,
but for example if open database before that or run other stuff, we loose kind of possiblity for gracefull shutdown for example,

and all the time we are trying to remove panic/expect/unwraps (just for sure) from everywhere (ok, except the tests)

tezedge_state/src/proposer/mod.rs
binier: > is it releated to low/high/expectected_peers count?

No it's not. Basically the problem is that `Acceptor` doesn't have a return type, so we don't instantly know that state machine refused that connection. Also there is no **refuse** incoming connection in tcp, instead we have to accept then drop the connection. So 100 in there is the bound how many peers we will accept at a time in bulk, before `execute_requests` so that we get request to disconnect rejected peers.

That solution works, but obviously it's far from pretty or optimal. I guess we could add associated type to the `Acceptor` for return type, not sure if that will work out well.

tezedge_state/src/proposer/mod.rs
binier: It's not used at the moment and just was tired of keeping that impl in sync with `handle_event_ref`. Basically it was for more flexibility, so that `NetworkEvent` can be owned during testing, instead of being a ref (like it is with mio). Maybe it's better to keep it there so that if we want to handle owned events for some test, we can uncomment that.

tezedge_state/src/proposer/proposal_loader.rs
binier: Yes. It's kinda a dirty trick to make it work for testing, because we need cloning in testing and there we don't need record/replay functionality, so that clone won't ever be called.

In real run, whenever we use `mio` types from that generic argument is not clonable, so we can't ever clone `TezedgeProposer` and hit that code.

Basically once I make recorder/replayer generic types on proposer, those can be removed.

tezedge_state/src/proposer/proposal_persister.rs
```diff
@@ -0,0 +1,80 @@
...
+pub struct ProposalPersisterHandle {
+    sender: Option<mpsc::SyncSender<RecordedProposal>>,
+    join_handle: Option<std::thread::JoinHandle<()>>,
+}
+
+impl Drop for ProposalPersisterHandle {
+    fn drop(&mut self) {
+        drop(self.sender.take());
+        if let Some(jh) = self.join_handle.take() {
+            jh.join().unwrap();
```
bkontur, line 20: @binier 
I am not sure, but I readsomewhere, that causing panic in drop, could lead to unexpected behavior,
and there could be possibly issue, that other Drops are not called or something like that,
I would rather do `match h.join()` and log or just do it without `unwrap`

tezedge_state/src/proposer/proposal_persister.rs
```diff
@@ -0,0 +1,80 @@
...
+    pub fn persist<P>(&mut self, proposal: P)
+    where
+        P: Into<RecordedProposal>,
+    {
+        // TODO: better error handling.
+        self.sender
+            .as_mut()
+            .unwrap()
+            .send(proposal.into())
+            .expect("recorded proposal persister disconnected!");
```
bkontur, line 42: @binier 
I guess, persister should be used when node is running, so here cannot be `expect`,
this TODO needs to be fixed

tezedge_state/src/proposer/proposal_persister.rs
```diff
@@ -0,0 +1,80 @@
...
+    channel: mpsc::Receiver<RecordedProposal>,
+}
+
+impl ProposalPersister {
+    pub fn start() -> ProposalPersisterHandle {
+        let (tx, rx) = mpsc::sync_channel(1024);
+
+        let state = Self { channel: rx };
+
+        let join_handle = std::thread::spawn(move || run(state));
```
bkontur, line 56: @binier 
please spawn this with thread name like we do everywhere for std threads,
also this makes you propage error, `-> Result<ProposalPersisterHandle, Error>` :)

e.g.:
```
let proposer_thread_handle = std::thread::Builder::new()
                .name("tezedge-proposer".to_owned())
                .spawn(move || {
                    proposer.extend_potential_peers(Instant::now(), initial_potential_peers);
                    run(
                        proposer,
                        proposer_rx,
                        network_event_notifier,
                        shell_command_notifier,
                    );
                })
                .map_err(TezedgeStateManagerSpawnError::IoError)?;
```

tezedge_state/src/proposer/proposal_persister.rs
binier: I guess it comes down to how we should handle errors from which we can't recover. Maybe we should just log error and move on without persister, but idk. If for some reason, if persister failed, that means that those proposals are gone.

tezedge_state/src/proposer/mod.rs
binier: It's not really an issue even if we receive such an event, nothing will break, though ok I can add log in there.

tezedge_state/src/lib.rs
binier: Am gonna keep it just in case we want to validate proposal somehow, but will move it to `src/proposals/mod.rs` as you suggested.

tezedge_state/src/peer_address.rs
binier: It's not enough, because `simulator` crate needs it, and it's not a test module, it's more of a framework.

tezedge_state/src/proposer/mod.rs
binier: Nvm, removed.

tezedge_state/src/proposer/proposal_persister.rs
bkontur: @binier @jurajselep 
I guess, this is easy, if storing of proposal (for replay) fails here, node cannot crash, and node has to still work:
- bakers should still bake
- indexers should still index data
- node should still apply new blocks
- node should still do a business

node can not crash on "any error", maybe some errors related to business, maybe, if automatic protocol upgrade fails or "disk is full", but I dont see any other relevant error,

if storing of proposal for replay fails, ok, we loose possibility for correct replay, but anyway you can log the whole proposal to log, and if somebody will send us `store-proposals-for-replay-log.data` + `slog.log`, we could still reproduce "correct" replay.

but I am almost sure, that node can not crash on storing proposal here, how would you explain that to baker, that he lost money, because we cannot store proposal? :)

so, back to the topic, so as I wrote in other comments, we better should not use `expect/unwrap/panic` anywhere in code (ok, in test ok), I know we use it on startup, which is also no good, but it will be a good strict rule not to use anywhere just in case

tezedge_state/src/proposer/proposal_persister.rs
```diff
@@ -0,0 +1,83 @@
...
+impl ProposalPersister {
+    pub fn start() -> ProposalPersisterHandle {
+        let (tx, rx) = mpsc::sync_channel(1024);
+
+        let state = Self { channel: rx };
+
+        let join_handle = std::thread::Builder::new()
+            .name("proposal-persister".to_owned())
+            .spawn(move || run(state))
+            .expect("failed to spawn proposal-persister thread");
```
bkontur, line 59: @binier 
please, change `expect` to propage error with Result instead,
I would leave that responsibility to caller,
caller should decide on error, if continue without storing proposals or if fail or propage error higher,

if this is called on starting of node, ok, then light/main.rs we can decide to crash the node immediatelly

tezedge_state/src/proposer/proposal_persister.rs
```diff
@@ -0,0 +1,83 @@
...
+        ProposalPersisterHandle {
+            sender: Some(tx),
+            join_handle: Some(join_handle),
+        }
+    }
+}
+
+fn run(state: ProposalPersister) {
+    let file = File::create("/tmp/tezedge/recorded_proposals.log")
+        .expect("couldn't open/create file for persisting proposals");
```
bkontur, line 70: @binier 
well, yes, another expect

tezedge_state/src/proposer/proposal_persister.rs
```diff
@@ -0,0 +1,83 @@
...
+
+        ProposalPersisterHandle {
+            sender: Some(tx),
+            join_handle: Some(join_handle),
+        }
+    }
+}
+
+fn run(state: ProposalPersister) {
+    let file = File::create("/tmp/tezedge/recorded_proposals.log")
```
bkontur, line 69: @binier 
- I guess, this file should be maybe better handled before starting a new thread
- and also `/tmp/tezedge/recorded_proposals.log` cannot be hard-coded here, but need to be as a configuration argument of node

tezedge_state/src/proposer/proposal_persister.rs
```diff
@@ -0,0 +1,83 @@
...
+        .expect("couldn't open/create file for persisting proposals");
+    let mut buf_writer = BufWriter::new(file);
+
+    loop {
+        let proposal = match state.channel.recv() {
+            Ok(proposal) => proposal,
+            Err(_) => break,
+        };
+
+        serde_json::to_writer(&mut buf_writer, &proposal).unwrap();
```
bkontur, line 79: @binier 
unwraps also

tezedge_state/src/proposer/mod.rs
bkontur: @binier 
yes, please, add here at least a log,
we should log any error from the begining and later can decide or fix if it occures too often

tezedge_state/src/state/assert_state.rs
```diff
@@ -0,0 +1,83 @@
...
+fn assert_peer_counts(state: &TezedgeState) {
+    assert!(state.potential_peers.len() <= state.config.max_potential_peers);
+    assert!(state.pending_peers.len() <= state.config.max_pending_peers);
+    assert!(state.connected_peers.len() <= state.config.max_connected_peers);
+}
+
+#[inline]
+fn assert_no_overlapping_peers(state: &TezedgeState) {
+    fn panic_overlap<A: Display>(addr: A, first: PeerContainer, second: PeerContainer) {
+        panic!(
```
bkontur, line 42: @binier @jurajselep 
I am really not sure about this panic and asserts,
do we really want to crash the node in production with them?

shouldnt it be more `debug_assert*` just in debug mode?

tezedge_state/src/state/connected_peers.rs
```diff
@@ -0,0 +1,230 @@
...
+use crate::state::pending_peers::HandshakeResult;
+use crate::state::ThrottleQuota;
+use crate::{PeerAddress, PeerCrypto, Port};
+use tezos_messages::p2p::encoding::peer::{PeerMessage, PeerMessageResponse};
+use tezos_messages::p2p::encoding::prelude::NetworkVersion;
+
+/// Peer who have undergone handshake.
+#[derive(Getters, CopyGetters, Debug, Clone)]
+pub struct ConnectedPeer {
+    // #[get = "pub"]
```
bkontur, line 24: @binier 
hmm, I would remove those commented annotations,
and internal state shouldnt be exposed as `pub`, if we dont want getters here, better use `pub(crate)`

tezedge_state/src/state/connected_peers.rs
```diff
@@ -0,0 +1,230 @@
...
+            peers,
+        }
+    }
+
+    #[inline]
+    pub fn len(&self) -> usize {
+        self.peers.len()
+    }
+
+    // fn find_index(&self, address: &PeerAddress) -> Option<usize> {
```
bkontur, line 132: @binier 
remove please, if not needed

tezedge_state/src/state/assert_state.rs
binier: Those aren't used in running node, just testing/fuzzing. Maybe that needs to be under cfg flag as well,

tezedge_state/src/proposer/proposal_persister.rs
binier: Yeah that is called just once when starting the node.

tezedge_state/src/state/connected_peers.rs
binier: That is an internal structure used only in `TezedgeState`. It's goal is to guard against such access. Anyways removed to comments.

tezedge_state/src/proposer/proposal_persister.rs
binier: done.

tezedge_state/src/proposer/mio_manager.rs
binier: True, but we have those expects everywhere in "light_node/main.rs" for example. We can create task for it, to remove all `unwraps` and `expects` everywhere and have graceful shutdown.

tezedge_state/src/proposer/proposal_persister.rs
binier: @bkontur 
Ok substituted it with logging, so if error happens we log the error and proposal with it. But those proposals will flood the logs and make everything slower and eventually logs will grow so large that we will be out of space.

Maybe we should log here only once and then stop recording proposals, as it's not as important as keeping the node going.

tezedge_state/src/proposer/proposal_loader.rs
binier: Removed.

tezedge_state/src/state/assert_state.rs
binier: Also `debug_assert` is not enough, because fuzzer compiles node in `release` mode as well as when doing simulations or other tests.

tezedge_state/src/proposer/proposal_loader/file_proposal_loader.rs
```diff
@@ -0,0 +1,33 @@
...
+pub struct FileProposalLoader {
+    reader: StreamDeserializer<'static, IoRead<BufReader<File>>, RecordedProposal>,
+}
+
+impl FileProposalLoader {
+    pub fn new<P>(file_name: P) -> Self
+    where
+        P: AsRef<Path>,
+    {
+        let file = File::open(file_name).expect("couldn't open file for loading proposals");
```
bkontur, line 21: @binier 
another expect

tezedge_state/tests/simulate_one_peer_all_message_sequences.rs
```diff
@@ -0,0 +1,376 @@
...
+    fn read(&mut self, _: &mut [u8]) -> io::Result<usize> {
+        Ok(0)
+    }
+}
+
+impl io::Write for FakeWritable {
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        match self {
+            Self::NoLimit => Ok(buf.len()),
+            // Self::Limit(limit) => Ok(buf.len().min(*limit)),
```
bkontur, line 37: @binier 
is this needed?

tezedge_state/src/state/requests.rs
```diff
@@ -0,0 +1,327 @@
...
+        state.accept(PendingRequestProposal {
+            effects: &mut effects,
+            time_passed: Duration::new(0, 0),
+            req_id,
+            message: PendingRequestMsg::StartListeningForNewPeersError {
+                error: std::io::ErrorKind::AddrInUse,
+            },
+        });
+
+        match state.requests.get(req_id).unwrap().status {
```
bkontur, line 276: @binier 
unwrap

binier: thats in test

tezedge_state/tests/simulate_one_peer_all_message_sequences.rs
binier: Yeah

bkontur (changes requested):
@binier @jurajselep @tizoc @akoptelov 
finally I walked throught all files :D,
it looks pretty cool :)

Zura, please check my comments,
and mainly, there are still lots of unwraps/expects especially in proposal loading/recording. I think, state machine should be bullet-proof, so any accidentally forgotten unwrap/expect/panic could break it.

so, I would apply here simple rules,
- never use expect/unwraps/panics in code, allways return Result and leave responsibility to caller to decide
- only allowed place for expect/unwraps/panics should be tests (and one exception for light/main.rs)

tezedge_state/src/state/requests.rs
bkontur: ou, sorry :)

akoptelov:
@binier  I would expect fixing at least compilation warnings before merging...

Issue #655, opened 2021-07-06
Need more configurable chain parameters


elric1:
When starting tezedge, I see that you can specify `--network=foo` which selects a series of parameters such as the chain name, protocol, genesis block, etc. An example from the code of such a setting can be found in the code here:
```
    env.insert(TezosEnvironment::Sandbox, TezosEnvironmentConfiguration {
        genesis: GenesisChain {
            time: "2018-06-30T16:07:32Z".to_string(),
            block: "BLockGenesisGenesisGenesisGenesisGenesisf79b5d1CoW2".to_stri
ng(),
            protocol: "PtYuensgYBb3G3x1hLLbCmcav8ue8Kyd2khADcL5LsT5R1hcXex".to_s
tring(),
        },
        bootstrap_lookup_addresses: vec![],
        version: "SANDBOXED_TEZOS".to_string(),
        protocol_overrides: ProtocolOverrides {
            user_activated_upgrades: vec![],
            user_activated_protocol_overrides: vec![],
        },
        enable_testchain: false,
        patch_context_genesis_parameters: Some(PatchContext {
            key: "sandbox_parameter".to_string(),
            json: r#"{ "genesis_pubkey": "edpkuSLWfVU1Vq7Jg9FucPyKmma6otcMHac9zG
4oU1KMHSTBpJuGQ2" }"#.to_string(),
        }),
    });
```
For the tezos-k8s project, we need to be able to set some of these parameters arbitrarily. Here's the configuration section that we are using:
```
node_config_network:
  activation_account_name: baker0
  chain_name: SANDBOXED_TEZOS
  genesis:
    block: BLockGenesisGenesisGenesisGenesisGenesisf79b5d1CoW2
    protocol: PtYuensgYBb3G3x1hLLbCmcav8ue8Kyd2khADcL5LsT5R1hcXex
    timestamp: '2018-06-30T16:07:32Z'
```
We use the "activation account" to find the `genesis_pubkey` that ends up in the config file. But, as you can see, we need for various other fields from the above code block to be configurable.

tizoc:
Hi @elric1, what are you using for Octez? a custom network in the context file? can you show me the custom network value you are using?

tizoc:
Hello @elric1. This is now implemented and released in `v1.6.5`.

[...]

PR #669, opened 2021-07-12
Add support for custom networks specified by a config file

tizoc:
Solves #655

Adds new `custom` network, and `--custom-network-file=custom-network.json` command line flag that is required when `--network=custom` is used.

The json file follows the same format as Octez's config file, but only the "network" key is used.

The contents of that file are used to initialize an environment for `TezosEnvironment::Custom`, which is then filled in when initializing `TEZOS_ENV` (quite hacky, any better idea of how to handle this?).

### DIFF
--- a/tezos/api/src/environment.rs
+++ b/tezos/api/src/environment.rs
@@ -1,8 +1,10 @@
 // Copyright (c) SimpleStaking, Viable Systems and Tezedge Contributors
 // SPDX-License-Identifier: MIT
 
-use std::path::PathBuf;
+use std::io;
+use std::path::{Path, PathBuf};
 use std::str::FromStr;
+use std::sync::RwLock;
 use std::{
     collections::HashMap,
     convert::{TryFrom, TryInto},
@@ -27,9 +29,12 @@ use crypto::{base58::FromBase58CheckError, blake2b::Blake2bError};
 use tezos_messages::p2p::encoding::prelude::{BlockHeader, BlockHeaderBuilder};
 
 use crate::ffi::{GenesisChain, PatchContext, ProtocolOverrides};
+use crate::octez_config::OctezConfig;
 
 lazy_static! {
     pub static ref TEZOS_ENV: HashMap<TezosEnvironment, TezosEnvironmentConfiguration> = init();
+    pub static ref CUSTOM_NETWORK_CONFIGURATION: RwLock<Option<TezosEnvironmentConfiguration>> =
+        RwLock::new(None);
 }
 
 pub const PROTOCOL_HASH_ZERO_BASE58_CHECK: &str =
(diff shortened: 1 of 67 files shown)


tizoc:
@bkontur I would like your thoughts on this one. In particular, the way I initialize the custom network is pretty hacky.

Not sure about the documentation, should we just duplicate the description of the custom network format specified in `tezos-node config --help`?

tezos/api/src/environment.rs
```diff
@@ -27,9 +29,12 @@ use crypto::{base58::FromBase58CheckError, blake2b::Blake2bError};
 use tezos_messages::p2p::encoding::prelude::{BlockHeader, BlockHeaderBuilder};
 
 use crate::ffi::{GenesisChain, PatchContext, ProtocolOverrides};
+use crate::octez_config::OctezConfig;
 
 lazy_static! {
     pub static ref TEZOS_ENV: HashMap<TezosEnvironment, TezosEnvironmentConfiguration> = init();
```
bkontur, line 35: @tizoc 
I dont like this `lazy_static TEZOS_ENV` , I would like to remove it totally,
I guess, now we dont need it, only just for tests,

so I would like to remove this lazy_static,
and then we just need to fix this:
```
let env = crate::configuration::Environment::from_args();
    let tezos_env = environment::TEZOS_ENV
        .get(&env.tezos_network)
        .unwrap_or_else(|| {
            panic!(
                "No tezos environment version configured for: {:?}",
                env.tezos_network
            )
        });
```

we just need to add `tezos_env: TezosEnvironmentConfiguration` to 
```
pub struct Environment {
    ...
    pub tezos_network: TezosEnvironment,
    ...
```

and join it with `tezos_network`


so in that main function we just call:
```
let env = crate::configuration::Environment::from_args();
let tezos_env = env.tezoz_env()
```

so, that actual hashmap, we dont need it, we just do that `init` as a part of `configuration.rs` on startup and there we just handle that custom network from file, like you did

I guess, this solution/refactor will clean the stuff a lot

tizoc: @bkontur sounds good to me. Only issue is that right now I have to work on things of much higher priority, so I will not be able to get back to this for some time. It can be merged with another TODO + jira for the refactor, or otherwise left for later.

tizoc:
Just saw that some tests are broken right now (because "Custom" network doesn't always have an environment).

bkontur:
@tizoc 
ok, let me check that, once I am done with my stuff,

tizoc:
@bkontur don't worry about the test failures, it is trivial to fix, I will push it in a few minutes once I have confirmed the fix (just filtering out `Custom` when iterating).

bkontur:
@tizoc 
tomorrow morning, I will rebase and check this also

bkontur (approved):

tizoc:
@bkontur cool, much much better.

PR #672, opened 2021-07-12
TE-584 / TE-585 Slog logging configuration refactor

bkontur:


tizoc:
@bkontur do we have a way to capture the output from OCaml other than seeing it on stdout?

From IPC calls we should always get the exceptions as part of the response because we have a catch-all that captures the backtrace and sends it back, but if the protocol runner process dies for any other reason that will not happen.

tizoc:
The exception capture I am talking about: https://gitlab.com/tezedge/tezos/-/blob/tezedge-ffi/src/lib_ffi/ffi_main.ml#L37-54

bkontur:
> 
> 
> @bkontur do we have a way to capture the output from OCaml other than seeing it on stdout?
> 
> From IPC calls we should always get the exceptions as part of the response because we have a catch-all that captures the backtrace and sends it back, but if the protocol runner process dies for any other reason that will not happen.


@tizoc 
maybe, you could try to set some stdout pipe/redirects here:
```
let process = Command::new(&self.executable_path)
            .arg("--sock-cmd")
            .arg(&self.sock_cmd_path)
            .arg("--endpoint")
            .arg(&self.endpoint_name)
            .arg("--log-level")
            .arg(&self.log_level.as_str().to_lowercase())
            .spawn()
```

and then `protocol_runner/src/main.rs` is using slog TermDecorator, so it should be probably added to the slog, or someting like that

PR #676, opened 2021-07-13
Log protocol runners stdout/stderr output

tizoc:
Very hacky implementation, not meant to be merged in its current state, but will be used for debugging.

What would be a better way to handle this? I saw that there is a [tokio::process](https://docs.rs/tokio/1.8.1/tokio/process/index.html) module that could be used to work with this in a more async way.


tizoc:
Ok, definitely a mix of what can be found in `tokio::process` and `tokio::io` is needed for a better solution that doesn't launch these extra threads, will look into it tomorrow.

tizoc:
@bkontur I pass the tokio runtime handle around to make things more explicit, but the alternative is to just obtain a handle through this method https://docs.rs/tokio/1.8.1/tokio/runtime/struct.Handle.html#method.current

Passing the handle around is more cumbersome but has the advantage of forcing you to make sure that things are initialized before trying to call stuff that depends on the tokio runtime being initialized (finding tests that needed to be updated for example was easy thanks to this).

Although I lean to the more cumbersome approach, I am ok with both, what do you think?

PR #682, opened 2021-07-14
Upgrade tokio dependency to 1.8

tizoc:


tizoc:
Merged as part of #676

PR #683, opened 2021-07-15
Bootstrap time test

akoptelov:


.drone.yml
```diff
@@ -5840,6 +5840,142 @@ steps:
       ref:
       - refs/pull/*/head
 
+---
+##############################################################################################################
+# This pipeline bootstraps a tezedge node to 1000 and measures memory allocation peaks during that
```
tizoc, line 5845: s/memory allocation peaks during that/the time it takes to apply those blocks/

akoptelov:
This shouldn't run on push, only for PR...

tizoc:
@akoptelov because the nodes are using the network for bootstrapping, there is going to be some variance between runs. A better way would be to compare the speed using the replay tool Sebastien wrote, which skips the network. Check TE-609 please, it is related. The idea would be to produce a snapshot through the peer based bootstrapping process we now use for the other tests, and then use that snapshot (for both this test and the replay test mentioned in TE-609).

.drone.yml
```diff
@@ -5840,6 +5840,145 @@ steps:
...
+      - name: tools
+        path: /tools
+    commands:
+      - OCTEZ_IP=$(perl -e 'use Socket; my $a = gethostbyname("octez-node"); print inet_ntoa($a)')
+      - git reset --hard origin/$${DRONE_TARGET_BRANCH}
+      - rm -rf ./tezos-node-data ./light-node-data
+      - cargo clean && cargo build --release --package light-node --package protocol-runner
+      - export LD_LIBRARY_PATH=tezos/sys/lib_tezos/artifacts/
+      - >
+        target/release/light-node --protocol-runner target/release/protocol-runner
```
bkontur, line 5945: @akoptelov 
we should create one pipeline which builds old tezedge binary (release and/or debug),
and once build, use `synchronize_ci.sh` to copy that binary to other servers,
and all pipelines which required old binary should depends on,
just to build old tezedge once

.drone.yml
```diff
@@ -5840,6 +5840,148 @@ steps:
...
+    image: tezedge/tezedge-ci-builder
+    user: root
+    volumes:
+      - name: build
+        path: /build
+      - name: tools
+        path: /tools
+    commands:
+      # let octez node to be up and running
+      - sleep 15
```
bkontur, line 5908: @akoptelov 
you dont need this sleep, it is not accurate,
we have a step for that:
```
  - name: wait-octez-node-to-start
    image: tezedge/tezos-node-bootstrap:latest
    pull: if-not-exists
    commands:
      - tezos-node-bootstrap bootstrap --level=0 --nodes $${OCTEZ_NODE_RPC_CONTEXT_ROOT}
```

.drone.yml
```diff
@@ -5840,6 +5840,148 @@ steps:
...
+      - OCTEZ_IP=$(perl -e 'use Socket; my $a = gethostbyname("octez-node"); print inet_ntoa($a)')
+      # use alloc instead of jemalloc by commenting out lines between these two
```
bkontur, line 5910: @akoptelov 
outdated comment here, I guess :)

tizoc:
@akoptelov about what I said above, it is wrong. Juraj told me earlier today that as long as the same identity is used, the order in which things are received from the network should stay the same.

akoptelov:
@bkontur Now this test and also wrk tests use prebuilt target branch binaries.

.drone.yml
```diff
@@ -5974,6 +5967,217 @@ steps:
...
+        path: /identities
+      - name: snapshots
+        path: /snapshots
+    commands:
+      - tezos-node config init --network florencenet
+      - cp /identities/identity_1.json $${TEZOS_NODE_DIR}/identity.json
+      - tezos-node snapshot import /snapshots/octez-009-florence-$${BLOCK_LEVEL}.snapshot
+      - tezos-node run --no-bootstrap-peers --rpc-addr 0.0.0.0:8732
+
+  - name: bootstrapping-tezedge-target-branch
```
bkontur, line 6028: @akoptelov 
`bootstrapping-tezedge-target-branch` -> `bootstrapping-octez-bootstrap-node`

bkontur (approved):
@akoptelov 
very cool, you simplify that a lot with that old build,

maybe we could also store that old build under commit hash, so it could be even reused between pipeline,  but this is very nice-to-have, but build it everytime is ok and clean

bkontur:
@akoptelov 
pls. fix that one typo and rebase/squash/merge to develop please

PR #686, opened 2021-07-15
Collect timing messages into a single bigger message before sending

tizoc:
Basic initial version, I think there are still possible improvements. For example, preallocate N buffers that get cycled around instead of allocating a new one when sending a combined message. Whenever the receiving side is done processing the buffer, it can return it back to the producer.

## Some measurements:

First 11k blocks from florencenet applied with replay-test with the Irmin context and original libtezos without the latest modifications (which save a bit of time but not enough to matter).

###  timings disabled

```
51.82user 5.90system 1:02.61elapsed 92%CPU (0avgtext+0avgdata 474336maxresident)k
2481784inputs+1642160outputs (497major+158501minor)pagefaults 0swaps
```

### timings enabled

```
102.90user 24.76system 1:25.54elapsed 149%CPU (0avgtext+0avgdata 479560maxresident)k
753112inputs+4166736outputs (295major+160016minor)pagefaults 0swaps
```

### timings enabled with packing (100 messages in one)

```
70.45user 7.45system 0:58.52elapsed 133%CPU (0avgtext+0avgdata 481160maxresident)k
80inputs+4132344outputs (0major+159147minor)pagefaults 0swaps
```

Tried bigger and smaller message buffer values but it didn't make much of a difference.

### DIFF
--- a/tezos/timing/src/lib.rs
+++ b/tezos/timing/src/lib.rs
@@ -2,10 +2,14 @@
 // SPDX-License-Identifier: MIT
 
 use std::{
+    cell::Cell,
     collections::HashMap,
     convert::TryInto,
     path::PathBuf,
-    sync::mpsc::{sync_channel, Receiver, SyncSender},
+    sync::{
+        mpsc::{sync_channel, Receiver, SendError, SyncSender},
+        Mutex,
+    },
     time::{Duration, Instant},
 };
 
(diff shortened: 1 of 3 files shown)


tezos/timing/src/lib.rs
```diff
@@ -278,8 +282,61 @@ pub struct Action {
...
+    }
+
+    /// Sends messages, delayed and combined into a single bigger message.
+    ///
+    /// Reaching the delayed messages limit or receiving a commit message will trigger the send
+    /// to the underlying channel, otherwise the messages will be kept in the buffer.
+    pub fn send(&self, msg: TimingMessage) -> Result<(), SendError<Vec<TimingMessage>>> {
+        let must_not_delay = self.is_immediate_message(&msg);
+        let limit = Self::DELAYED_MESSAGES_LIMIT - 1;
+        let mut buffer = self.buffer.lock().unwrap();
```
bkontur, line 320: @tizoc 
please, check this unwrap

bkontur (approved):
@tizoc 
I guess, ok, just check that unwrap and clippy stuff please

tezos/timing/src/lib.rs
```diff
@@ -278,8 +282,61 @@ pub struct Action {
...
+
+            self.sender.send(pack)
+        } else {
+            Ok(())
+        }
+    }
+}
+
+pub static TIMING_CHANNEL: Lazy<BufferedTimingChannel> = Lazy::new(|| {
+    let (sender, receiver) = sync_channel(1000);
```
bkontur, line 339: @tizoc 
is this enought? e.g. for big block with protocol switching?

tezos/timing/src/lib.rs
tizoc: Uff, good catch.

tezos/timing/src/lib.rs
tizoc: We actually have a bigger buffer now if you combine everything. The original buffer was 10k, now it is 10 time less but we are combining 100 messages into one, so thats 10x capacity (100k messages in total). Reverting it to 10k will not hurt tho.

tizoc:
@bkontur ok, I increased the channel buffer back to 10k, addressed the clippy warnings and got rid of that unwrap.

I am sure we can lower the overhead even more, but will require more time. I would like to know if you have any other ideas @sebastiencs

bkontur (approved):

sebastiencs:
@bkontur @tizoc I'm suprised to see that timing is making such difference, or even a difference at all.
I don't have other ideas right now but yes I will find a way to make timing with zero overhead

tizoc:
@sebastiencs I wasn't expecting it either, specially not on the channel send, but after changing everything else that is from where it most of it come from, will show you the numbers on slack.

PR #689, opened 2021-07-16
Drone ci tests for multibackend

mambisi:


.drone.yml
```diff
@@ -398,6 +398,36 @@ steps:
...
+  image: tezedge/tezedge-ci-builder:latest
+  pull: if-not-exists
+  user: root
+  environment:
+    SODIUM_USE_PKG_CONFIG: 1
+    RUST_BACKTRACE: 1
+  commands:
+    - cd storage
+    - cargo test --release --features maindb-backend-sled
+- name: run-unit-tests-commit-log
```
bkontur, line 421: @mambisi please add a new line before all `-name` step

bkontur:
@mambisi 
please, clean all those unused imports, everywhere
image

.drone.yml
mambisi: Ok

storage/src/lib.rs
```diff
@@ -703,29 +703,35 @@ pub mod tests_common {
...
+                        PredecessorStorage::descriptor(&db_cache),
+                        BlockAdditionalData::descriptor(&db_cache),
+                    ],
+                    &cfg,
+                )?);
+                TezedgeDatabaseBackendOptions::RocksDB(database::rockdb_backend::RocksDBBackend::from_db(kv)?)
+            } else if cfg!(feature = "maindb-backend-sled") {
+                TezedgeDatabaseBackendOptions::SledDB(database::sled_backend::SledDBBackend::new(path.join("db"))?)
+            } else {
+                TezedgeDatabaseBackendOptions::SledDB(database::sled_backend::SledDBBackend::new(path.join("db"))?)
```
bkontur, line 729: @mambisi @jurajselep 
do we want to run tests by default with `sled` or `rocksdb` ?
I would say better rocksdb, because it is default for node also,
so please change it to `rocksdb`

mambisi: Ok i will change the default to RockDB

storage/tests/sequences.rs
```diff
@@ -97,15 +97,23 @@ fn generator_test_batch() -> Result<(), Error> {
...
+            let db = open_kv(
+                &path,
+                vec![Sequences::descriptor(&cache)],
+                &DbConfiguration::default(),
+            )?;
+            TezedgeDatabaseBackendOptions::RocksDB(database::rockdb_backend::RocksDBBackend::from_db(Arc::new(db))?)
+        } else if cfg!(feature = "maindb-backend-sled") {
+            TezedgeDatabaseBackendOptions::SledDB(database::sled_backend::SledDBBackend::new(path.join("db"))?)
+        } else {
+            TezedgeDatabaseBackendOptions::SledDB(database::sled_backend::SledDBBackend::new(path.join("db"))?)
```
bkontur, line 112: @mambisi 
also default rocksdb

storage/tests/sequences.rs
```diff
@@ -127,15 +135,22 @@ fn generator_test_continuation_after_persist() -> Result<(), Error> {
...
```
bkontur, line 149: @mambisi 
also rocksdb

.drone.yml
```diff
@@ -398,6 +398,36 @@ steps:
...
+  image: tezedge/tezedge-ci-builder:latest
+  pull: if-not-exists
+  user: root
+  environment:
+    SODIUM_USE_PKG_CONFIG: 1
+    RUST_BACKTRACE: 1
+  commands:
+    - cd storage
+    - cargo test --lib commit_log -- --nocapture
```
bkontur, line 430: @mambisi 
please add here also `--release`  because all before was with `--release`  so this last one, does not need to compile/download all deps for debug

bkontur (changes requested):
@mambisi 
please check PR comments

.drone.yml
mambisi: Ok

bkontur (approved):

PR #690, opened 2021-07-16
open api specification for tezedge

mambisi:
This PR add open api specification to tezedge

### DIFF
--- a/spec/tezedge-openapi.json
+++ b/spec/tezedge-openapi.json
@@ -1,0 +1,5292 @@
+{
+  "openapi": "3.0.0",
+  "info": {
+    "title": "Tezedge",
+    "termsOfService": "https://www.viablesystems.io/",
+    "contact": {
+      "name": "API Support",
+      "url": "https://github.com/tezedge/tezedge",
+      "email": "jurajselep@viablesystems.io"
+    },
+    "license": {
+      "name": "MIT License",
+      "url": "https://github.com/tezedge/tezedge/blob/master/LICENSE"
+    },
+    "version": "1.6.6"
+  },
+  "servers": [
+    {
+      "url": "http://master.dev.tezedge.com:{port}",
+      "description": "Development server",
+      "variables": {
+        "port": {
+          "enum": [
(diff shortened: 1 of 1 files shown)


spec/florence.yml
```diff
@@ -0,0 +1,10890 @@
+openapi: 3.0.0
+info:
+  title: Tezedge Protocol RPCs
+  termsOfService: https://www.viablesystems.io/
+  contact:
+    name: API Support
+    url: https://github.com/tezedge/tezedge
+    email: support@example.com
```
bkontur, line 8: @mambisi @jurajselep 
Mambisi, please put here some correct email

mambisi:
Which email though? @jurajselep

spec/florence.yml
```diff
@@ -0,0 +1,10890 @@
...
+  license:
+    name: MIT License
+    url: https://github.com/tezedge/tezedge/blob/master/LICENSE
+  version: 1.5.0
```
bkontur, line 12: @mambisi 
put here please: `v1.6.6`

spec/tezedge-openapi.json
```diff
@@ -0,0 +1,5293 @@
+{
+  "openapi": "3.0.0",
+  "info": {
```
bkontur, line 3: @mambisi @jurajselep 
please, make sure that information are correct and relevant (email, links) and not like `This is a sample server for a pet store.`

and copy the same information to all files

bkontur:
@mambisi @jurajselep @directcuteo 
and I would better move that `spec` directory to the rpc: `rpc/spec`,
I dont know how explorer is reading that, but we should expose this files throught RPC,
like:
`http://localhost:18732/openapi/tezedge-openapi.json`

bkontur (changes requested):
@mambisi please PR comments

spec/florence.yml
mambisi: Ok

bkontur (approved):

PR #705, opened 2021-07-22
Disable throttling on pipelines that bootstrap the node from a peer

tizoc:


bkontur:
@akoptelov 
do we have other/specific tests for throttling?

akoptelov:
There is one, still locally in WIP, in `shell/tests`.

PR #706, opened 2021-07-22
Fix parsing of pool cli argument limits.

tizoc:
Readonly api pool limits were being ignored.

### DIFF
--- a/light_node/src/configuration.rs
+++ b/light_node/src/configuration.rs
@@ -121,9 +121,11 @@ pub struct Ffi {
 }
 
 impl Ffi {
+    // These are used for parsing the command-line arguments, hence the
+    // "-" suffix for the non-empty versions.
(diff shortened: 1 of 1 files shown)


light_node/src/configuration.rs
```diff
@@ -676,20 +676,26 @@ fn pool_cfg(
     args: &clap::ArgMatches,
     pool_name_discriminator: &str,
 ) -> TezosApiConnectionPoolConfiguration {
+    // Adjust this so that it properly matches the command-line argument.
```
akoptelov, line 679: I would suggest update constants values instead of constructing them in run-time.

akoptelov (commented):
Just one comment on constants.

light_node/src/configuration.rs
tizoc: Done. Better now?

PR #707, opened 2021-07-23
[TEST] Test asan failure

akoptelov:


networking/src/p2p/peer.rs
```diff
@@ -228,8 +222,6 @@ pub struct Peer {
     peer_id_marker: String,
     peer_metadata: MetadataMessage,
     peer_compatible_network_version: NetworkVersion,
-    throttle_quota: Arc<std::sync::Mutex<quota::ThrottleQuota>>,
```
bkontur, line 231: @akoptelov 
are we reverting throthling?

tizoc: I think he is just experimenting. This ASAN issue started manifesting itself when the throttling was merged.

akoptelov: Yeah, I tried to remove throttling manually, but that didn't work, the test is still getting killed. I'll try to do proper `git revert`, possibly I missed something.

PR #708, opened 2021-07-23
Bindings for Tezos lib_storage (Irmin and TezEdge in-mem) contexts

tizoc:


bkontur:
@tizoc @sebastiencs @akoptelov 
we should discuss new structure of `tezos` package:
**now:**
api
client
encoding
encoding-derive
identity
interop
lib_storage_ffi
messages
new_context
python-tests
spsc
sys
timing
wrapper

it gets little bit messy:

maybe we should group related stuff together:
`api`
`identity`
`encoding/encoding` (or better name, base)
`encoding/derive`
`encoding/messages`
`interop/client` (merge actual interop with client)
`interop/sys`
`python-tests`
`wrapper`
`context/new_context` (rename `new_context`, find better name)
`context/timing`
`context/lib_storage_ffi`
`spsc`  (not sure where to put this, this is not even tezos specific, maybe put it higher, and reused in other modules)

tizoc:
@bkontur `spsc` should go elsewhere for sure, it is a quite generic utility. Maybe we need a new namespace for that kind of code. The structure you mention sounds good to me, except `context/new_context` which need a better name (`new_context` was a temporary name because for some time we had both `context` and `new_context`). There is also some cleanup that could be done, I think right now we have more layers than needed (every time I want to add a new call I have to touch many places).

PR #710, opened 2021-07-27
Feat: configuration option to limit riker's threadpool size

binier:
Memory usage grows upto 3gb for node when using server with higer(64) cpu core count.

As it seems this is caused by riker's threadpool as it grows with the number of cores in the system, so we should have a way to restrict that.


light_node/src/main.rs
```diff
@@ -233,10 +233,19 @@ fn block_on_actors(
             .num_of_peers_for_bootstrap_threshold(),
     );
 
+    let mut actor_system_cfg = riker::load_config();
```
bkontur, line 236: @binier 
hmm, env.riker_threads is not used anywhere,
I would better do it like this, if it is set, then use it, else use cpu_get

```
    if env.riker_threads > 0 {
        actor_system_cfg
            .set("dispatcher.pool_size", env.riker_threads as i64)
            .unwrap();
    } else {
        let thread_count = env.riker_threads.min(num_cpus::get());
        actor_system_cfg
            .set("dispatcher.pool_size", env.riker_threads as i64)
            .unwrap();
    }
```

light_node/src/main.rs
```diff
@@ -233,10 +233,19 @@ fn block_on_actors(
...
+    if env.riker_threads > 0 {
+        let thread_count = env.riker_threads.min(num_cpus::get());
+        actor_system_cfg
+            .set("dispatcher.pool_size", thread_count as i64)
+            .unwrap();
```
bkontur, line 241: @binier 
please use `expect` with message instead of unwrap

light_node/src/main.rs
```diff
@@ -233,10 +233,19 @@ fn block_on_actors(
             .num_of_peers_for_bootstrap_threshold(),
     );
 
+    let mut actor_system_cfg = riker::load_config();
+    if env.riker_threads > 0 {
+        let thread_count = env.riker_threads.min(num_cpus::get());
+        actor_system_cfg
+            .set("dispatcher.pool_size", thread_count as i64)
```
bkontur, line 240: @binier 
could you, please, take a look if it is possible to change this:
```
.name_prefix("pool-thread-#") -> riker-thread-#
```

bkontur (changes requested):
@binier 
please check PR comments

light_node/src/main.rs
binier: `let thread_count = env.riker_threads.min(num_cpus::get());`

It is used bellow. We calculate minimum of available cores and `env.riker_threads`

If `riker_threads == 0`, then we don't need to override `dispatcher.pool_size` at all.

light_node/src/main.rs
binier: Looked into it and not possible. Everything which could give us ability to change that is private and not exposed in public interface by riker.

light_node/src/main.rs
binier: It rikers source code inside `riker::load_config()` unwrap is used for that. But ok will replace it with `expect`.

light_node/src/main.rs
bkontur: @binier 
ahaaa, now I see, it calculates min

bkontur (approved):

bkontur (approved):
@binier 
ok, merge please

PR #719, opened 2021-07-30
Update RocksDB dependency

binier:
- Updates RocksDB bindings (**rust-rocksdb**) from **0.15** to **0.17**.
- In latest master/develop, rocksdb isn't compiled with **frame pointers** enabled, because the [patch](https://github.com/tezedge/tezedge/blob/36a4b593cc6c60c599474f35a6ed3ff7b6a7c85f/Cargo.toml#L42) is being ignored. You can see message when compiling: 
  ```
  warning: Patch `librocksdb-sys v6.11.4 (https://github.com/tezedge/rust-rocksdb.git?tag=v1.5.0-tezedge-compatible#717c99c7)` 
  was not used in the crate graph.
  Check that the patched package version and available features are compatible
  with the dependency requirements. If the patch has a different version from
  what is locked in the Cargo.lock file, run `cargo update` to use the new
  version. This may also occur with an optional dependency that is not enabled.
  ```
  And in **Cargo.lock**: 
  https://github.com/tezedge/tezedge/blob/36a4b593cc6c60c599474f35a6ed3ff7b6a7c85f/Cargo.lock#L4327-L4330
  
  This pr should fix that too.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,4 +39,4 @@ members = [
 
 [patch.crates-io]
 ocaml-boxroot-sys = { git = "https://gitlab.com/bruno.deferrari/ocaml-boxroot.git", branch = "ocaml-410-headers" }
-librocksdb-sys = { git = "https://github.com/tezedge/rust-rocksdb.git", tag = "v1.5.0-tezedge-compatible" }
+librocksdb-sys = { git = "https://github.com/tezedge/rust-rocksdb.git", tag = "tezedge-v0.17.0-1" }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -309,9 +309,9 @@ dependencies = [
 
 [[package]]
 name = "bindgen"
-version = "0.57.0"
+version = "0.59.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fd4865004a46a0aafb2a0a5eb19d3c9fc46ee5f063a6cfc605c69ac9ecf5263d"
+checksum = "453c49e5950bb0eb63bb3df640e31618846c89d5b7faa54040d76e98e0134375"
 dependencies = [
  "bitflags",
  "cexpr",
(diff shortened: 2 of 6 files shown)


bkontur:
@binier @vlad9486 
`rocksdb isn't compiled with frame pointers enabled`
we did not want to use frame-pointers by default, but only on-demand with `RUSTFLAGS=-Cforce-frame-pointers=yes`
so this is still working, right?

Cargo.toml
```diff
@@ -39,4 +39,4 @@ members = [
 
 [patch.crates-io]
 ocaml-boxroot-sys = { git = "https://gitlab.com/bruno.deferrari/ocaml-boxroot.git", branch = "ocaml-410-headers" }
-librocksdb-sys = { git = "https://github.com/tezedge/rust-rocksdb.git", tag = "v1.5.0-tezedge-compatible" }
+librocksdb-sys = { git = "https://github.com/tezedge/rust-rocksdb.git", tag = "tezedge-v1.0.0" }
```
bkontur, line 42: @binier @vlad9486 
I checked that fork, but this tag is not based on rocksdb `v0.17.0`, why not?
but `origin/tezedge` is based on rocksdb `v0.17.0`
shouldnt be all above tag `v0.17.0`??

and I would change our tag from `tezedge-v1.0.0` to `tezedge-v0.17.0`, it will be much clear



image

binier: image
It's 1 commit after release of 0.17, just to add enable frame pointers.

About the tag, I just wanted the tag to not be based on anything, so that if we want to change something in rocksdb.

If we don't do it this way, there are two options:
- Have just one tag: `tezedge-v0.17.0`, and when changing **rust-rocksdb** in any way, we simply move that tag to newer commit. That will be problematic since when moving the tag, older releases might break.
- We could change version's patch number, but then if **rust-rocksdb** releases a patch, that will make version misleading.

Another option and probably best is to add revision number or something like that to tag, and on change we increment that `tezedge-v0.17.0-1`, `tezedge-v0.17.0-2`, etc...

bkontur: @binier 
very well, I like your last option, `tezedge-v0.17.0-1`,
we will exactly know which version of rocksdb it is and also, which is our fix/patch version

bkontur: @binier 
ignore that rebase studd, now I see it, my repo was not updated correctly
```
 ! [rejected]        tezedge-v1.0.0 -> tezedge-v1.0.0  (would clobber existing tag)
```

binier:
> @binier @vlad9486
> `rocksdb isn't compiled with frame pointers enabled`
> we did not want to use frame-pointers by default, but only on-demand with `RUSTFLAGS=-Cforce-frame-pointers=yes`
> so this is still working, right?

Yes that is the case, it's not default. But before this pr, even with that env you wouldn't get frame pointers.

bkontur (approved):
@binier 
cool, lets wait for drone CI to pass, then merge this please

bkontur (approved):

PR #728, opened 2021-08-03
Fix find_tree, must work for directories and values

tizoc:


tizoc:
Replaced by #734 which is simpler and actually solves the problem

tezos/new_context/src/working_tree/working_tree.rs
```diff
@@ -451,16 +451,38 @@ impl WorkingTree {
...
         let mut storage = self.index.storage.borrow_mut();
 
-        let tree = self.find_raw_tree(root, key, &mut storage)?;
-
-        if tree.is_empty() {
-            return Ok(None);
+        if let Ok(tree_id) = self.find_raw_tree(root, path, &mut storage) {
+            if let Some(node_id) = storage.get_tree_node_id(tree_id, *file) {
+                let node = storage.get_node(node_id)?;
+                match node.get_entry() {
```
tizoc, line 467: @sebastiencs I need some help here because I may be misunderstanding how things are supposed to work here. I am getting a node as a result to the `get_node(node_id)` call, but then from `node.get_entry()` I get a `None`. What am I missing?

tizoc: Also, in this example if I comment `A` or `B` it works, but otherwise it doesn't. There seems to be an issue related to how the `storage` data is filled when a checkout is made.

```
    context = context.index.checkout(&context_hash_1)?.expect(&format!(
        "Commit not found: {}",
        context_hash_1.to_base58_check()
    ));
    // == A ==
    // context = context.add(
    //     &context_key!("data/rolls/owner/current/index/123"),
    //     &[1, 2, 3, 4, 5, 6, 7],
    // )?;
    // == B ==
    //assert!(context
    //    .find_tree(&context_key!("data/rolls/owner/current/index/123"))?
    //    .is_some());
    assert!(context
        .find_tree(&context_key!("data/rolls/owner/current/index"))?
        .is_some());
```

tizoc: If I do a `mem` on a children in that path that turns out to be a leaf (a value and not a directory), then `find_tree` also works as expected after that call.

tizoc: Ok, turns out I have to use `self.index.node_entry(node_id, &mut storage)`, it is that function that will try to retrieve the entry and put it in the storage.

PR #729, opened 2021-08-04
Fix chrono panics

binier:


shell/src/validation/mod.rs
```diff
@@ -516,6 +516,35 @@ mod tests {
...
+            .fitness(fitness!([0], [0, 0, 1]))
+            .context(
+                "CoVmAcMV64uAQo8XvfLr9VDuz7HVZLT4cgK1w1qYmTjQNbGwQwDd"
+                    .try_into()
+                    .unwrap(),
+            )
+            .protocol_data(vec![0, 1, 2, 3, 4, 5, 6, 7, 8])
+            .build()
+            .unwrap();
+        is_future_block(&block_header);
```
bkontur, line 545: @binier 
please add here assert
```
assert!(is_future_block(&block_header).is_err());
or 
assert!(is_future_block(&block_header).is_ok());
```

bkontur (approved):
@binier 
cool, please check that one assert and please rebase/squash/merge to develop

shell/src/validation/mod.rs
binier: added.

rpc/src/helpers.rs
```diff
@@ -164,20 +164,20 @@ impl BlockHeaderShellInfo {
...
         block: &BlockHeaderWithHash,
         block_json_data: &BlockJsonData,
         block_additional_data: &BlockAdditionalData,
         chain_id: &ChainId,
-    ) -> Self {
+    ) -> Result<Self, TimestampOutOfRangeError> {
         let header: &BlockHeader = &block.header;
         let predecessor = header.predecessor().to_base58_check();
-        let timestamp = ts_to_rfc3339(header.timestamp());
+        let timestamp = ts_to_rfc3339(header.timestamp())?;
```
akoptelov, line 180: I'm not sure if that is possible, but please check if it would be easier to perform the conversion on use, not on construction.

rpc/src/services/base_services.rs
```diff
@@ -39,6 +39,14 @@ pub enum RpcServiceError {
     StorageError { error: StorageError },
     #[fail(display = "No data found error, reason: {:?}", reason)]
     NoDataFoundError { reason: String },
+    #[fail(display = "Invalid or out-of-range datetime")]
+    TimestampOutOfRangeError,
+}
+
+impl From<TimestampOutOfRangeError> for RpcServiceError {
+    fn from(_: TimestampOutOfRangeError) -> Self {
```
akoptelov, line 47: Can you also add the error itself to the resulting one? That is cleaner even if the underlying error is a unit structure, like this one.

binier:
> I would suggest a small fix on error handling. Another thing is that sometimes it might be more useful to do the conversion later, but I'm not sure if that is possible and feasible here. And finally, why is this conversion function defined in `tezos_messages` while it is only used in `rpc`?

Not sure why it's defined it there, but probably it's since that function is used on tezos messages only, not sure if that makes sense. Anyways, this is more of a quickfix and didn't want to change much from the code itself, just get rid of panics.

rpc/src/services/base_services.rs
binier: Ok can do.

PR #730, opened 2021-08-04
Split build into two pipelines

akoptelov:


.drone.yml
```diff
@@ -5819,6 +5878,7 @@ trigger:
 
 depends_on:
   - corr/real-time-environment-cyclictest-latencies
+  - build-old-tezedge-binaries
```
bkontur, line 5881: @akoptelov 
maybe I would add here also `build-tezedge-binaries` just to be explicit

bkontur (approved):
@akoptelov 
cool, rebase/squash/merge please

PR #733, opened 2021-08-04
Fix build paths for tests that have it wrong

tizoc:


tizoc:
Replaced by #734 which is simpler and actually solves the problem

PR #739, opened 2021-08-05
Te 652 rikerout

bkontur:


light_node/src/main.rs
```diff
@@ -364,6 +364,7 @@ fn block_on_actors(
             shell_compatibility_version,
             env.p2p,
             env.identity.expected_pow,
+            Arc::new(init_storage_data.chain_id),
```
binier, line 367: Why `Arc` ? Do we need chain_id in `PeerManager`?

binier (changes requested):
Please check comments and rebase your branch to latest **p2p-simulator** branch.

light_node/src/main.rs
bkontur: @tizoc 
arc removed, chain_id is needed in tezedge_state

bkontur:
@binier @jurajselep 
Zura, could you please check this PR, when ok, I would like to merge it to p2p-simulator branch, there is:
- a removed peer_manager actor + changed light-node initialization of TezedgeStateManager (state + proposer)
- removed usage of network_channel for several shell actors (chain_manager, peer_branch_bootstrapper) replaced by proposer

this is something, which you have to do anyway and it makes other things easier for both

shell/src/peer_manager.rs
```diff
@@ -91,17 +95,26 @@ impl From<std::io::Error> for NotifyProposerError {
     }
 }
 
+impl From<std::sync::mpsc::SendError<ProposerMsg>> for NotifyProposerError {
+    fn from(err: std::sync::mpsc::SendError<ProposerMsg>) -> Self {
+        Self::SendError(err)
+    }
+}
+
+pub type ProposerHandleRef = ProposerHandle;
```
binier, line 104: Why the alias?

binier (changes requested):
PeerManager actor still is here. Can we get rid of it? Doesn't look like it's used anywhere.

shell/src/peer_manager.rs
bkontur: @binier 
peer_manager is not used anywhere, I just did not want to remove it to aviod rebase conflicts,
but when you approve this, I will clean it and merge to your branch

bkontur: @binier 
to that alias, I had there before:
```
pub type ProposerHandleRef = Arc<ProposerHandle>;
```
not needed, removing

bkontur: @binier 
removed all peer_manger riker actor related stuff + removed ProposerHandleRef

binier (approved):
Looks good! Please rebase to latest **p2p-simulator**, let's make sure tests pass and we can merge it.

bkontur:
@binier 
ok, thx, merged,

tests passed, except these two:
- benchmark for stream decoding was removed (we should rework it or remove from drone)
-- http://ci.tezedge.com/tezedge/tezedge/2040/34/7
- bootstrap time test 
-- http://ci.tezedge.com/tezedge/tezedge/2040/34/7
-- we have now slower bootstrap with state machine (I guess, once we are done with refactor, it will be fixed or even better)

but both are also problem for `origin/p2p-simulator` branch:
http://ci.tezedge.com/tezedge/tezedge/2024/34/7
http://ci.tezedge.com/tezedge/tezedge/2024/29/5
so they are not related to this PR

PR #746, opened 2021-08-06
Networking measurement tweaks

adonagy:


apps/node_monitoring/Cargo.lock
```diff
@@ -876,7 +876,7 @@ dependencies = [
 [[package]]
 name = "netinfo"
 version = "0.5.1"
-source = "git+https://github.com/adonagy/netinfo.git?branch=fix/brokend-dependecies#615e483fcbfccd0c5750ddd447a60c0b361f0bb9"
+source = "git+https://github.com/adonagy/netinfo.git?branch=fix/brokend-dependecies#99bedb9353478f0251227f247226ba2e2881991b"
```
bkontur, line 879: @adonagy 
please, anything should depends on fixed tagged version and not on branch,
we had already problem with this with rocksdb fork for frame-pointers, where it depended on branch, we should change it to tag

bkontur (changes requested):
@adonagy
please, anything should depends on fixed tagged version and not on branch,
we had already problem with this with rocksdb fork for frame-pointers, where it depended on branch, we should change it to tag

bkontur:
@adonagy 
and another thing, could you please move those test in drone.yml before perf tests. that corr tests will be together :)

bkontur (approved):

PR #747, opened 2021-08-06
Fuzz target

dkuehr:
A new _fuzz_ profile is included to generate an optimized build of TE while keeping debug symbols, frame pointers, and runtime checks. Some _rustc_ flags can't be applied directly in the `Cargo.toml` so changes were also needed in run.sh and this should be the way to build it: `./run.sh fuzz [rest of node arguments]`

This also introduces "dynamic" coverage by installing a signal handler (SIGUSR2) in light-node which calls the llvm-gcov internal functions that dump the program counters into `.gcda` files.
A new `poll.sh` script is included inside a `coverage/` directory, when run it will send the signal to the light-node process and generate `.gcov` files via `rust-cov`.

This requires (not needed for normal builds):
- `cargo install cargo-binutils`
- `rustup component add llvm-tools-preview`

With this new basic functionality we can:
- Run fuzzers against builds that keep most debugging features but are faster.
- Write new tools to monitor the node's coverage dynamically by calling the `poll.sh` script and parsing the resulting `.gcov` files.
-

### DIFF
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -10,6 +10,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+            toolchain: nightly
+            default: true
+            override: true
       - uses: actions-rs/audit-check@v1
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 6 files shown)


tizoc (approved):

tizoc:
sanitizer failure is unrelated to this change, it is in the ffi context tests, probably because the OCaml runtime is not being properly cleaned up

PR #748, opened 2021-08-06
Make storage independent from libtezos

vlad9486:


tizoc:
@bkontur the solution is quite a bit more complicated than this (already discussed it with @vlad9486). I was going to solve it as part of the static linking changes, but I started working on it already as part of a standalone PR. There are a lot of "false" dependencies to libtezos from the shell and the light node binary, what needs to be done is to move stuff around (mostly configuration structs and error type definitions) so that those can be depended on without pulling all the libtezos stuff in.

tizoc:
Closing.

PR #750, opened 2021-08-09
P2P latency benchmark in CI

mambisi:
Add node's P2P latency benchmark in CI
Closes https://github.com/tezedge/tezedge/pull/726


.drone.yml
```diff
@@ -6009,7 +6009,119 @@ depends_on:
...
+          block=$${b:-$block}
+          echo "===> Block level $block"
+        done
+        '
+      - echo "Bootstrapped. Killind the node and waiting for 10 seconds."
+      # terminate the node gracefully and wait to allow `valgrind` dump data
+      - kill -INT $(cat /var/run/tezedge.pid) && sleep 10
+      # restart node for client to connect
+      # ------------------------------------------
+      - git clone https://github.com/mambisi/tezedge.git --branch storage-latency-temp latency
```
bkontur, line 6080: @mambisi @jurajselep 
what about `storage-latency-temp` branch, we should better merge it to tezedge repo, right?

.drone.yml
```diff
@@ -6009,7 +6009,119 @@ depends_on:
   - build-tezedge-binaries
   - build-old-tezedge-binaries
   - corr/real-time-environment-cyclictest-latencies
+---
+##############################################################################################################
+# This pipeline calculates the node's p2p latency
```
bkontur, line 6014: @mambisi @jurajselep 
here is "node's latency",
but step is named "perf/node-latency",
but you use binary which is called: `storage-metrics`,
so could you please clean that all up, 
that we exactly now what does this test do?
so at least, please name stuff correctly,
and please add here a short description, how does it work, like for example check: `wrk-performance-test-tezedge-new-vs-octez`

.drone.yml
bkontur: @mambisi @jurajselep 
I dont like, that we are referecing your fork git :),
but like you told me, it depends on Zura's stuff, which is not merged yet,
at least make you branch `storage-latency-temp` stable and tag it and then here for clone, use a that tag please

mambisi: Ohk I will do that

.drone.yml
mambisi: Ohk

PR #752, opened 2021-08-10
Te 652 rikerout

bkontur:


rpc/src/server/rpc_server.rs
```diff
@@ -0,0 +1,281 @@
...
+        if let Some(RpcServerThreadHandle::NotRunning(notify_rpc_server_rx)) =
+            self.rpc_server_thread_handle.take()
+        {
+            // 1. spawn mpsc queue receiver
+            // start to listen for incoming p2p connections and state machine processing
+            let notify_rpc_server_listener_thread = {
+                let env = self.env.clone();
+
+                std::thread::Builder::new()
+                    .name("rpc-notify-lstnr-thread".to_owned())
```
binier, line 119: Typo here, should be `"rpc-notify-listener-thread"`. Or is it for some reason to make name shorter?

bkontur: @binier @adonagy 
Ado, told me that, for some thread explorer, the linux uses just 15 char for thread name,
so I could rewrite it like `rpc-ntf-lstnr-thrd` :)

binier: Oh ok, makes sense. "thread" word is probably extra, so you could have just `rpc-notifier` or something like that.

bkontur: @binier 
right, so we have now:
`rpc-notifier`
`chainman-notifier`
and coming soon `ws-monitor-notifier`

PR #753, opened 2021-08-13
Calculate block header hash on decoding

akoptelov:
Now the block hash is calculated on encoding and stored along with other fields. Similar approach can be used for other messages, but I would prefer to finish with removing `serde` first.


storage/src/lib.rs
```diff
@@ -71,10 +71,13 @@ pub struct BlockHeaderWithHash {
...
-        Ok(BlockHeaderWithHash {
-            hash: block_header.message_hash()?.try_into()?,
-            header: Arc::new(block_header),
-        })
+        let hash = if let Some(hash) = block_header.hash().as_ref() {
+            hash.as_slice().try_into()?
+        } else {
+            block_header.message_hash()?.try_into()?
+        };
+        let header = Arc::new(block_header);
```
bkontur, line 79: @akoptelov @tizoc 
I am still thinking about this structure,
for example everywhere, we construct BlockHeaderWithHash, we need to clone BlockHeader data:
```
BlockHeaderWithHash::new(message.current_block_header().clone())?;
```
and I think it is also wastefull
for example, what if all messages like: `message.current_block_header`, has `Arc<BlockHeader>` instead of `BlockHeader`, what do you think?

tezos/messages/src/p2p/encoding/block_header.rs
```diff
@@ -92,11 +92,49 @@ pub struct BlockHeader {
...
+
+    #[get = "pub"]
+    #[serde(skip)]
+    #[builder(default)]
+    #[encoding(hash)]
+    hash: EncodingHash,
+}
+
+#[derive(Clone, Debug)]
+pub struct EncodingHash(pub Option<Vec<u8>>);
```
bkontur, line 104: @akoptelov 
I would be more specific here:
`EncodingHash` -> `EncodingBlockHash`

storage/src/lib.rs
```diff
@@ -71,10 +71,13 @@ pub struct BlockHeaderWithHash {
 impl BlockHeaderWithHash {
     /// Create block header extensions from plain block header
     pub fn new(block_header: BlockHeader) -> Result<Self, MessageHashError> {
-        Ok(BlockHeaderWithHash {
-            hash: block_header.message_hash()?.try_into()?,
-            header: Arc::new(block_header),
-        })
+        let hash = if let Some(hash) = block_header.hash().as_ref() {
```
bkontur, line 74: @akoptelov 
what if we put this if to `BlockHeader` directly or thought some trait
```
impl BlockHeader {
 fn block_hash(&self) -> Result<BlockHash, ...> {...}
}
```

bkontur: @akoptelov 
and we would need this also for `pub struct Operation {`,

storage/src/lib.rs
akoptelov: Then we need to redesign messages completely. Not something that can be done as a quick fix.

storage/src/lib.rs
akoptelov: Probably there is even better solution, like a separate node containing data and its hash, but that should be done after I change encoding stuff. If you don't mind, let's keep it as it for now.

tezos/messages/src/p2p/encoding/block_header.rs
akoptelov: It is general enough, probably it will be moved to some common place to be used in other messages.

storage/src/lib.rs
akoptelov: @bkontur The fix is far from perfect, and probably it is even better to postpone it after encoding is redesigned, to do that properly. And now we can just give this changes to @dkuehr so he can continue fuzzing.

akoptelov:
@bkontur The fix is far from perfect, and probably it is even better to postpone it after encoding is redesigned, to do that properly. And now we can just give this changes to @dkuehr so he can continue fuzzing.

bkontur (approved):
@akoptelov 
ok, you know the best encodings now,
please, at least, check fmt, github action for fmt is failed,

and also, please, create child issue here under technical debt https://viablesystems.atlassian.net/browse/TE-574
with all your thought for proper/final/best solution

and then merge to develop

PR #755, opened 2021-08-14
P2p simulator bootstrap test

bkontur:


.drone.yml
```diff
@@ -6532,7 +6532,7 @@ steps:
       - name: tools
         path: /tools
     environment:
-      TROTTLING_QUOTA_DISABLE: false
+      TROTTLING_QUOTA_DISABLE: true
```
binier, line 6763: Typo here :) Should be `TROTTLING_QUOTA_DISABLE`. And in old node run.

bkontur: @binier 
ou, thank you, now it is good everywhere?

PR #757, opened 2021-08-16
[WIP] New encoding [TE-523]

akoptelov:
This is preliminary version for the new encoding.

What is left:
- Remove unused serde stuff
- Sort out errors (now serde error is used)
- Figure out protocol encoding (now it looks like it is not used, and there are some unclear points still)


akoptelov:
@dkuehr you can play with this branch.

tizoc:
Do all those hongfuzz files need to be part of the repository? because it is a lot of files. Or is it just temporary?

bkontur:
> 
> 
> Do all those hongfuzz files need to be part of the repository? because it is a lot of files. Or is it just temporary?

@akoptelov @tizoc 
maybe for code review it could be better do another PR wihtout that thousands of files :)

akoptelov:
> @akoptelov @tizoc 
> maybe for code review it could be better do another PR wihtout that thousands of files :)

Well, it's not for review yet)

tezos/encoding-derive/src/nom.rs
```diff
@@ -115,19 +115,14 @@ fn generate_bytes_nom_read(span: Span) -> TokenStream {
 
 fn generate_struct_nom_read(encoding: &StructEncoding) -> TokenStream {
     let generate_nom_read = match encoding.fields.len() {
-        0 => generate_struct_no_fields_nom_read,
+        0 => unreachable!("No decoding for empty struct"),
```
bkontur, line 118: @akoptelov could this Be triggered in runtime or just in compile tíme?

akoptelov: Compile time only.

tezos/encoding-derive/src/gen.rs
```diff
@@ -0,0 +1,237 @@
...
+    if let Some((inner_encoding, compose_func, prefix)) = decomposed {
+        let compose_func = format_ident!("{}", compose_func);
+        let inner_generator = generator_for_encoding(inner_encoding, prefix);
+        quote!(tezos_encoding::generator::#compose_func(#inner_generator, #generator))
+    } else {
+        generator
+    }
+}
+
+lazy_static::lazy_static! {
```
tizoc, line 119: Is using https://doc.rust-lang.org/std/lazy/struct.Lazy.html instead of `lazy_static` an option here? I have been thinking that we should start switching to that.

tezos/encoding/src/generator.rs
```diff
@@ -0,0 +1,689 @@
...
+    compose(len, item, |len, item| vec![item; len])
+}
+
+/// Helper composer for producing [std::opt::Option] generator out of element generator
+pub fn option<G1, G2>(item: G1, presence: G2) -> impl Generator<Item = Option<G1::Item>>
+where
+    G1: Generator,
+    G2: Generator<Item = bool>,
+{
+    // TODO use optimized version that does not issue [None] multiple times.
```
tizoc, line 536: @akoptelov how important is this TODO? does it deserve an issue on Jira?

tezos/messages/src/p2p/encoding/deactivate.rs
```diff
@@ -5,10 +5,11 @@ use getset::Getters;
 use serde::{Deserialize, Serialize};
 
 use crypto::hash::ChainId;
+use tezos_encoding::enc::BinWriter;
 use tezos_encoding::encoding::HasEncoding;
 use tezos_encoding::nom::NomReader;
 
-#[derive(Serialize, Deserialize, Debug, Getters, Clone, HasEncoding, NomReader)]
+#[derive(Serialize, Deserialize, Debug, Getters, Clone, HasEncoding, NomReader, BinWriter)]
```
tizoc, line 12: @akoptelov do we still need Serialize/Deserialize for all these structs?

tezos/encoding/src/generator.rs
akoptelov: This is mostly for testing purposes. So my answer is no.

tezos/messages/src/p2p/encoding/deactivate.rs
akoptelov: I think we still use it for some messages to get JSON. But let me check that.

tizoc:
@akoptelov not a very thorough review, but other than what I commented one, it looks very good overall.

The dhat tests are failing on drone, what is causing that?

akoptelov:
> The dhat tests are failing on drone, what is causing that?
I've added also each message encoding there, so +2M for that.

tezos/messages/src/p2p/encoding/deactivate.rs
akoptelov: Also it is used in ffi for some of the messages. I've removed `Deserialize` in some cases. Later we'll remove the rest. https://viablesystems.atlassian.net/browse/TE-693.

PR #758, opened 2021-08-17
Document working tree and rename tree to directory

sebastiencs:
- Rename tree to directory
- Rename entry to object
- Add more documentation
- Address https://viablesystems.atlassian.net/browse/TE-663, it uses a bitfield to serialize `Inode::pointers` to save space
- Add `TezedgeIndex::find_node` which returns `None` when the path doesn't exist.
  It address an issue we had where `TezedgeIndex::find_raw_tree` return an empty directory in that case.
- Move `tezos/new_context` to `tezos/context`
- Rename action to query


tizoc:
@sebastiencs cool. I was also thinking about "entry" -> "object". Not sure about "blob", maybe it is ok, but I was considering making it "value" (or even "file" or something like that).

sebastiencs:
@tizoc I think `blob` is a good name: it's a clear difference with the other types (`tree`, `directory`, `blob`).
`value`, in my opinion, is too "generic": Does a `value` contain/include a `tree/directory` ?

Agree for `entry -> object`, I will change that

tizoc:
@sebastiencs "values" are the "leafs" of the tree. The reason I don't like BLOB is that 1) "Large" is part of "_B_**L**_OB_", and what we are storing is not large at all, just a few bytes most of the time 2) we are still serializing that, it would make sense to call all the serialized objects in the repository "blobs", but not what comes before that.

Right now we are not giving any interpretation to values, but that is only because we don't take advantage of the defined schema yet (which makes us double-serialize the data, and include in the serialized format information that is already available in either the schema, or the original serialized format).

- `object` = `directory || value/file/BLOB || commit`
- `tree` = a structure composed of `directory || value/file/BLOB` (even a single value is a "tree" in Irmin's terminology, that is why we use `directory`, to disambiguate that).
- `directories` = nodes of the tree with 1 or more children, "folders" in file-system terms or git (but "directories" too)
- `value/BLOBs/files` = "leafs" of the tree, "files" in git

tizoc:
Regarding tree, all the "tree" operations in the tezos context API work with both values/BLOBs/files and directories.

tizoc:
That one in particular is important because a bunch of the issues found by Patrick's fuzzer were related to that.

tizoc:
@sebastiencs btw, it is probably a good moment for `tezos/new_context` -> `tezos/context`

tezos/context/src/lib.rs
```diff
@@ -60,13 +60,13 @@ pub type ContextValue = Vec<u8>;
 pub type TreeId = i32;
 
 /// Tree in String form needed for JSON RPCs
-pub type StringTreeMap = BTreeMap<String, StringTreeEntry>;
+pub type StringDirectoryMap = BTreeMap<String, StringTreeObject>;
```
tizoc, line 63: Is it this a StringTreeObject or StringDirectoryObject? (I don't even remember for what this code was)

tizoc: What I mean, is, either the left side or the right side is wrong, because one says Tree and the other Directory. I think in this case both should be Tree but I am no 100% sure.

tezos/context/src/lib.rs
```diff
@@ -60,13 +60,13 @@ pub type ContextValue = Vec<u8>;
...
 
 /// Tree in String form needed for JSON RPCs
 #[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(untagged)]
-pub enum StringTreeEntry {
-    Tree(StringTreeMap),
+pub enum StringTreeObject {
```
tizoc, line 68: Ok, seeing this, the answer to the above is that indeed both should be Tree

tezos/context/src/lib.rs
```diff
@@ -60,13 +60,13 @@ pub type ContextValue = Vec<u8>;
...
+    Directory(StringDirectoryMap),
```
tizoc, line 69: Ufff or maybe not :/

tezos/context/src/lib.rs
tizoc: @sebastiencs nevermind, I think the way it is now is correct. It is "directory" because it is the mapping that represents a directory, but the children are Tree's

tezos/context/src/tezedge_context.rs
```diff
@@ -44,11 +44,13 @@ pub struct PatchContextFunction {}
 #[derive(Clone)]
 pub struct TezedgeIndex {
     /// `repository` contains data that were commited and serialized.
```
tizoc, line 46: s/data/objects/ ?

tezos/context/src/tezedge_context.rs
```diff
@@ -70,98 +72,125 @@ impl TezedgeIndex {
...
     }
 
-    pub fn get_entry(&self, hash: HashId, storage: &mut Storage) -> Result<Entry, MerkleError> {
-        match self.find_entry(hash, storage)? {
-            None => Err(MerkleError::EntryNotFound { hash_id: hash }),
-            Some(entry) => Ok(entry),
+    /// Fetches data from the repository and deserialize it into `storage`.
+    ///
+    /// Returns an error when the object was not found.
+    pub fn get_object(&self, hash: HashId, storage: &mut Storage) -> Result<Object, MerkleError> {
```
tizoc, line 114: "Fetches an object from the repository..." (I think we should explicitly say "object" instead of "data" when it is actual "objects" we are working with).

Also, would `fetch_object` be a better name here?

tezos/context/src/tezedge_context.rs
```diff
@@ -70,98 +72,125 @@ impl TezedgeIndex {
...
-    pub fn find_tree(&self, hash: HashId, storage: &mut Storage) -> Result<Option<Tree>, DBError> {
-        match self.find_entry(hash, storage)? {
-            Some(Entry::Tree(tree)) => Ok(Some(tree)),
-            Some(Entry::Blob(_)) => Err(DBError::FoundUnexpectedStructure {
+    /// Fetches the directory associated to this `hash` from the repository.
+    pub fn fetch_directory(
+        &self,
+        hash: HashId,
+        storage: &mut Storage,
+    ) -> Result<Option<DirectoryId>, DBError> {
```
tizoc, line 156: The naming we have is not very consistent, we have at least `fetch_`, `find_` and `get_`, and with different semantics each. I would like to have `find_` for anything that will return an `Option`, `fetch_` for versions that don't return an option and will return an error, and not sure about `get_`, I think the other two should cover it.

tizoc: Commits for example have find + get, and here directory has fetch + get.

tezos/context/src/tezedge_context.rs
```diff
@@ -229,123 +270,132 @@ impl TezedgeIndex {
...
     ///
-    /// # Arguments
-    ///
-    /// * `root` - reference to a tree in which we search
-    /// * `key` - sought path
-    pub fn find_raw_tree(
+    /// Fetches data from the repository if necessary,
+    /// Return an empty directory if no directory under this path exists or if a blob
+    /// (= value) is encountered along the way.
+    pub fn find_directory(
```
tizoc, line 331: This one for example is named `find_` but doesn't return an option.

tezos/context/src/tezedge_context.rs
```diff
@@ -438,29 +489,29 @@ impl TezedgeIndex {
     fn get_key_values_from_tree_recursively(
```
tizoc, line 489: This one is missing documentation. Also, I guess these methods that don't return the value but instead modify an input could use another naming convention that is different from what is used elsewhere. Maybe `collect_` ?

tezos/context/src/working_tree/serializer.rs
```diff
@@ -250,6 +252,85 @@ fn serialize_hash_id(hash_id: u32, output: &mut Vec<u8>) -> Result<usize, Serial
     }
 }
 
+#[derive(Copy, Clone, Default, Debug)]
```
tizoc, line 255: Please add a comment here introducing what PointerDescriptor is

tezos/context/src/working_tree/serializer.rs
```diff
@@ -650,27 +718,25 @@ impl<'a> Iterator for HashIdIterator<'a> {
...
 
-                    // Entry::Commit.root_hash
+                    // Object::Commit.root_hash
                     let (root_hash, _) = deserialize_hash_id(self.data.get(1 + nbytes..)?).ok()?;
                     self.pos = self.data.len();
 
                     return root_hash;
                 } else if id == ID_INODE_POINTERS {
-                    pos += 10;
+                    pos += 13;
```
tizoc, line 733: why 13? can we give this a named constant + documentation?

tezos/context/src/working_tree/storage.rs
```diff
@@ -36,9 +36,16 @@ const FULL_31_BITS: usize = 0x7FFFFFFF;
 const FULL_28_BITS: usize = 0xFFFFFFF;
 const FULL_4_BITS: usize = 0xF;
 
+/// Length of a blob we consider inlined.
+///
+/// Do not consider blobs of length zero as inlined, this never
+/// happens when the node is running and fix a serialization issue
+/// during testing/fuzzing
+const BLOB_INLINED_RANGE: Range<usize> = 1..8;
```
tizoc, line 44: Maybe this will be more obvious if expressed as an inclusive range instead.

tezos/context/src/working_tree/storage.rs
```diff
@@ -422,42 +429,42 @@ pub enum Inode {
...
     /// Concatenation of all strings in the working tree.
     /// The working tree has `StringId` which refers to a data inside `StringInterner`.
     strings: StringInterner,
     /// Concatenation of all inodes.
     /// Note that the implementation of `Storage` attempt to hide as much as
     /// possible the existence of inodes to the working tree.
-    /// The working tree doesn't manipulate `InodeId` but `TreeStorageId` only.
-    /// A `TreeStorageId` might contains an `INodeId` but it's only the root
+    /// The working tree doesn't manipulate `InodeId` but `DirectoryId` only.
+    /// A `DirectoryId` might contains an `INodeId` but it's only the root
```
tizoc, line 467: s/INodeId/InodeId/

tezos/context/src/working_tree/storage.rs
```diff
@@ -616,14 +622,17 @@ impl Storage {
...
-    fn binary_search_in_tree(
+    /// Search `key` in the directory and return the result of `slice::binary_search_by`.
+    ///
+    /// `dir` must be sorted.
+    fn binary_search_in_dir(
         &self,
-        tree: &[(StringId, NodeId)],
+        dir: &[(StringId, NodeId)],
         key: &str,
     ) -> Result<Result<usize, usize>, StorageError> {
```
tizoc, line 632: This result type is very weird, please add some more text in the comment explaining it.

tezos/context/src/working_tree/storage.rs
```diff
@@ -1054,60 +1063,60 @@ impl Storage {
             .ok_or(StorageError::InodeNotFound)
     }
 
-    pub fn tree_insert(
+    pub fn dir_insert(
```
tizoc, line 1066: Missing docs

tezos/context/src/working_tree/working_tree.rs
```diff
@@ -11,29 +11,29 @@
 //! A storage with just one key `a/b/c` and its corresponding value `8` is represented like this:
 //!
 //! ``
-//! [commit] ---> [tree1] --a--> [tree2] --b--> [tree3] --c--> [blob_8]
+//! [commit] ---> [dir1] --a--> [dir2] --b--> [dir3] --c--> [blob_8]
```
tizoc, line 14: As mentioned in the above TODO, all this is quite out of date with the current implementation. It is not just about renaming, it needs to be adjusted (or replaced) so that it is still accurate.

tizoc (changes requested):
@sebastiencs looks good overall but I left some comments regarding naming conventions, etc

tezos/context/src/working_tree/working_tree.rs
```diff
@@ -712,108 +712,12 @@ impl WorkingTree {
...
         let mut storage = self.index.storage.borrow_mut();
 
-        match self.index.find_node(dir_id, key, &mut storage) {
+        match self.index.go_to_node(dir_id, key, &mut storage) {
             Ok(node_id) => node_id.is_some(),
             Err(_) => false,
         }
     }
 
-    // TODO: recursion is risky (stack overflow) and inefficient, try to do it iteratively..
```
tizoc, line 718: @sebastiencs not used anymore? (I was wondering why we had what seemed to be duplicate code here and in the context)

sebastiencs: @tizoc Yes it's a duplicate.
I'm not sure but at some point we moved some methods from the `WorkingTree` into `TezedgeIndex`, and I think we didn't remove some from the `WorkingTree`

sebastiencs:
@tizoc Your comments were addressed

tizoc (approved):
@sebastiencs ok cool. Please squash the commits that are related to documentation into one then the other refactors grouped as they make sense (I guess the find_node change + the inode pointers optimization also on it's own commit) and rebase over develop, then it can be merged.

PR #761, opened 2021-08-19
Improve timing channel

sebastiencs:
- Remove allocations when sending timing messages

### DIFF
--- a/tezos/context/src/timings.rs
+++ b/tezos/context/src/timings.rs
@@ -5,7 +5,7 @@ use std::time::{Duration, Instant, SystemTime, UNIX_EPOCH};
 
 use ocaml_interop::*;
 use tezos_api::ocaml_conv::{OCamlBlockHash, OCamlContextHash, OCamlOperationHash, from_ocaml::hash_as_bytes};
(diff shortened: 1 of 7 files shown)


tezos/context/src/timings.rs
```diff
@@ -80,32 +105,95 @@ pub fn commit(
     }
 }
 
+fn get_query_kind(name: &[u8]) -> Option<QueryKind> {
```
tizoc, line 108: @sebastiencs did this really make any improvement? I didn't consider that mapping from string to enum to be expensive

sebastiencs: @tizoc It makes quite a difference yes. Could we use integers (from ocaml to rust) instead of strings to determine the query name ?

tizoc: Yes, we can, or an enum. I assumed the compiler was going to be able to do something for such small strings, but I see it is not the case.

bkontur: @tizoc @sebastiencs 
I would prefer to do it like enum right now :), it is a part of ffi interface, which should be clear and exact instead of that "first byte and length" comparision like this:
```
 b'f' => {
            if name.len() == 3 {
                Some(QueryKind::Find)
            } else {
                Some(QueryKind::FindTree)
            }
        }
```

tizoc:
Closing this one, rebased and merged in https://github.com/tezedge/tezedge/pull/838

Also switched to using enums to represent the query types.

PR #766, opened 2021-08-25
Rolls owner snapshots for RPC

tizoc:
What is in here:

- At the end of each cycle, the new frozen rolls owner snapshots are returned as part of the apply block response
- When there is a protocol switch, the new constants (and cycle eras data) are returned (json-encoded) as part of the apply block response
- This data is stored by the shell, which makes it available without having to query the context.
- `endorsing_rights` and `baking_rights` RPCs now use that data to serve requests without having to query the context.


storage/src/lib.rs
```diff
@@ -477,6 +511,9 @@ pub mod initializer {
                 crate::ChainMetaStorage::descriptor(cache),
                 crate::PredecessorStorage::descriptor(cache),
                 crate::BlockAdditionalData::descriptor(&cache),
+                crate::CycleMetaStorage::descriptor(cache),
```
bkontur, line 514: @adonagy 
check also `impl TmpStorage {` and add there these storages,
this why interation tests are failing

PR #768, opened 2021-08-26
Directory shape optimization

sebastiencs:


sebastiencs:
@tizoc The PR is now ready for review.

- To recognize a shape, it hashes the `StringId` now.

tezos/context/src/tezedge_context.rs
```diff
@@ -558,6 +565,14 @@ impl TezedgeIndex {
             },
         }
     }
+
+    fn clone_string_interner_to_repo(&self) -> Result<(), MerkleError> {
```
tizoc, line 569: I see this being used in a bunch of places, but I don't fully understand when it needs to run and why it is necessary. Needs a comment.

tizoc: What confuses me the most is that I see it being called in methods that just read from the context. I would imagine that you only need to replace it on commits in cases where the context is modified in a way that introduces new shapes and strings, but that doesn't happen on reads.

tezos/context/src/working_tree/serializer.rs
```diff
@@ -471,6 +535,14 @@ impl From<StorageError> for DeserializationError {
     }
 }
 
+impl From<DBError> for DeserializationError {
```
tizoc, line 538: If possible, lets start using the `#[from]` annotation now that the `thiserror` [PR has been merged](https://github.com/tezedge/tezedge/pull/605). Why is the `Box` needed?

tezos/context/src/working_tree/serializer.rs
```diff
@@ -497,6 +569,79 @@ fn deserialize_hash_id(data: &[u8]) -> Result<(Option<HashId>, usize), Deseriali
     }
 }
 
+fn deserialize_shaped_directory(
+    data: &[u8],
+    storage: &mut Storage,
+    store: &ContextKeyValueStore,
```
tizoc, line 575: `store` + `storage` is super confusing, we need to figure out a better name for `storage`.

tezos/context/src/working_tree/shape.rs
```diff
@@ -0,0 +1,137 @@
...
+
+/// Contains the shape (key fragments) of a directory.
+///
+/// A `ShapeId` maps to a slice of `StringId`
+pub struct Shapes {
+    /// Map `ShapeHash` to its `ShapeId` and strings
+    hash_to_strings: BTreeMap<ShapeHash, (ShapeId, Box<[StringId]>)>,
+    /// Map the `ShapeId` to its `ShapeHash`.
+    id_to_hash: IndexMap<ShapeId, ShapeHash>,
+    temp: Vec<StringId>,
```
tizoc, line 69: The name of this one is not very descriptive, it needs a comment.

tezos/context/src/working_tree/serializer.rs
tizoc: For `ContextKeyValueStore` we use `repo` or `repository` elsewhere, I think we should at least use that elsewhere to make things less ambiguous.

tizoc: And for the two cases where I see `store` being used directly now (instead of just passed around), I see it is always about obtaining or creating a new shape. I think it would be better if here we passed around a wrapper that is specifically about shapes, so it would be just `shapes` instead and also `shapes.get(id)` and `shapes.create(dir)`, what do you think?

tizoc (changes requested):
@sebastiencs I need to do more passes over it, but I left some feedback for now. In general I think it needs more comments with clarifications.

tizoc:
@sebastiencs one more thing, I propose `shape -> directory_shape / dir_shape`, it is less ambiguous and easier to understand (or if you have a better name, suggestions welcome).

tezos/context/src/tezedge_context.rs
sebastiencs: @tizoc You are right, I only called `clone_string_interner_to_repo` on `commit` and all the tests passed: http://ci.tezedge.com/tezedge/tezedge/2417

tezos/context/src/working_tree/serializer.rs
sebastiencs: @tizoc The `Box` is required because it would make an infinite recursion between `DeserializationError` and `DBError`.
It seems that `#[from]` doesn't handle this case

tezos/context/src/working_tree/shape.rs
```diff
@@ -17,59 +17,62 @@ use serde::{Deserialize, Serialize};
...
+struct DirectoryShapeHash(u64);
 
 /// Contains the shape (key fragments) of a directory.
 ///
 /// A `ShapeId` maps to a slice of `StringId`
-pub struct Shapes {
-    /// Map `ShapeHash` to its `ShapeId` and strings
-    hash_to_strings: BTreeMap<ShapeHash, (ShapeId, Box<[StringId]>)>,
+pub struct DirectoryShapes {
+    /// Map `ShapeHash` to its `ShapeId` and strings.
```
tizoc, line 67: The comments still have the old name

tezos/context/src/kv_store/in_memory.rs
```diff
@@ -209,7 +219,36 @@ impl KeyValueStoreBackend for InMemory {
...
+        &mut self,
+        dir: &[(StringId, DirEntryId)],
+    ) -> Result<Option<DirectoryShapeId>, DBError> {
+        self.shapes.make_shape(dir).map_err(Into::into)
+    }
+
+    fn update_strings(&mut self, string_interner: &StringInterner) -> Result<(), DBError> {
+        // Clone it only if it is has changed
+        if string_interner != &self.string_interner {
+            self.string_interner = string_interner.clone();
```
tizoc, line 244: @sebastiencs how about the `push_str` here to avoid the clone, can that be done? And also just compare the length instead? And then the slice resulting from the difference between the two lengths is pushed.

tezos/context/src/tezedge_context.rs
tizoc: @sebastiencs ok great. Now that there is no actual cloning one this needs a better name. `update_repository_string_interner` or `syncronize_string_interner_to_repository` or something like that. Also please add a comment explaining what this is doing and why it is needed.

sebastiencs: @tizoc Done

tizoc (approved):
@sebastiencs ok looks good, but I left a few more comments regarding error types.

tezos/context/src/persistent/mod.rs
```diff
@@ -102,6 +126,12 @@ impl From<HashIdError> for DBError {
     }
 }
 
+impl From<DirectoryShapeError> for DBError {
```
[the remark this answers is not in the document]
sebastiencs: @tizoc Done

tezos/context/src/working_tree/serializer.rs
```diff
@@ -439,6 +501,8 @@ pub enum DeserializationError {
     StorageIdError { error: StorageError },
     InodeNotFoundInRepository,
     InodeEmptyInRepository,
+    DBError { error: Box<DBError> },
```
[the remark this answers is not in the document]
sebastiencs: @tizoc done, I've added the `Error` implementation on the error types

tezos/context/src/working_tree/storage.rs
```diff
@@ -328,9 +335,15 @@ impl BlobId {
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub struct DirEntryId(u32);
 
-#[derive(Debug)]
+#[derive(Debug, Error)]
 pub struct DirEntryIdError;
 
+impl std::fmt::Display for DirEntryIdError {
```
tizoc, line 341: I think you can just add `#[error("Message")]` to the whole struct and it should work, example: https://github.com/dtolnay/thiserror/blob/master/tests/test_from.rs#L6-L11

tezos/context/src/working_tree/storage.rs
```diff
@@ -204,6 +205,12 @@ pub enum StorageError {
     RootOfInodeNotAPointer,
 }
 
+impl std::fmt::Display for StorageError {
```
tizoc, line 208: Please add error annotations with text to each enum entry instead

tezos/context/src/working_tree/serializer.rs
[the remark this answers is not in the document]
tizoc: Cool, much better.

tezos/context/src/working_tree/storage.rs
sebastiencs: @tizoc Thanks

tezos/context/src/working_tree/storage.rs
sebastiencs: @tizoc Done

tizoc (approved):
@sebastiencs ok, looks good. Please squash the commits and if for some reason drone fails again because of that intermittent failure preventing you from merging I will do it.

PR #775, opened 2021-08-31
Fix `WorkingTree::hash` when the root is a value

sebastiencs:
`WorkingTree::hash` on a root value was behaving like if the root
was an empty directory, but it should return the hash of the
value (blob) instead.


tizoc (approved):
@sebastiencs looks good, but please add a simple test case for this in the tests for `lib_storage_ffi` (set a value for both irmin and tezedge contexts and compare the hash).

tizoc (approved):

PR #784, opened 2021-09-14
Asynchronous IPC

tizoc:
Also reorganizes code to get rid of any dependency from the shell process to libtezos.


tezos/context-api/src/lib.rs
```diff
@@ -0,0 +1,257 @@
...
+        }
+    }
+}
+
+// Must be in sync with ffi_config.ml
+#[derive(Serialize, Deserialize, Debug)]
+pub struct TezosContextConfiguration {
+    pub storage: TezosContextStorageConfiguration,
+    pub genesis: GenesisChain,
+    pub protocol_overrides: ProtocolOverrides,
```
bkontur, line 141: @tizoc 
nice cleanup :),
technically, `ProtocolOverrides` and `enable_testchain` are not related to context, they are just used with `apply_block` function, so maybe "sometime" we should move it from `TezosContextConfiguration` and `context-api`

tizoc: @bkontur yes. Also, while this solves the shell -> libtezos dependency, something that is still pending is the full split between our context storage and everything else so that it can be packaged as an opam package.

tizoc: Btw, if you spot anything else please point it out. For now I am leaving everything as-is, because the focus was on properly splitting things, but now may be a good time to start caring about this kind of stuff too while I work on the IPC messages.

PR #793, opened 2021-09-29
Optimize monitoring RPCs 

adonagy:
Includes optimization for the monitoring RPCs. 

Still some stuff left to do:

- [ ] Fix leftover TODOs
- [ ] cleanup code

### DIFF
--- a/rpc/src/server/mod.rs
+++ b/rpc/src/server/mod.rs
@@ -15,6 +15,8 @@ use hyper::service::{make_service_fn, service_fn};
 use hyper::{Body, Method, Request, Response};
 use slog::{error, Logger};
 use tokio::runtime::Handle;
+use uuid::Uuid;
+use futures::task::Waker;
 
 use crypto::hash::ChainId;
 use shell::mempool::CurrentMempoolStateStorageRef;
@@ -25,6 +27,7 @@ use tezos_messages::p2p::encoding::version::NetworkVersion;
 use tezos_wrapper::TezedgeContextClient;
 use tezos_wrapper::TezosApiConnectionPool;
 use url::Url;
+use shell::state::streaming_state::StreamCounter;
 
 use crate::{error_with_message, not_found, options};
 
@@ -47,6 +50,19 @@ pub type RpcServiceEnvironmentRef = Arc<RpcServiceEnvironment>;
 pub struct RpcCollectedState {
     #[get = "pub(crate)"]
     current_head: Arc<BlockHeaderWithHash>,
+
+    // Wakers for open streams (monitors) that access the mempool state
+    streams: HashMap<Uuid, Waker>
+}
+
+impl StreamCounter for RpcCollectedState {
+    fn get_streams(&self) -> HashMap<Uuid, Waker> {
+        self.streams.clone()
+    }
+
+    fn get_mutable_streams(&mut self) -> &mut HashMap<Uuid, Waker> {
+        &mut self.streams
+    }
 }
 
 /// Server environment parameters
--- a/rpc/src/server/mod.rs
+++ b/rpc/src/server/mod.rs
@@ -10,16 +10,17 @@ use std::{
     path::PathBuf,
 };
 
+use futures::task::Waker;
 use getset::{CopyGetters, Getters};
 use hyper::service::{make_service_fn, service_fn};
 use hyper::{Body, Method, Request, Response};
 use slog::{error, Logger};
 use tokio::runtime::Handle;
 use uuid::Uuid;
-use futures::task::Waker;
 
 use crypto::hash::ChainId;
 use shell::mempool::CurrentMempoolStateStorageRef;
+use shell::state::streaming_state::StreamCounter;
 use shell_integration::ShellConnectorRef;
 use storage::{BlockHeaderWithHash, PersistentStorage};
 use tezos_api::environment::TezosEnvironmentConfiguration;
@@ -27,7 +28,6 @@ use tezos_messages::p2p::encoding::version::NetworkVersion;
 use tezos_wrapper::TezedgeContextClient;
 use tezos_wrapper::TezosApiConnectionPool;
 use url::Url;
-use shell::state::streaming_state::StreamCounter;
 
 use crate::{error_with_message, not_found, options};
 
@@ -52,7 +52,7 @@ pub struct RpcCollectedState {
     current_head: Arc<BlockHeaderWithHash>,
 
     // Wakers for open streams (monitors) that access the mempool state
-    streams: HashMap<Uuid, Waker>
+    streams: HashMap<Uuid, Waker>,
 }
 
 impl StreamCounter for RpcCollectedState {
--- a/rpc/src/services/stream_services.rs
+++ b/rpc/src/services/stream_services.rs
@@ -16,9 +16,9 @@ use uuid::Uuid;
 
 use crypto::hash::{BlockHash, ChainId, OperationHash, ProtocolHash};
 use shell::mempool::CurrentMempoolStateStorageRef;
+use shell::state::streaming_state::StreamCounter;
 use storage::{BlockHeaderWithHash, BlockMetaStorage, BlockMetaStorageReader, PersistentStorage};
 use tezos_messages::{ts_to_rfc3339, TimestampOutOfRangeError};
-use shell::state::streaming_state::StreamCounter;
 
 use crate::server::RpcCollectedStateRef;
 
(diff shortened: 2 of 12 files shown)


shell/src/mempool/mempool_state.rs
```diff
@@ -224,6 +243,18 @@ impl MempoolState {
     pub fn operations(&self) -> &HashMap<OperationHash, Operation> {
         &self.operations
     }
+
+    // pub fn add_stream(&mut self, id: Uuid, waker: Waker) {
```
bkontur, line 247: @adonagy 
do we need this?
and please, check easy clippy stuff with `&` in this file

shell/src/state/streaming_state.rs
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) SimpleStaking, Viable Systems and Tezedge Contributors
+// SPDX-License-Identifier: MIT
+
+use std::collections::HashMap;
+
+use futures::task::Waker;
+use uuid::Uuid;
+
+pub trait StreamCounter {
```
bkontur, line 9: @adonagy 
I would better move this to the `shell-integration` module, because in the end, I want to remove `shell` dependency from `rpc/Cargo.toml`

rpc/src/services/stream_services.rs
```diff
@@ -70,163 +69,230 @@ pub struct MempoolOperationsQuery {
...
+            stream_id,
+            poll_counter: 0,
         }
     }
 
     fn yield_operations(&mut self) -> Poll<Option<Result<String, anyhow::Error>>> {
         let OperationMonitorStream {
-            chain_id,
             current_mempool_state_storage,
             log,
```
bkontur, line 220: @adonagy 
`log` is unused variable

rpc/src/services/stream_services.rs
```diff
@@ -291,105 +359,138 @@ impl Stream for HeadMonitorStream {
...
+            head_hash
+        } else {
+            // first poll
+            self.last_checked_head = Some(current_head.hash.clone());
+            // If there is no head with the desired protocol, [yield_head] returns Ok(None) which is transposed to None, meaning we
+            // would end the stream, in this case, we need to Pend.
+            if let Some(head_string_result) = self.yield_head(&current_head).transpose() {
+                return Poll::Ready(Some(head_string_result));
+            } else {
+                // cx.waker().wake_by_ref();
```
bkontur, line 396: @adonagy 
please, remove better all unneeded code

rpc/src/server/mod.rs
```diff
@@ -47,6 +50,19 @@ pub type RpcServiceEnvironmentRef = Arc<RpcServiceEnvironment>;
 pub struct RpcCollectedState {
     #[get = "pub(crate)"]
     current_head: Arc<BlockHeaderWithHash>,
+
+    // Wakers for open streams (monitors) that access the mempool state
+    streams: HashMap<Uuid, Waker>,
```
bkontur, line 55: @adonagy 
I would prefer to spread that hacked `uuid` as less as possible, so 
I would better clean this little:
- move `streaming_state` to `shell-integration`
- create alias type in `streaming_state`, e.g.
```
pub type StreamWakers = HashMap<Uuid, Waker>;
```
- remove `uuid` from `rpc/Cargo.toml` and `shell/Cargo.toml`
- add that hacked `uuid` just to `shell-integration/Cargo.toml`

bkontur: anyway, if this will be fixed: https://github.com/serde-rs/serde/issues/2096
we could return to the normal `uuid` dependency

bkontur: @adonagy 
and also if we want to add here that hacked `uuid`, please change it everywhere, I see it in `monitoring` module,
that we use just the same `uuid` everywhere

shell/src/state/streaming_state.rs
adonagy: Ok, will do

shell-integration/src/streaming_state.rs
```diff
@@ -0,0 +1,30 @@
...
+use std::collections::HashMap;
+
+use futures::task::Waker;
+use uuid::Uuid;
+
+pub type StreamWakers = HashMap<Uuid, Waker>;
+
+pub trait StreamCounter {
+    fn get_mutable_streams(&mut self) -> &mut StreamWakers;
+    fn get_streams(&self) -> StreamWakers;
```
bkontur, line 13: @adonagy 
if you return `& StreamWakers`, does it work?
I guess, that cloning is unnecessery

here the iter takes just reference, right?

```
   self.get_streams().iter().for_each(|(_, waker)| waker.wake_by_ref())
```

rpc/src/services/stream_services.rs
```diff
@@ -70,163 +69,232 @@ pub struct MempoolOperationsQuery {
...
-    chain_id: ChainId,
+    _chain_id: ChainId,
     current_mempool_state_storage: CurrentMempoolStateStorageRef,
     state: RpcCollectedStateRef,
     last_checked_head: BlockHash,
     log: Logger,
-    delay: Option<Interval>,
-    streamed_operations: Option<HashSet<String>>,
+    contains_waker: bool,
+    stream_id: Uuid,
```
bkontur, line 190: @adonagy 
well, I would maybe do the wrapper here, so you dont need to spread `uuid` dependecy to rpc module :)
```
pub type StreamId = Uuid;

pub fn generate_stream_id() -> StreamId {
  Uuid::new_v4()
}

pub type StreamWakers = HashMap<StreamId, Waker>;
```

or maybe some wrapper, but then you need to reimpolement Eq,Hash,PartialEq...
```
pub struct StreamId(Uuid) {
   
}

impl StreamId {
   
pub fn generate() -> Self {..}
}

```

I guess the first version is enought

shell-integration/src/streaming_state.rs
adonagy: Yep, that's a very good point :) Adjusted accordingly

rpc/src/services/stream_services.rs
adonagy: I've gone with the your first suggestion. The trait would have been nice, but we would have to either:

1. lock the rwlock on the state that's implementing the trait
2. use a trait object to call `generate_stream_id` directly, this would add a dynamic dispatch the runtime (and could cause a small performance hit maybe)

Both cases add unnecessary complexity, so I went with the most simple wrapper approach :)

rpc/src/services/stream_services.rs
```diff
@@ -70,163 +68,232 @@ pub struct MempoolOperationsQuery {
...
             current_mempool_state_storage,
             state,
             last_checked_head,
             log,
-            delay: None,
+            contains_waker: false,
             query: mempool_operaions_query,
-            streamed_operations: None,
+            streamed_operations: HashSet::new(),
+            stream_id,
```
bkontur, line 214: @adonagy 
why not just:
```
stream_id: generate_stream_id(),
```

rpc/src/services/stream_services.rs
```diff
@@ -236,20 +303,24 @@ impl HeadMonitorStream {
...
     ) -> Self {
+        let stream_id = generate_stream_id();
         Self {
             state,
             protocol,
             last_checked_head: None,
-            delay: None,
             block_meta_storage: BlockMetaStorage::new(persistent_storage),
+            contains_waker: false,
+            stream_id,
```
bkontur, line 315: @adonagy 
also here:
```
stream_id: generate_stream_id(),
```

shell-integration/src/streaming_state.rs
```diff
@@ -0,0 +1,35 @@
...
+use uuid::Uuid;
+
+pub type StreamWakers = HashMap<Uuid, Waker>;
+pub type StreamId = Uuid;
+
+pub trait StreamCounter {
+    fn get_mutable_streams(&mut self) -> &mut StreamWakers;
+    fn get_streams(&self) -> &StreamWakers;
+
+    fn add_stream(&mut self, id: Uuid, waker: Waker) {
```
bkontur, line 16: @adonagy 
also, please adjust api `Uuid` -> `StreamId`

shell-integration/src/streaming_state.rs
```diff
@@ -0,0 +1,35 @@
+// Copyright (c) SimpleStaking, Viable Systems and Tezedge Contributors
+// SPDX-License-Identifier: MIT
+
+use std::collections::HashMap;
+
+use futures::task::Waker;
+use uuid::Uuid;
+
+pub type StreamWakers = HashMap<Uuid, Waker>;
```
bkontur, line 9: @adonagy 
:D :D :D last one here: `Uuid -> StreamId`

bkontur (approved):
@adonagy 
please, check that last one :D

shell-integration/src/streaming_state.rs
adonagy: ... 😄

bkontur: nice :)

adonagy:
@bkontur Thank you for the review! Will squash and merge after CI passes

PR #794, opened 2021-09-30
Support for concurrent checkouts in the context API

tizoc:


tezos/context/src/tezedge_context.rs
```diff
@@ -612,23 +629,28 @@ impl IndexApi<TezedgeContext> for TezedgeIndex {
...
+                None => return Ok(None),
+            };
+
+            match self.fetch_directory(commit.root_hash, &mut storage)? {
+                Some(dir_id) => dir_id,
+                None => return Ok(None),
+            }
         };
 
         let tree = WorkingTree::new_with_directory(self.clone(), dir_id);
```
sebastiencs, line 650:
```suggestion
        let tree = WorkingTree::new_with_directory(index.clone(), dir_id);
```
This needs to use the newly created index

tizoc: Oh, good catch

tizoc:
@sebastiencs so, it is not the most optimal way to do things because it clones the interner on each checkout. I wanted to separate between "readonly" and "for-commit" checkouts, but that would not work for Octez because it will not differentiate.

sebastiencs (approved):

PR #801, opened 2021-10-06
Implement persistent context

sebastiencs:


tezos/context/src/serialize/persistent.rs
```diff
@@ -0,0 +1,1799 @@
...
+    let mut pointers: [Option<PointerToInode>; 32] = Default::default();
+
+    for (index, pointer_index) in indexes_iter.enumerate() {
+        let offset_length = offsets_header.get(index);
+        let (absolute_offset, nbytes) =
+            deserialize_offset(&data[pos..], offset_length, object_offset)?;
+
+        pos += nbytes;
+
+        // TODO: Move this outside the loop
```
tizoc, line 1199: TODO here

tezos/context/src/tezedge_context.rs
```diff
@@ -574,35 +632,110 @@ impl TezedgeIndex {
     /// - The repository needs those interned strings when it sends the directory shapes to the
     ///   read only protocol runner.
     fn synchronize_interned_strings_to_repository(&self) -> Result<(), MerkleError> {
-        let storage = self.storage.borrow();
+        let mut strings = self.string_interner.borrow_mut();
         let mut repository = self.repository.write()?;
-        repository.synchronize_strings_from(&storage.strings);
+        repository.synchronize_strings_from(&strings);
+
+        // TODO: Don't do this hack
```
tizoc, line 639: What is the hack about?

tezos/context/src/tezedge_context.rs
```diff
@@ -746,7 +836,7 @@ impl IndexApi<TezedgeContext> for TezedgeIndex {
 pub struct TezedgeContext {
     /// Index used for fetching and saving objects from/to the repository.
     pub index: TezedgeIndex,
-    pub parent_commit_hash: Option<HashId>,
+    pub parent_commit_ref: Option<ObjectReference>,
```
tizoc, line 839: Maybe it is ok this way, but in my mind the reference would be just the hash (we have an index that maps hashes to commit objects).

tezos/context/src/working_tree/mod.rs
```diff
@@ -77,6 +88,41 @@ pub enum Object {
     Commit(Box<Commit>),
 }
 
+#[derive(Debug, Copy, Clone, Serialize, Deserialize, Hash, PartialEq, Eq)]
+pub struct ObjectReference {
```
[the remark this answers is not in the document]
tizoc: How about `KeyValueStoreBackend` having a type parameter? it is there that `ObjectReference` is used right? So the inmem implementation would have `HashId` and the ondisk implementation `AbsoluteOffset`. `TezedgeContext`, because it contains the repo inside will probably need to be parametrized over that type too.

tezos/context/src/working_tree/storage.rs
```diff
@@ -485,14 +513,19 @@ pub struct Storage {
     blobs: Vec<u8>,
     /// Concatenation of all strings in the working tree.
     /// The working tree has `StringId` which refers to a data inside `StringInterner`.
-    pub strings: StringInterner,
+    // pub strings: Option<&'a mut StringInterner>,
```
tizoc, line 516: this is commented out here, not used anymore? (iirc it was not part of the index or repository), if so lets remove it

tezos/context/src/working_tree/storage.rs
```diff
@@ -547,13 +581,12 @@ impl Storage {
         let blobs_cap = self.blobs.capacity();
         let temp_dir_cap = self.temp_dir.capacity();
         let inodes_cap = self.inodes.capacity();
-        let strings = self.strings.memory_usage();
+        let strings = Default::default();
```
tizoc, line 584: this will not result in the actual stats right?

tezos/context/src/working_tree/storage.rs
```diff
@@ -1351,7 +1403,7 @@ impl Storage {
     }
 
     pub fn clear(&mut self) {
-        self.strings.clear();
+        // self.strings.clear();
```
tizoc, line 1406: leftover commented out code

tezos/conv/src/to_ocaml.rs
```diff
@@ -108,6 +108,7 @@ impl_to_ocaml_variant! {
     ContextKvStoreConfiguration => OCamlContextKvStoreConfiguration {
         ContextKvStoreConfiguration::ReadOnlyIpc,
         ContextKvStoreConfiguration::InMem,
+        ContextKvStoreConfiguration::OnDisk,
```
tizoc, line 111: Oh, I need to add this to the OCaml side right? I guess that is why `InMem` points to the persistent implementation right now. I will add it tomorrow.

tezos/context/src/working_tree/string_interner.rs
```diff
@@ -84,16 +125,100 @@ impl BigStrings {
     }
 
     fn clear(&mut self) {
-        let cap = self.strings.capacity();
+        // let cap = self.strings.capacity();
```
tizoc, line 128: A bunch of commented out code here and in the next methods. Should not be called anymore or needs to be reimplemented?

tezos/context/src/working_tree/working_tree.rs
```diff
@@ -736,41 +792,59 @@ impl WorkingTree {
...
-        parent_commit_hash: Option<HashId>,
-        store: &mut ContextKeyValueStore,
-        commit_to_storage: bool,
+        parent_commit_ref: Option<ObjectReference>,
+        repository: &mut ContextKeyValueStore,
+        serialize_function: Option<SerializeObjectSignature>,
+        offset: Option<AbsoluteOffset>,
     ) -> Result<PostCommitData, MerkleError> {
-        let root_hash = self.get_root_directory_hash(store)?;
+        // {
```
tizoc, line 800: More commented out code here

tizoc (commented):
Overall it looks good, but in the future once it stabilizes we have to figure out how to start simplifying the implementation, feels a bit daunting right now. One obvious place that I think could be simplified quite a bit is everything related to serialization and deserialization, I think all that manual code can probably be replaced by a declarative implementation that performs as well.

tezos/context/src/initializer.rs
```diff
@@ -49,7 +56,9 @@ pub fn initialize_tezedge_index(
                     )),
                 }
             }
-            ContextKvStoreConfiguration::InMem => Arc::new(RwLock::new(InMemory::try_new()?)),
+            ContextKvStoreConfiguration::InMem => Arc::new(RwLock::new(Persistent::try_new()?)),
```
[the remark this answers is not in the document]
tizoc: Something that is missing that I need to fix is handling `OnDisk` on the OCaml side

tezos/context/src/kv_store/readonly_ipc.rs
```diff
@@ -136,6 +137,94 @@ impl KeyValueStoreBackend for ReadonlyIpcBackend {
...
+            buffer.extend_from_slice(bytes);
+            return Ok(buffer);
+        };
+
+        self.client
+            .get_object_bytes(object_ref, buffer)
+            .map_err(|reason| DBError::IpcAccessError { reason })
+    }
+
+    fn commit(
```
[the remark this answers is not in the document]
sebastiencs: I might be wrong but I thought `commit`, on the readonly-ipc backend, was used to produce the hash ?
This `ReadonlyIpcBackend::commit` doesn't write anything to storage.
See `WorkingTree::prepare_commit`, it takes a parameter `serialize_function` which is `None` in that case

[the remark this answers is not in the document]
tizoc: @sebastiencs I don't think this will ever be called (readonly protocol runners will not call commit). In any case, it doesn't hurt for it to work like this, but please add a comment clarifying that readonly-ipc never really uses commit to make that clear.

tezos/context/src/kv_store/persistent.rs
```diff
@@ -0,0 +1,484 @@
...
+    fn block_applied(
+        &mut self,
+        _referenced_older_objects: Vec<HashId>,
+    ) -> Result<(), GarbageCollectionError> {
+        Ok(())
+    }
+}
+
+impl Flushable for Persistent {
+    fn flush(&self) -> Result<(), anyhow::Error> {
```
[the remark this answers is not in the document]
tizoc: @sebastiencs I see this is not used anywhere, how about adding the fsync calls here and always call it on commit for now? we can then change the logic to separate bootstrap/import workloads to call flush less often without it being specific to this backend.

tezos/context/src/kv_store/persistent.rs
```diff
@@ -0,0 +1,532 @@
...
+                Some(offset),
+            )
+            .map_err(Box::new)?;
+
+        let commit_hash = get_commit_hash(commit_ref, self).map_err(Box::new)?;
+
+        self.commit_to_disk(&output)
+            .map_err(|err| DBError::CommitToDiskError { err })?;
+
+        self.put_context_hash(commit_ref)?;
```
tizoc, line 511: @sebastiencs  this here is happening after all the previous calls to `fsync()` right? cannot we lose that data then?

tezos/context/src/working_tree/working_tree.rs
```diff
@@ -898,80 +969,97 @@ impl WorkingTree {
...
+                    let object_hash_id =
+                        match dir_entry.object_hash_id(data.repository, storage, strings)? {
+                            Some(hash_id) => hash_id,
+                            None => return Ok(()), // Object is an inlined blob, we don't serialize them.
+                        };
 
-                    if child_dir_entry.is_commited() {
-                        data.add_older_object(child_dir_entry, storage)?;
+                    if dir_entry.is_commited() {
+                        data.add_older_object(dir_entry, storage, strings)?; // TODO: Don't run this line on persistent context
```
tizoc, line 1003: @sebastiencs what about this TODO?

tezos/context/src/kv_store/persistent.rs
sebastiencs: Ah right, `put_context_hash` must be called before `commit_to_disk`

tezos/context/src/working_tree/working_tree.rs
sebastiencs: @tizoc, I will address it now

tezos/context/src/kv_store/persistent.rs
```diff
@@ -0,0 +1,559 @@
...
+    ///
+    big_strings_file: File,
+
+    hashes: Hashes,
+    shapes: DirectoryShapes,
+    string_interner: StringInterner,
+
+    pub context_hashes: Map<u64, ObjectReference>,
+
+    lock_file: Lock,
```
tizoc, line 96: @sebastiencs clippy says this is never read, lets use `_lock_file` instead (or add a clippy annotation, not sure what is the proper way to be explicit about it)

sebastiencs: @tizoc Done, I've added a clippy annotation

tizoc:
@sebastiencs I see hash ids are still 32bits, which makes me worry about it not being enough soon, wasn't the idea to increase the size? How many hashes do we have now?

Other than that, looks good to me, please combine groups of related commits into fewer commits as you think makes sense.

sebastiencs:
@tizoc Last time I checked we still could double the number of hashes with 32 bits, I will check again when storage is fully bootstrapped

PR #802, opened 2021-10-06
Make storage flush only once when needed, fix TmpStorage bug that might hang

vlad9486:


storage/src/database/tezedge_database.rs
```diff
@@ -139,16 +140,19 @@ impl<S: KVStoreKeyValueSchema> TezedgeDatabaseWithIterator<S> for TezedgeDatabas
 
 pub struct TezedgeDatabase {
     backend: Arc<TezedgeDatabaseBackend>,
+    log: slog::Logger,
 }
 
 impl TezedgeDatabase {
-    pub fn new(backend_option: TezedgeDatabaseBackendOptions) -> Self {
+    pub fn new(backend_option: TezedgeDatabaseBackendOptions, log: &Logger) -> Self {
```
bkontur, line 147: @vlad9486 
this is not recommended, to get reference and then just clone it, 
change pls. `log: &Logger` -> `log: Logger`, so that caller can decide to clone or give ownership

storage/src/commit_log/mod.rs
```diff
@@ -294,9 +298,7 @@ impl CommitLogs {
                     })?;
             match commit_log.flush() {
```
bkontur, line 299: @vlad9486 
I would write this just like:
```
if let Err(e) = commit_log.flush() {
    slog::error!(&self.log, "Failed to flush commit logs, reason: {:?}", e);
}
```
if we dont care about `Ok`

bkontur:
@vlad9486 
please, check at least clippy warnings `needless_borrow`,
and rebase it to the actual develop

PR #817, opened 2021-10-25
Operation content encoding

akoptelov:
Implemented all operation contents with decoding.

There is an `protocol::proto_XXX::operation::Operation` struct that is just an `p2p::encoding::operation::Operation` but with expanded `body` content. It can be obtained from an operation like this: `operation.try_into()`.


akoptelov:
@bkontur can we merge this one?

bkontur:
> 
> 
> @bkontur can we merge this one?

@akoptelov  well, its up to you :), I would maybe try to integrate and use them here:
```
async fn convert_block_operations_metadata(
    chain_id: ChainId,
    operations_proto_metadata_bytes: Vec<Vec<Vec<u8>>>,
    block_additional_data: &BlockAdditionalData,
    operations: Vec<OperationsForBlocksMessage>,
    env: &RpcServiceEnvironment,
) -> Result<BlockOperations, RpcServiceError> {
    // TODO: TE-521 - rewrite encoding part to rust
    let response = env
        .tezos_protocol_api()
        .readable_connection()
        .await?
        .apply_block_operations_metadata(
            chain_id,
            ApplyBlockRequest::convert_operations(operations),
            operations_proto_metadata_bytes,
            block_additional_data.protocol_hash.clone(),
            block_additional_data.next_protocol_hash.clone(),
        )
        .await
        .map_err(|e| RpcServiceError::UnexpectedError {
            reason: format!("Failed to call ffi, reason: {}", e),
        })?;

    serde_json::from_str::<BlockOperations>(&response).map_err(|e| e.into())
}
```

- from `block_additional_data.protocol_hash/next_protocol_hash`, so you can check. if we have rust encodinds `SupportedProtocol::try_from(block_additional_data.protocol_hash)`
  - if we have rust encodings for protocol, just use it directly do decode
  - if we dont have rust encodings for protocol, just trigger that fallback `env
        .tezos_protocol_api()
        .readable_connection()
        .await?
        .apply_block_operations_metadata(` as we do know
- that fallback would be usefull, if we implement (sometimes in future) dynamic protocol upgrade, that new protocol will come from p2p and there is not need for recompile or restart node, it would work and apply blocks, so this fallback would be used for new protocol

akoptelov:
> I would maybe try to integrate and use them here:
> ...

@bkontur I would gladly do that, but let's postpone it slightly, otherwise we will have nothing to do in the future... ;)

On the other hand, I badly need this for prevalidation, and would rather have this in `develop` instead of in a separate branch.

So, if you're OK with the current code, please approve and I'll merge this.

akoptelov:
One more thing, there are two more changes in encodings that are not reflected here. Let me add them.

akoptelov:
@bkontur Done.

PR #825, opened 2021-10-30
Make blocking calls in RPCs not block the scheduler

tizoc:


tizoc:
@binier about to merge this one, looks good to you?

tizoc:
Hmmm.. now that I look at this again, it should probably use `spawn_blocking` instead of `block_in_place`, the calling function is async.

tizoc:
Although maybe I tried spawn_blocking before ant it didn't work because values need to be send+sync, not sure now🧐

PR #841, opened 2021-11-12
Improve SignaturePublicKey(Hash) [TE-720]

akoptelov:
- Implement `TryFrom<SignaturePublicKey>` for `SignaturePublicKeyHash`
  to get hash directly from signature
- Do not build intermediate hex representation when building signatures
- Use base58check for Debug formatting

This is a prerequisite for endorsing stuff ([TE-713](https://viablesystems.atlassian.net/browse/TE-713))


crypto/src/hash.rs
```diff
@@ -306,6 +306,44 @@ pub fn chain_id_from_block_hash(block_hash: &BlockHash) -> Result<ChainId, Blake
         .unwrap_or_else(|_| unreachable!("ChainId is created from slice of correct size")))
 }
 
+#[derive(Debug, Error)]
+pub enum TryFromPKError {
```
bkontur, line 310: @akoptelov 
isnt it possible to refactor/use/clean a `ConversionError`  here instead of new `TryFromPKError` ?

bkontur:
@akoptelov 
and maybe it could be simplified and put together: `signature_public_key.rs`  + `signature_public_key_hash.rs`,
this was incrementally added stuff, maybe it could be simplified now

crypto/src/hash.rs
akoptelov: No, it's not. `ConversionError` is defined for signatures, in `tezos_messages` crate, and is quite overloaded to be used here. Here is the error that might happen solely for getting a hash of one typed hash and storing it into another.

PR #846, opened 2021-11-16
Docker: Add missing libz to distroless build

sebastiencs:


binier:
Committed directly to master that fix:
https://github.com/tezedge/tezedge/commit/b6c83d02971ad9ca21d92518f534c35d9f798a81

Did you run that locally to see if it works? For me `libz.so.1` didn't exist, only `libzstd.so.1`.

PR #854, opened 2021-11-26
Serialize `Hashid` on 6 bytes

sebastiencs:
This change `HashId` serialization size from 4 bytes to 6 bytes.
`HashId` will still be serialized on 4 bytes when the underlying `u64` fits in 32 bytes

### DIFF
--- a/tezos/context/src/kv_store/persistent.rs
+++ b/tezos/context/src/kv_store/persistent.rs
@@ -315,16 +315,16 @@ fn deserialize_hashes(
     let mut offset = commit_index_file.start();
     let end = commit_index_file.offset().as_u64();
 
-    let mut hash_id_bytes = [0u8; 4];
+    let mut hash_id_bytes = [0u8; 8];
     let mut hash_offset_bytes = [0u8; 8];
     let mut commit_hash: ObjectHash = Default::default();
 
     while offset < end {
         // commit index file is a sequence of entries that look like:
-        // [hash_id u32 ne bytes | offset u64 ne bytes | hash <HASH_LEN> bytes]
-        commit_index_file.read_exact_at(&mut hash_id_bytes, offset.into())?;
-        offset += hash_id_bytes.len() as u64;
-        let hash_id = u32::from_le_bytes(hash_id_bytes);
+        // [hash_id 6 le bytes | offset u64 le bytes | hash <HASH_LEN> bytes]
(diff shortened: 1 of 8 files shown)


tezos/context/src/kv_store/persistent.rs
```diff
@@ -315,16 +315,16 @@ fn deserialize_hashes(
...
     let mut hash_offset_bytes = [0u8; 8];
     let mut commit_hash: ObjectHash = Default::default();
 
     while offset < end {
         // commit index file is a sequence of entries that look like:
-        // [hash_id u32 ne bytes | offset u64 ne bytes | hash <HASH_LEN> bytes]
-        commit_index_file.read_exact_at(&mut hash_id_bytes, offset.into())?;
-        offset += hash_id_bytes.len() as u64;
-        let hash_id = u32::from_le_bytes(hash_id_bytes);
+        // [hash_id 6 le bytes | offset u64 le bytes | hash <HASH_LEN> bytes]
```
tizoc, line 324: ah good, I had already forgotten about the NE thing

tezos/context/src/serialize/mod.rs
```diff
@@ -278,53 +278,61 @@ pub fn deserialize_hash_id(data: &[u8]) -> Result<(Option<HashId>, usize), Deser
     let byte_hash_id = data.get(0).copied().ok_or(UnexpectedEOF)?;
 
     if byte_hash_id & 1 << 7 != 0 {
-        // The HashId is in 3 bytes
-        let hash_id = data.get(0..3).ok_or(UnexpectedEOF)?;
+        // The HashId is in 4 bytes
```
tizoc, line 281: @sebastiencs so, serialized hashids will be 4 bytes until they don't fit there anymore? how is this tested?

sebastiencs: @tizoc Yes, `HashId` will be 4 bytes as long as they fit in 31 bits. Starting with 32 bits, they will be serialized on 6 bytes.
The tests for this are there: https://github.com/tezedge/tezedge/pull/854/files#diff-eba41f64fb01eb676751ac27bd6a3a994a283b4e5ec57fb7af8977ad4d31a37cR1678

Before this PR, there is a similar behavior where HashId are serialized in 3 or 4 bytes.

PR #856, opened 2021-11-26
MessagePack encoding for action and state snapshot storage

binier:


tizoc:
@binier how much of a difference did this make?

edgekv/src/datastore.rs
```diff
@@ -574,7 +572,7 @@ impl DataStore {
 impl Drop for DataStore {
     fn drop(&mut self) {
         // TODO - TE-721: handle this error
-        self.sync_all(false);
+        let _ = self.sync_all(false);
```
tizoc, line 575: @binier `let _ =` will silence the warning but silently discard the error, something we should avoid

tizoc: There doesn't seem to be a way to obtain the current logger unless it is passed around and stored. What we have here is a common situation and it is kind of a PITA to solve because of that.

binier:
About 40% less disk usage after all headers and operations are synced and the difference grows. Due to smaller storage, ram usage is much lower as well. 2400gb vs 1600gb.

Also bootstrap is faster since pressure on disk is reduced and we are less bottlenecked by storage.

edgekv/src/datastore.rs
binier: Yes I agree. I just wanted to get rid of compilation warning.

PR #858, opened 2021-11-29
Truncate database files to their last valid sizes on restart 

sebastiencs:


tezos/context/src/persistent/file.rs
```diff
@@ -71,11 +73,12 @@ impl FileType {
             FileType::Strings => Path::new("strings.db"),
             FileType::Hashes => Path::new("hashes.db"),
             FileType::BigStrings => Path::new("big_strings.db"),
+            FileType::Sizes => Path::new("sizes.db"),
         }
     }
 }
 
-pub struct File {
+pub struct File<const T: FileType> {
```
tizoc, line 81: I like this a lot, but I was thinking about finally switching from nightly to current stable. IIRC the only unstable dependency we have right now is the `#[bench]` attribute, but it is easy to fix by adding an extra dependency. This would add a new dependency on nightly, and one that may not be stabilized soon.

sebastiencs: @tizoc I've updated the code so that nightly won't be required.
It now uses a generic based on a `u64`, which is available on stable:
```
struct File<const T: u64>
```

tizoc: @sebastiencs ok, great. Another reason for avoiding such features, is that if we want to package the storage so that it can be linked directly into the OCaml code (instead of the much more complicated approach we are using it now), depending on nightly features will make it harder (because Octez itself has some Rust dependencies, and ideally we should be able to compile everything with the Rust version it expects, or at least, one that is stable)

tizoc (approved):

PR #872, opened 2021-12-08
Add `ChunkedVec` to control reallocations of vectors

sebastiencs:


tizoc (approved):
Review wasn't very thorough, but on the surface it looks good. Please merge and re-deploy the storage server to see how it behaves.

tizoc:
@sebastiencs cool, I will go over it after the meeting

tezos/context/src/chunks/vec.rs
```diff
@@ -0,0 +1,378 @@
...
+
+            debug_assert_eq!(slice.len(), slice_length);
+
+            Some(Cow::Owned(slice))
+        }
+    }
+}
+
+impl<T> ChunkedVec<T> {
+    /// Returns a new `ChunkedSlice<T>` without allocating
```
tizoc, line 167: `ChunkedSlice` is mentioned a few times in the docs and examples but it doesn't exist anymore.

tezos/context/src/chunks/string.rs
```diff
@@ -0,0 +1,186 @@
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ChunkedString {
+    list_of_chunks: Vec<String>,
+    chunk_capacity: usize,
+    /// Number of bytes
+    nbytes: usize,
+}
+
+impl ChunkedString {
+    /// Returns a new `ChunkedSlice<T>` without allocating
```
tizoc, line 18: should be `ChunckedString`

tezos/context/src/lib.rs
```diff
@@ -113,6 +113,9 @@
 //! functionality that interacts with the repository (commit and checkout).
 //!
 
+#![feature(cow_is_borrowed)]
```
tizoc, line 116: Ouch, is this needed? cannot it be implemented in our own code instead?

tizoc: Looking at the diff I don't see where it is used btw.

tezos/context/src/working_tree/storage.rs
```diff
@@ -1400,36 +1414,36 @@ impl Storage {
 
     pub fn clear(&mut self) {
         if self.blobs.capacity() > 2048 {
-            self.blobs = Vec::with_capacity(2048);
+            self.blobs = ChunkedVec::with_chunk_capacity(2048);
```
tizoc, line 1417: These numbers repeat, and are also different from the current defaults set in `Storage::new` which I guess was not intended, so I think it would be good to define constants for them.

tizoc (changes requested):
@sebastiencs overall it looks good to me, but I left a few comments that need to be addressed first

tezos/context/src/lib.rs
sebastiencs: Thanks, it's not used I forgot to remove that line

tizoc (approved):
@sebastiencs were you able to figure out the issues with the hashes file? or if it could be related to the changes on this PR?

If you think this PR is unrelated, just go ahead and squash and merge.

PR #874, opened 2021-12-09
Move block application to state machine

binier:
- Remove context hash to block header unused index.
- Move protocol runner and context creation/initialization to shell_automaton (state machine).
- Move block application to shell_automaton (state machine).


light_node/src/main.rs
```diff
@@ -159,16 +150,24 @@ fn block_on_actors(
 
     // initialize shell automaton manager
     let (mut shell_automaton_manager, rpc_shell_automaton_channel) = ShellAutomatonManager::new(
+        tezos_protocol_api.clone(),
```
tizoc, line 153: Here it is clonned, but then wrapped in Arc later, shouldn't you be wrapping it first and passing `Arc::clone(&wrapped)` instead?

storage/tests/block_storage.rs
```diff
@@ -57,22 +54,6 @@ fn test_put_block_header_twice() -> Result<(), Error> {
     Ok(())
 }
 
-#[test]
```
tizoc, line 60: @binier aren't we going to do this anymore?

tizoc: I don't know if it is used for anything right now, I think you mentioned that it wasn't, just in these tests. But it is useful to do this to be able to check the consistency between the main db and the context storage at startup.

tezos/protocol-ipc-client/src/lib.rs
```diff
@@ -90,54 +94,68 @@ impl ProtocolRunnerConfiguration {
...
+        } else {
+            self.rx.receive().await?
+        };
+        Ok(result)
     }
 }
 
-impl ProtocolRunnerInstance {
-    pub const PROCESS_TERMINATE_WAIT_TIMEOUT: Duration = Duration::from_secs(10);
+/// Manages the execution of the protocol runners and access to their functionality.
```
tizoc, line 121: Remember that we want to be able to support multiple protocol runners, this combination of `ProtocolRunnerInstance` and `ProtocolRunnerApi` doesn't play well with that.

tizoc: Well, it all depends on how it ends being represented in the state machine. We probably will have many instances of this struct inside the state machine, right? with the connections explicitly handled there (instead of being asked-for by the caller like it works in the actors system right now, which is not a great way to do it).

light_node/src/main.rs
binier: no I need mutable access to it from state machine.

PR #882, opened 2021-12-15
Make state representable in JSON

akoptelov:
Use `HashMap`/`BTreeMap` with keys that are seiralized to JSON strings, otherwise such maps cannot be converted to JSON and thus state cannot be viewed in UI.

### DIFF
--- a/crypto/src/hash.rs
+++ b/crypto/src/hash.rs
@@ -1,7 +1,7 @@
 // Copyright (c) SimpleStaking, Viable Systems and Tezedge Contributors
 // SPDX-License-Identifier: MIT
 
-use std::convert::TryFrom;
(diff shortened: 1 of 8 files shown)


binier:
@akoptelov Maybe we should use this way instead? https://docs.rs/serde_with/latest/serde_with/#advanced-serde_as-usage

With current approach we will have to wrap every type that we will use as a key for `HashMap<...>`, if it's not `String`.

akoptelov:
@binier sorry, missed your comment, merged already. My point is that we should *always* use base58 for hashes. Now the blocker is some piece inside our storage that still somehow uses `serde`. When that would be fixed, there would be no need to wrap hashes.

PR #883, opened 2021-12-15
Compute checksums of the context files

sebastiencs:
Check the integrity of the context at the file level, by computing the checksum of the files.
This change the format of `sizes.db` and will make it incompatible with previous version.

The integrity check occurs in another thread, so that it doesn't block the main thread when initializing the context:
We won't have to change the value of `--initialize-context-timeout-in-secs`

### DIFF
--- a/light_node/src/configuration.rs
+++ b/light_node/src/configuration.rs
@@ -26,7 +26,7 @@ use tezos_api::environment::{TezosEnvironment, ZcashParams};
 use tezos_context_api::{
     ContextKvStoreConfiguration, PatchContext, SupportedContextKeyValueStore,
     TezosContextIrminStorageConfiguration, TezosContextStorageConfiguration,
-    TezosContextTezEdgeStorageConfiguration,
+    TezosContextTezEdgeStorageConfiguration, TezosContextTezedgeOnDiskBackendOptions,
 };
 
 #[derive(Debug, Clone)]
@@ -256,6 +256,12 @@ pub fn tezos_app() -> App<'static, 'static> {
                     }
                 }
             }))
+        .arg(Arg::with_name("context-integrity-check")
+            .long("context-integrity-check")
+            .global(true)
+            .takes_value(true)
+            .value_name("BOOL")
+            .help("Enable or not the integrity check on persistent tezedge context"))
         .arg(Arg::with_name("identity-file")
             .long("identity-file")
             .global(true)
@@ -1072,6 +1078,13 @@ impl Environment {
                             e
                         )
                     });
+
+                let startup_check = args
+                    .value_of("context-integrity-check")
+                    .unwrap_or("false")
+                    .parse::<bool>()
+                    .expect("Provided value cannot be converted to bool");
+
                 let context_kv_store = args
                     .value_of("context-kv-store")
                     .unwrap_or(Storage::DEFAULT_CONTEXT_KV_STORE_BACKEND)
@@ -1080,10 +1093,13 @@ impl Environment {
                         SupportedContextKeyValueStore::InMem => ContextKvStoreConfiguration::InMem,
                         SupportedContextKeyValueStore::OnDisk => {
                             ContextKvStoreConfiguration::OnDisk(
-                                get_final_path(&tezos_data_dir, "context".into())
-                                    .into_os_string()
-                                    .into_string()
-                                    .unwrap(),
+                                TezosContextTezedgeOnDiskBackendOptions {
+                                    base_path: get_final_path(&tezos_data_dir, "context".into())
+                                        .into_os_string()
+                                        .into_string()
+                                        .unwrap(),
+                                    startup_check,
+                                },
                             )
                         }
                     })
--- a/tezos/context/src/initializer.rs
+++ b/tezos/context/src/initializer.rs
@@ -2,6 +2,8 @@
 // SPDX-License-Identifier: MIT
 
 use std::sync::{Arc, RwLock};
+use std::thread::JoinHandle;
+use std::time::Duration;
 
 use ipc::IpcError;
 use ocaml_interop::BoxRoot;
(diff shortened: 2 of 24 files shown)


tizoc (changes requested):
I need to do another pass over this tomorrow, but other than the comments I left, one thing I think should change is that the checksum check at startup should be optional (probably off by default), and maybe also an extra command to just do the check without starting the node. Is it probably going to take a while, and startup is already slow.

tezos/context/src/initializer.rs
```diff
@@ -53,6 +54,44 @@ pub enum IndexInitializationError {
...
 }
 
+fn spawn_reload_database(
+    repository: Arc<RwLock<ContextKeyValueStore>>,
+) -> std::io::Result<JoinHandle<()>> {
+    let thread = std::thread::Builder::new().name("db-reload".to_string());
+    let (sender, recv) = std::sync::mpsc::channel();
+
+    let result = thread.spawn(move || {
+        println!("Reloading context");
```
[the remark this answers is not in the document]
sebastiencs: Ok, yes it prints correctly:
```
Dec 17 11:12:48.854 INFO [OCaml-out] Reloading context
Dec 17 11:12:48.973 INFO [OCaml-out] Context reloaded
```

tezos/context/src/kv_store/persistent.rs
```diff
@@ -406,16 +423,51 @@ impl Persistent {
...
+            &mut self.data_file,
+            &mut self.shape_file,
+            &mut self.shape_index_file,
+            &mut self.commit_index_file,
+            &mut self.strings_file,
+            &mut self.big_strings_file,
+            &mut self.hashes.hashes_file,
+        )?;
+
+        let shapes =
```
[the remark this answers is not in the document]
sebastiencs: Yes it's probably worth it, as they takes long time, I will do that

tezos/context/src/kv_store/persistent.rs
```diff
@@ -406,16 +520,91 @@ impl Persistent {
         Ok(())
     }
 
+    fn reload_database(&mut self) -> Result<(), IndexInitializationError> {
+        let list_sizes = FileSizes::make_list_from_file(&self.sizes_file);
+
+        let commit_counter = Self::truncate_files_with_correct_sizes(
```
tizoc, line 526: @sebastiencs when is this called? on every reload? on every load?

tezos/context/src/kv_store/persistent.rs
```diff
@@ -439,32 +628,29 @@ impl Persistent {
...
-
-    output.extend_from_slice(&file_sizes.commit_counter.to_le_bytes());
-    output.extend_from_slice(&file_sizes.data_size.to_le_bytes());
-    output.extend_from_slice(&file_sizes.shape_size.to_le_bytes());
-    output.extend_from_slice(&file_sizes.shape_index_size.to_le_bytes());
-    output.extend_from_slice(&file_sizes.commit_index_size.to_le_bytes());
-    output.extend_from_slice(&file_sizes.strings_size.to_le_bytes());
-    output.extend_from_slice(&file_sizes.hashes_size.to_le_bytes());
-    output.extend_from_slice(&file_sizes.big_strings_size.to_le_bytes());
+    let output = bincode::serialize(file_sizes).unwrap();
```
tizoc, line 631: what about this unwrap? can this never fail?

tezos/context/src/kv_store/persistent.rs
```diff
@@ -500,32 +686,20 @@ impl FileSizes {
...
-                commit_counter: u64::from_le_bytes(rest[0..8].try_into().unwrap()),
-                data_size: u64::from_le_bytes(rest[8..16].try_into().unwrap()),
-                shape_size: u64::from_le_bytes(rest[16..24].try_into().unwrap()),
-                shape_index_size: u64::from_le_bytes(rest[24..32].try_into().unwrap()),
-                commit_index_size: u64::from_le_bytes(rest[32..40].try_into().unwrap()),
-                strings_size: u64::from_le_bytes(rest[40..48].try_into().unwrap()),
-                hashes_size: u64::from_le_bytes(rest[48..56].try_into().unwrap()),
-                big_strings_size: u64::from_le_bytes(rest[56..64].try_into().unwrap()),
-            };
+            let sizes: FileSizes = bincode::deserialize(rest).unwrap();
```
tizoc, line 689: Another `unwrap` this without clarification about if it can fail or not

tezos/context/src/persistent/file.rs
```diff
@@ -225,6 +244,49 @@ impl<const T: TaggedFile> File<T> {
...
+        while offset < end {
+            let length_to_read = if offset + buffer.len() as u64 > end {
+                (end - offset) as usize
+            } else {
+                buffer.len()
+            };
+
+            let buffer = &mut buffer[..length_to_read];
+
+            self.read_exact_at(buffer, (offset as u64).into()).unwrap();
```
tizoc, line 262: Another `unwrap` without comment.

tezos/context/src/working_tree/working_tree.rs
```diff
@@ -172,7 +172,7 @@ impl TreeWalkerLevel {
         let children_iter = if should_continue {
             if let WorkingTreeRoot::Directory(dir_id) = &root.root {
                 let storage = root.index.storage.borrow();
-                let strings = root.index.string_interner.borrow();
+                let strings = root.index.get_string_interner().unwrap();
```
tizoc, line 175: Another `unwrap` without comment.

tezos/context/src/kv_store/persistent.rs
tizoc: I remember that it was changed so that it doesn't happen at startup, but then we need either the tool or a command line flag to enable that check (or even better both).

tizoc (changes requested):
@sebastiencs looks good but there are some unwraps without clarification about what they can fail, and it is not clear when the check would be performed, or how to trigger it.

sebastiencs:
@tizoc Comments addressed.
There is a new command line flag `tezedge-integrity-check` (default to `false`), it will enable or not the computation of checksums.
Note that even if `tezedge-integrity-check` is `false`, it will always check the sizes of the files.

tizoc:
> There is a new command line flag `tezedge-integrity-check` (default to `false`), it will enable or not the computation of checksums.

@sebastiencs it will enable the *check* at startup, but checksums will still be computed even when the flag is not enabled right?

light_node/src/configuration.rs
```diff
@@ -256,6 +256,12 @@ pub fn tezos_app() -> App<'static, 'static> {
                     }
                 }
             }))
+        .arg(Arg::with_name("tezedge-integrity-check")
```
tizoc, line 259: @sebastiencs let's use `--context-integrity-check` here.

sebastiencs:
@tizoc Yes sorry.
`--context-integrity-check` will control if we recompute the checksums of each file on startup (and compare them with their values in `sizes.db`).

The checksums will always be computed on every commit (to update `sizes.db`), the value of `--context-integrity-check` will not change that

tizoc (approved):
@sebastiencs ok, I don't have anything else to add, please squash commits in whatever way you think makes sense and lets have this merged.

Btw, can you use the tool you are working on to test this on the storage server to get a rough idea of the time it takes to verify the integrity of the context?

PR #887, opened 2021-12-20
When recording timings, recognize the protocol through an FFI call, n…

tizoc:
…ot by block hash.

### DIFF
--- a/tezos/context/src/ffi.rs
+++ b/tezos/context/src/ffi.rs
@@ -33,7 +33,7 @@ use crate::{
 };
 use tezos_context_api::TezosContextTezEdgeStorageConfiguration;
 use tezos_conv::{
-    OCamlBlockHash, OCamlContextHash, OCamlOperationHash,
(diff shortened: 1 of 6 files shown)


tizoc:
One small detail, the protocol can only be detected after a checkout (because the information is in the context).

sebastiencs:
@tizoc The code looks good to me, I just have a  few questions for my understanding:

- In what case the hook `tezedge_timing_set_protocol` would be called with a `None` `protocol_hash` ?
Are there some block/queries that are not related to a protocol ?
- `tezedge_timing_set_protocol` is called after every checkout ? Is there other cases where it is called ?

tizoc:
@sebastiencs the `None` protocol was a leftover from before, you were reading an outdated version. It will always contain a value.

`tezedge_timing_set_protocol` will be called after every checkout (when applying the block). Now I made another change to also call it after a commit if the protocol has changed.

sebastiencs (approved):

PR #889, opened 2021-12-20
Fix `HashId` with inodes

sebastiencs:
- Fix HashId created before commit on inodes.
  They are recomputed with the correct index (index on disk) and by removing the working tree bit
- Do not recompute hashes from deserialized inodes

### DIFF
--- a/tezos/context/src/serialize/mod.rs
+++ b/tezos/context/src/serialize/mod.rs
@@ -11,6 +11,7 @@ use tezos_timing::SerializeStats;
 use thiserror::Error;
 
 use crate::{
+    hash::HashingError,
     kv_store::HashId,
     persistent::DBError,
     working_tree::{
@@ -210,6 +211,11 @@ pub enum SerializationError {
     },
     #[error("Missing Offset")]
     MissingOffset,
+    #[error("Hashing Error: {error}")]
+    HashingError {
+        #[from]
+        error: HashingError,
+    },
 }
 
 #[derive(Debug, Error)]
--- a/tezos/context/src/serialize/mod.rs
+++ b/tezos/context/src/serialize/mod.rs
@@ -306,11 +306,17 @@ pub fn deserialize_hash_id(data: &[u8]) -> Result<(Option<HashId>, usize), Deser
     }
 }
 
-pub fn serialize_hash_id(
-    hash_id: u64,
+pub fn serialize_hash_id_impl(
+    hash_id: Option<HashId>,
     output: &mut Vec<u8>,
+    repository: &mut ContextKeyValueStore,
     stats: &mut SerializeStats,
 ) -> Result<(), SerializationError> {
+    let hash_id = match hash_id {
+        Some(hash_id) => repository.make_hash_id_ready_for_commit(hash_id)?.as_u64(),
+        None => 0,
+    };
+
     stats.highest_hash_id = stats.highest_hash_id.max(hash_id);
 
     if hash_id & FULL_31_BITS == hash_id {
@@ -336,4 +342,17 @@ pub fn serialize_hash_id(
         // MSB to determine if the HashId is compact or not
         Err(SerializationError::HashIdTooBig)
     }
+}
+
+pub fn serialize_hash_id<T>(
+    hash_id: T,
+    output: &mut Vec<u8>,
+    repository: &mut ContextKeyValueStore,
+    stats: &mut SerializeStats,
+) -> Result<(), SerializationError>
+where
+    T: Into<Option<HashId>>,
+{
+    let hash_id: Option<HashId> = hash_id.into();
+    serialize_hash_id_impl(hash_id, output, repository, stats)
 }
--- a/tezos/context/src/serialize/persistent.rs
+++ b/tezos/context/src/serialize/persistent.rs
@@ -595,18 +595,20 @@ impl PointersOffsetsHeader {
     fn from_pointers(
         object_offset: AbsoluteOffset,
         pointers: &[Option<PointerToInode>; 32],
-    ) -> Self {
+    ) -> Result<Self, SerializationError> {
         let mut bitfield = Self::default();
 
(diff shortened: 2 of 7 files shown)


tezos/context/src/serialize/mod.rs
```diff
@@ -306,11 +312,17 @@ pub fn deserialize_hash_id(data: &[u8]) -> Result<(Option<HashId>, usize), Deser
...
-    hash_id: u64,
+pub fn serialize_hash_id_impl(
+    hash_id: Option<HashId>,
     output: &mut Vec<u8>,
+    repository: &mut ContextKeyValueStore,
     stats: &mut SerializeStats,
 ) -> Result<(), SerializationError> {
+    let hash_id = match hash_id {
+        Some(hash_id) => repository.make_hash_id_ready_for_commit(hash_id)?.as_u64(),
+        None => 0,
```
tizoc, line 323: @sebastiencs how does this work? I am not clear on this logic here with a possible 0 and also the next line that gets the max hash id.

tezos/context/src/serialize/persistent.rs
```diff
@@ -1684,35 +1690,36 @@ mod tests {
 
     #[test]
     fn test_hash_id() {
+        let mut repo = Persistent::try_new(None).expect("failed to create context");
         let mut output = Vec::with_capacity(10);
         let mut stats = Default::default();
 
-        let number = 10101;
+        let number = HashId::new(10101).unwrap();
```
tizoc, line 1697: `HashId::try_new` would be a better name.

tezos/context/src/working_tree/storage.rs
```diff
@@ -398,7 +401,8 @@ pub struct PointerToInodeInner {
     hash_id: B48,
     is_commited: bool,
     inode_id: B31,
-    offset: B64,
+    is_offset_available: bool,
```
tizoc, line 404: How does this work?

tezos/context/src/serialize/mod.rs
sebastiencs: A `HashId` serialized to `0` will be deserialized as `Option::<HashId>::None`.
We have to handle this case for the parent of a commit, sometimes it can have no parent:
https://github.com/tezedge/tezedge/blob/develop/tezos/context/src/serialize/in_memory.rs#L538

The max hash id is only for statistics, it was useful for the in-memory context, to get an idea of the highest `HashId`

tezos/context/src/serialize/persistent.rs
sebastiencs: hmm not sure about that, `HashId::new` is only a call to `NonZeroU64::new`:
https://github.com/tezedge/tezedge/blob/develop/tezos/context/src/kv_store/mod.rs#L60
`NonZeroU64::new` is not `NonZeroU64::try_new`

tezos/context/src/working_tree/storage.rs
sebastiencs: This is to make sure that when we call, `PointerToInodeInner::get_offset`, the method return a valid (previously set) offset.
Without this bit, `get_offset` could return an invalid offset and we would have no error.
https://github.com/tezedge/tezedge/pull/889/files#diff-cfe31f13c183c59ed45a323b4b25dffd667af5d4b4d5d9e5af3cbd0a30a965b0R505

tezos/context/src/serialize/persistent.rs
tizoc: You are right. I always assumed `try_new` was expected when failure was possible.

tezos/context/src/working_tree/storage.rs
tizoc: I see, makes sense. But wouldn't it be better to use some sentinel value (like `MAX(u64)`) and not lose the bit? or what about 0? it isn't a valid offset right? (these are negative right? not absolute positions in the file)

tezos/context/src/working_tree/storage.rs
```diff
@@ -430,14 +435,16 @@ impl PointerToInode {
                     .with_hash_id(hash_id.map(|h| h.as_u64()).unwrap_or(0))
                     .with_is_commited(true)
                     .with_inode_id(inode_id.0)
-                    .with_offset(offset.map(|o| o.as_u64()).unwrap_or(0)), //.with_offset(0),
+                    .with_is_offset_available(offset.is_some())
+                    .with_offset(offset.map(|o| o.as_u64()).unwrap_or(0)),
```
tizoc, line 439: @sebastiencs I see here for example that 0 is used when the offset is not available, cannot that work as a sentinel value?

tezos/context/src/working_tree/storage.rs
sebastiencs: Ah yes, you are right, I will use `0` as sentinel value.

> these are negative right? not absolute positions in the file

Note that here, in `PointerToInodeInner`, the offset is the absolute offset, not relative.

Relative offset are used in the file, but when they are deserialized into the working tree (`PointerToInodeInner` and `DirEntryInner`), they are directly converted to absolute offset

tizoc: > Relative offset are used in the file, but when they are deserialized into the working tree (PointerToInodeInner and DirEntryInner), they are directly converted to absolute offset

Ok, related to this, remember that we don't want to load inodes that are not going to be used unnecessarily, in H that becomes a big issue is not handled properly.

tizoc:
@sebastiencs I don't have anything else to comment for now. Once ready lets merge it, ideally before the release so that it is included (this one is more or a priority than the integrity check IMO)

tizoc (approved):

sebastiencs:
@tizoc I will merge once the CI on Drone pass

tizoc:
@sebastiencs will need to be rebased before merging

tizoc:
Other than that every relevant test on drone passed

PR #894, opened 2021-12-21
Lazily read and deserialize inodes

sebastiencs:
There are 2 cases where the inodes are still fully read from disk:
- `WorkingTree::list`
- `fold`

### DIFF
--- a/tezos/context/src/working_tree/storage.rs
+++ b/tezos/context/src/working_tree/storage.rs
@@ -221,12 +221,16 @@ pub enum StorageError {
     DirEntryNotFound,
     #[error("InodeNotFound: Inode has not been found")]
     InodeNotFound,
+    #[error("InodeNotFound: Inode pointers has not been found")]
+    InodePointersNotFound,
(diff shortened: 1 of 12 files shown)


tezos/context/src/working_tree/storage.rs
```diff
@@ -1100,9 +1236,13 @@ impl Storage {
         match inode {
             Inode::Pointers { pointers, .. } => {
                 for pointer in pointers.iter().filter_map(|p| p.as_ref()) {
-                    let inode_id = pointer.inode_id();
-                    let inode = self.get_inode(inode_id)?;
-                    self.iter_inodes_recursive_unsorted(inode, fun)?;
+                    // When the inode is not deserialized, ignore it
```
tizoc, line 1239: @sebastiencs when is this version that skips non-deserialized inodes used? when committing?

tizoc (approved):

tezos/context/src/working_tree/storage.rs
sebastiencs: Yes on commit, to serialize and hash them

Issue #907, opened 2021-12-24
Test fails with v1.14.0


jdsika:
```
error: test failed, to rerun pass '-p tezos_interop --test bootstrap_storage_protocol_v1_test'

Caused by:
  process didn't exit successfully: `/home/it-tezos/tezedge/target/release/deps/bootstrap_storage_protocol_v1_test-a42ce6ec30b8881f` (signal: 4, SIGILL: illegal instruction)
```

jdsika:
```
LD_LIBRARY_PATH=./tezos/sys/lib_tezos/artifacts cargo run --release --bin light-node -- --config-file ../configs/tezedge.config
warning: /home/it-tezos/tezedge/storage/Cargo.toml: version requirement `0.5.3+zstd.1.4.5` for dependency `zstd` includes semver metadata which will be ignored, removing the metadata is recommended to avoid confusion
    Finished release [optimized] target(s) in 0.35s
     Running `target/release/light-node --config-file ../configs/tezedge.config`
thread 'main' panicked at 'Failed to launch protocol runner: SocketTimeout', light_node/src/main.rs:118:10
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

```

jdsika:
Config [here](https://github.com/tezedge/tezedge/issues/852#issuecomment-987838649)

tizoc:
Hello @jdsika, can you tell me about about your setup? (CPU, operating system, etc, output from `cat /proc/cpuinfo` would be great).

I suspect the problem you are seeing is related to https://github.com/supranational/blst/issues/10 (which would happen if your CPU doesn't support ADX).

jdsika:
Here is my [cpu.log](https://github.com/tezedge/tezedge/files/7780168/cpu.log)

tizoc:
@jdsika ok, I see there is no ADX, confirming that the issue is what I mentioned. Octez provides alternative binaries that don't require ADX, we will probably need to do the same. I need to investigate how, because the requirement comes from one of the cryptographic libraries.

jdsika:
Thanks, so I will wait for the next release. I mean the server is not THAAT old. I think they should be supported...

tizoc:
@jdsika hello, can you try rebuilding tezedge from the `no-adx` branch to see if that makes any difference ?

jdsika:
```
[...]

PR #909, opened 2021-12-28
Version without ADX support requirement

tizoc:


jdsika:
confirm working with #907 - when will this be merged and dockerized?

tizoc:
@jdsika next release (maybe today) will include this change.

PR #912, opened 2022-01-03
Add tools for tezedge persistent context

sebastiencs:
Examples:
- `cargo run --bin context-tool -- build-integrity -c PATH_TO_CONTEXT` (create a valid `sizes.db` file)
- `cargo run --bin context-tool -- is-valid-context -c PATH_TO_CONTEXT` (check if context is valid)
- `cargo run --bin context-tool -- context-size -c PATH_TO_CONTEXT` (display detailed context sizes)
- `cargo run --bin context-tool -- make-snapshot -c PATH_TO_CONTEXT` (create a snapshot based on a single commit)

### DIFF
--- a/tezos/context/src/working_tree/storage.rs
+++ b/tezos/context/src/working_tree/storage.rs
@@ -26,7 +26,7 @@ use crate::{
     hash::HashingError,
     kv_store::{index_map::IndexMap, HashId},
     working_tree::ObjectReference,
-    ContextKeyValueStore,
+    ContextKeyValueStore, ObjectHash,
 };
 use crate::{hash::index as index_of_key, serialize::persistent::AbsoluteOffset};
 
@@ -535,6 +535,11 @@ impl PointerToInode {
         } else {
             None
         }
+    }
+
+    pub fn hash_id_opt(&self) -> Option<HashId> {
+        let inner = self.inner.get();
+        HashId::new(inner.hash_id())
     }
 
     pub fn is_commited(&self) -> bool {
@@ -1065,6 +1070,7 @@ impl Storage {
         })
     }
 
+    #[allow(clippy::too_many_arguments)]
     fn insert_inode(
         &mut self,
         depth: u32,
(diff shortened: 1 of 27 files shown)


tizoc:
@sebastiencs before I go over the code, first thing is, please add a short README with a few examples and a list of supported commands with a short description.

sebastiencs:
@tizoc README file created

tezos/context/src/working_tree/working_tree.rs
```diff
@@ -1012,6 +1099,163 @@ impl WorkingTree {
         hash_directory(root, repository, &storage, &strings).map_err(MerkleError::from)
     }
 
+    fn traverse_working_tree_recursive(
```
tizoc, line 1102: @sebastiencs please add a comment here mentioning that this is only used by the tool, or move the implementation to a separate file so that it is obvious. That way it will be clear that the unwraps and panics here are not much of an issue.

tezos/context/src/working_tree/storage.rs
```diff
@@ -687,6 +700,81 @@ impl Storage {
...
+                ptr.set_hash_id(Some(new_hash_id));
+            }
+        }
+    }
+
+    pub fn forget_references(&mut self) {
+        self.offsets_to_hash_id = Default::default();
+
+        for (_, dir_entry_id) in self.directories.iter() {
+            let dir_entry = self.get_dir_entry(*dir_entry_id).unwrap();
```
tizoc, line 763: Same here and for anything that is tool-specific.

tizoc (approved):
@sebastiencs other than what I commented about separating possibly-panic code that is only used for the tool, looks good. There are other commands I would like to add later, like for example reading the value of a key (or even a whole subtree) in a specific commit.

sebastiencs:
I've moved `context-tool` specifics methods into their own modules

> There are other commands I would like to add later, like for example reading the value of a key (or even a whole subtree) in a specific commit.

I will check, but I think some existing methods can be re-used here, such as `TezedgeIndex::get_context_tree_by_prefix`.

tizoc:
@sebastiencs looks good. re: extra commands, those can be done later as a separate PR.

PR #915, opened 2022-01-10
Reduce memory used by inodes in the context

sebastiencs:


tezos/context/src/working_tree/storage.rs
```diff
@@ -239,6 +271,19 @@ impl From<DirEntryIdError> for StorageError {
...
 
+impl From<FatPointerIdError> for StorageError {
+    fn from(_: FatPointerIdError) -> Self {
+        Self::FatPointerIdError
+    }
+}
+
+impl From<std::convert::Infallible> for StorageError {
+    fn from(_: std::convert::Infallible) -> Self {
+        // This implementation exists only to be able to use `?` on a Result<_, Infallible>
```
tizoc, line 282: cool, I didn't know about `Infallible`

tezos/context/src/working_tree/storage.rs
```diff
@@ -399,175 +444,478 @@ impl TryFrom<usize> for InodeId {
...
+impl From<InodeId> for u32 {
+    fn from(v: InodeId) -> Self {
+        v.0
+    }
+}
+
+#[derive(Clone, Debug, Copy, PartialEq, Eq)]
+pub struct FatPointerId(u32);
+
+impl TryInto<usize> for FatPointerId {
```
tizoc, line 462: @sebastiencs when can this fail?

tizoc (approved):
@sebastiencs look ok to me, but to be honest, I only looked at it very superficially (it is quite a big diff). What I would like is more documentation on the new types that have been defined, while I understand the overall approach, I am not very clear on the specific components.

If it is well tested already, feel free to merge it, otherwise I can give it another pass after you have added some more docs.

tezos/context/src/working_tree/storage.rs
sebastiencs: Ah yes, it cannot fail. I will change the trait to `Into`

sebastiencs:
@tizoc Thanks, I will add more documentation

sebastiencs:
@tizoc I've updated the PR with more documentations

tizoc:
Weird, I don't know why the comments I added are not associated with the lines. But doesn't matter, those were not important.

tizoc (approved):

PR #917, opened 2022-01-14
Implement conditional codec for hashes

akoptelov:


crypto/src/hash.rs
```diff
@@ -67,16 +67,7 @@ pub enum FromBytesError {
...
-            Clone,
-            PartialEq,
-            Eq,
-            Serialize,
-            Deserialize,
-            std::cmp::PartialOrd,
-            std::cmp::Ord,
-            std::hash::Hash,
-        )]
+        #[derive(Clone, PartialEq, Eq, std::cmp::PartialOrd, std::cmp::Ord, std::hash::Hash)]
```
vlad9486, line 70: `PartialOrd`, `Ord`, `Hash` should be in std prelude. Full path is unnecessary.

crypto/src/seeded_step.rs
```diff
@@ -47,6 +47,8 @@ impl Step {
     }
 
     /// Returns new sequence number
+    // TODO rename it
```
tizoc, line 50: @akoptelov what about this todo? `next_step` or `step` instead?

tizoc (approved):
Other than what I just commented, looks good to me.

PR #918, opened 2022-01-14
Don't exclude operations that are not for the current branch from the…

tizoc:
… mempool pending operations result

### DIFF
--- a/shell_automaton/src/mempool/mempool_effects.rs
+++ b/shell_automaton/src/mempool/mempool_effects.rs
@@ -347,21 +347,13 @@ where
                     return;
                 }
             };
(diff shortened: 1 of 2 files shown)


tizoc:
Is there an easy way to produce encoded operation strings? I would like to add at least one test for this case.

akoptelov (approved):

akoptelov:
> Is there an easy way to produce encoded operation strings? I would like to add at least one test for this case.

There are multiple examples of JSON operations in Octez repo. You can take some and do
```
tezos-codec encode encode 011-PtHangz2.operation from operation.json
```

PR #933, opened 2022-02-07
Load context from disk and apply blocks in memory + Reduce memory

sebastiencs:
- Load the persistent context and apply blocks in memory
- Make `HashId` 6 bytes instead of 8. They are heavily used in the in-memory context and it saves memory
- Replace `HashMap` with `BTreeMap`. I don't know yet the impact on the memory usage, but it doesn't seem to change much
- Use `SortedMap` instead of `BTreeMap` in the garbage collector. This saves some spaces
- Add some logs in the garbage collector, activated with an environment variable

### DIFF
--- a/tezos/context/src/kv_store/in_memory.rs
+++ b/tezos/context/src/kv_store/in_memory.rs
@@ -23,6 +23,7 @@ use tezos_timing::{RepositoryMemoryUsage, SerializeStats};
 use crate::{
     chunks::ChunkedVec,
     gc::{
+        sorted_map::SortedMap,
         worker::{Command, Cycles, GCThread, GC_PENDING_HASHIDS, PRESERVE_CYCLE_COUNT},
         GarbageCollectionError, GarbageCollector,
     },
@@ -174,7 +175,7 @@ impl HashValueStore {
 }
 
 pub struct InMemory {
-    current_cycle: BTreeMap<HashId, Arc<[u8]>>,
+    current_cycle: SortedMap<HashId, Arc<[u8]>>,
     pub hashes: HashValueStore,
     sender: Option<Sender<Command>>,
     pub context_hashes: Map<u64, HashId>,
--- a/tezos/context/src/kv_store/in_memory.rs
+++ b/tezos/context/src/kv_store/in_memory.rs
@@ -383,6 +383,7 @@ impl InMemory {
                         recv,
                         free_ids: prod,
                         pending: Vec::new(),
+                        debug: false,
                     }
                     .run()
                 })?;
@@ -558,7 +559,7 @@ impl InMemory {
         &mut self,
         mut batch: ChunkedVec<(HashId, Arc<[u8]>)>,
     ) -> Result<(), DBError> {
-        while let Some(chunk) = batch.take_first() {
+        while let Some(chunk) = batch.pop_first_chunk() {
             for (hash_id, value) in chunk.into_iter() {
                 self.hashes.insert_value_at(hash_id, Arc::clone(&value))?;
                 self.current_cycle.push((hash_id, value));
--- a/tezos/context/src/kv_store/in_memory.rs
+++ b/tezos/context/src/kv_store/in_memory.rs
@@ -27,7 +27,10 @@ use crate::{
         GarbageCollectionError, GarbageCollector,
     },
     hash::ObjectHash,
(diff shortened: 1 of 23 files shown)


tezos/context/src/kv_store/in_memory.rs
```diff
@@ -386,14 +380,15 @@ impl InMemory {
...
+                        debug: false,
                     }
                     .run()
                 })?;
 
             (Some(sender), Some(cons), Some(thread_handle))
         };
 
-        let current_cycle = Default::default();
+        let current_cycle = ChunkedVec::with_chunk_capacity(512 * 1024);
```
tizoc, line 391: Same constant as above here too

tezos/context/src/kv_store/in_memory.rs
```diff
@@ -414,6 +409,130 @@ impl InMemory {
         })
     }
 
+    /// Reload context from disk
+    fn reload_database(&mut self) -> Result<(), ReloadError> {
+        let (tree, parent_hash, commit) = {
+            let mut ondisk = Persistent::try_new(PersistentConfiguration {
+                db_path: Some("/tmp/tezedge/context".to_string()),
```
tizoc, line 416: @sebastiencs this could come from somewhere else right? it is hardcoded right now

tezos/context/src/kv_store/in_memory.rs
```diff
@@ -434,17 +553,25 @@ impl InMemory {
...
         }
         Ok(())
     }
 
     pub fn new_cycle_started(&mut self) {
         if let Some(sender) = &self.sender {
-            let values_in_cycle = std::mem::take(&mut self.current_cycle);
+            let values_in_cycle = std::mem::replace(
+                &mut self.current_cycle,
+                ChunkedVec::with_chunk_capacity(512 * 1024),
```
tizoc, line 573: Same constant here again

tezos/context/src/tezedge_context.rs
```diff
@@ -1120,8 +1124,12 @@ impl TezedgeContext {
             repository.commit(&self.tree, self.parent_commit_ref, author, message, date)?
         };
 
+        let mem = self.get_memory_usage()?;
+
+        // println!("MEMORY_USAGE={:#?}", mem);
```
tizoc, line 1129: Leftover debug comment

tezos/context/src/working_tree/storage.rs
```diff
@@ -624,7 +624,7 @@ impl Iterator for PointersBitfieldIterator {
 
 /// Points to a subslice of `Storage::thin_pointers`
 ///
-/// An inode pointer may have 0 to 32 pointers.
+/// An inode pointer may have 1 to 32 pointers.
```
tizoc, line 627: Oh, good catch

tezos/context/src/working_tree/working_tree.rs
```diff
@@ -87,8 +88,8 @@ use super::{
 
 pub struct PostCommitData {
     pub commit_ref: ObjectReference,
-    pub batch: Vec<(HashId, Arc<[u8]>)>,
-    pub reused: Vec<HashId>,
+    pub batch: ChunkedVec<(HashId, Arc<[u8]>)>,
```
tizoc, line 91: General question, as long as the GC runs in parallel we cannot get rid of all the `Arc`s right? I think we discussed before but would it be possible to use coarser locking? having locks around big chunks of data instead of single values, I think that could save a lot of memory

tezos/context/src/working_tree/working_tree.rs
```diff
@@ -1124,7 +1125,7 @@ impl WorkingTree {
                         data.add_older_object(dir_entry, storage, strings)?;
                         return Ok(());
                     }
-                    dir_entry.set_commited(true);
+                    // dir_entry.set_commited(true);
```
tizoc, line 1128: is this not needed anymore?

tizoc (changes requested):
@sebastiencs left some comments. Looks good in general but I need to do another pass later

tezos/context/src/working_tree/working_tree.rs
sebastiencs: Yes, we need to `Arc` because of the GC. I will use coarser locking in my next PR https://github.com/tezedge/tezedge/pull/940

sebastiencs:
Closing for https://github.com/tezedge/tezedge/pull/940

PR #935, opened 2022-02-08
Store block metadata and is_applied=true for all blocks in snapshot

tizoc:


binier:
Just a note. We won't need this with new bootstrap, since it will rely on persisted current head and won't download/store any blocks that are before current head.

tizoc:
@binier cool, because what you describe is exactly how I assumed things worked, but then when trying to debug something related to snapshots and the fuzzer with @sebastiencs I noticed that the node was still applying old blocks. Note that the change here doesn't seem to be enough, because if the metadata is present then the node tries to fetch more data that is not present because I am not copying it, and that causes an error.

PR #945, opened 2022-02-16
use fixed size array for Nonce::new()

dkuehr:
Now instead of taking `&[u8]` we use `&[u8; NONCE_SIZE]` and remove the length check assertion. Every call to `Nonce::new` was fixed to the new argument type.

In case of `generate_nonces` a possible mismatch in sizes is converted to the already existing `Blake2bError::InvalidLenght`, however in practice this shouldn't happen unless a bug is present in the hashing function `blake2b::digest_256`.


akoptelov (approved):
lgtm

PR #965, opened 2022-03-01
Protocol runner restarts in state machine

tizoc:
- [x] handle restarts
- [ ] ~remove unwraps~ for another PR
- [ ] ~connections pool~ for another PR
- [x] retry logic (for block application at least)

### DIFF
--- a/shell/src/shell_automaton_manager.rs
+++ b/shell/src/shell_automaton_manager.rs
@@ -71,7 +71,7 @@ impl P2p {
 
 enum ShellAutomatonThreadHandle {
     Running(std::thread::JoinHandle<()>),
-    NotRunning(ShellAutomaton<ServiceDefault, MioInternalEventsContainer>),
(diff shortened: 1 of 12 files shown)


tizoc:
Except for the part that the block applier doesn't behave as we would want (retry the block when the failure is caused by the protocol runner going down), this initial version should be very similar (or even slightly better) than what we had in the actor system (detects the protocol runner failures earlier).

shell/src/shell_automaton_manager.rs
```diff
@@ -147,20 +147,23 @@ impl ShellAutomatonManager {
             log.new(o!("service" => "quota")),
         );
 
-        let protocol_service =
-            ProtocolServiceDefault::new(mio_service.waker(), Arc::new(protocol_runner_api.clone()));
+        let prevalidator_service = PrevalidatorServiceDefault::new(
```
tizoc, line 150: @binier che line 74 above (github doesn't let me put the comment there). Would it be ok to wrap that enum value in `Box` as clippy suggests? it is huge

PR #966, opened 2022-03-02
Disable retried application in mempool

akoptelov:
The node will always retry block application, but mempool will optionally ignore blocks applied after retry.

### DIFF
--- a/light_node/src/configuration.rs
+++ b/light_node/src/configuration.rs
@@ -1090,7 +1090,7 @@ impl Environment {
                         s.parse()
                             .expect("Boolean value expected for disable-endorsements-precheck")
                     }),
-                disable_apply_retry: args.value_of("disable-apply-retry").map_or(true, |s| {
+                disable_apply_retry: args.value_of("disable-apply-retry").map_or(false, |s| {
                     s.parse()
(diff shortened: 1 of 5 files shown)


binier:
@akoptelov 
Will endorsement be added in the mempool for the block that was applied but ignored by the mempool? Probably not since precheck of prevalidator will be for current head in the mempool, not last applied block. Unless we can have endorsements for future blocks in the mempool.
https://github.com/tezedge/tezedge/blob/7259ce9ab2030a32ec5a187cff769885fb98531e/shell_automaton/src/mempool/mempool_reducer.rs#L464

akoptelov:
@binier they will be in the mempool. This is a question if that is good or bad. Also cleaning up old endorsement doesn't look perfect to me. My point is that this fix is not the ultimate one. This is just to fix the issue we saw on tezedge.com, that caused node to hang. With this fix, the node won't hang in the same situation, it will not broadcast that problematic block. This is an improvement. There are still a lot of corner cases, I know, but I'm not aware of any serious flaw that is connected to this exact fix.

akoptelov:
BTW, now I made that option false by default, so behavior will be just as it was before my previous fix, unless changed explicitly.

PR #967, opened 2022-03-03
Replace chrono crate

akoptelov:
- Replaced `chrono` with `time` in Tezedge sources
- Bumped `slog-term` to get rid of `chrono` dep
- Patched `ipmpsc` with `vergen` updated to `6`
- Patched `vergen` with `chrono` replaced with `time`
- Replaced `zip` with git `master`that has `chrono` replaced with `time` (dev dependency, shouln't be an issue)
- Udated `zstd` dependency in `storage` to match native lib version with the one from `zip`

### DIFF
--- a/apps/node_monitoring/src/monitors/resource.rs
+++ b/apps/node_monitoring/src/monitors/resource.rs
@@ -9,7 +9,6 @@ use std::hash::Hash;
 use std::sync::{Arc, RwLock};
 use std::time::Instant;
 
-use chrono::Utc;
 use thiserror::Error;
 use getset::Getters;
 use merge::Merge;
@@ -17,6 +16,7 @@ use netinfo::Netinfo;
 use serde::Serialize;
 use slog::{error, warn, Logger};
 use sysinfo::{System, SystemExt};
+use time::OffsetDateTime;
 
 use crate::display_info::{NodeInfo, OcamlDiskData, TezedgeDiskData};
 use crate::monitors::alerts::Alerts;
@@ -441,13 +441,13 @@ impl ResourceMonitor {
                     };
 
                     ResourceUtilization {
-                        timestamp: chrono::Local::now().timestamp(),
+                        timestamp: OffsetDateTime::now_utc().unix_timestamp(),
                         memory: MemoryStats {
                             node: node_memory,
                             validators: validators_memory,
                         },
-                        tezedge_disk: Some(node_disk.try_into()?),
                         ocaml_disk: None,
+                        tezedge_disk: Some(node_disk.try_into()?),
                         cpu: CpuStats {
                             node: node_cpu,
                             validators: validators_cpu,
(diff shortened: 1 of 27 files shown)


logging/src/detailed_json.rs
```diff
@@ -60,5 +62,12 @@ pub fn default<W>(io: W) -> slog_json::Json<W>
 where
     W: io::Write,
 {
-    new_with_ts_fn(io, |_: &Record| chrono::Local::now().to_rfc3339()).build()
+    const INVALID_TIME: &str = "invalid timestamp";
+    new_with_ts_fn(io, |_: &Record| match OffsetDateTime::now_local() {
```
akoptelov, line 66: Guys, here local timestamp is used. Wouldn't it be better to use UTC instead?

apps/node_monitoring/src/monitors/resource.rs
```diff
@@ -441,13 +441,13 @@ impl ResourceMonitor {
                     };
 
                     ResourceUtilization {
-                        timestamp: chrono::Local::now().timestamp(),
+                        timestamp: OffsetDateTime::now_utc().unix_timestamp(),
```
adonagy, line 444: This time is what is reported in the FE explorer. Shouldn't we keep the local time? On the other hand, maybe we should unify everything under UTC, just as you are saying here: https://github.com/tezedge/tezedge/pull/967/files#r818413396

The block timestamps are in UTC, so maybe it is the wisest to use UTC everywhere

logging/src/detailed_json.rs
adonagy: As I said above, I agree using UTC everywhere, as the block timestamps are UTCs, and when we are troubleshooting stuff the blocks and their timestamps are referenced a lot.

adonagy (approved):
If we agree on using UTC everywhere, LGTM 👍

apps/node_monitoring/src/monitors/resource.rs
tizoc: +1 for UTC everywhere

shell/Cargo.toml
```diff
@@ -38,9 +38,9 @@ tezos_protocol_ipc_client = { path = "../tezos/protocol-ipc-client" }
 [dev-dependencies]
 serial_test = "0.5"
 slog-async = "2.6"
-slog-term = "2.8"
+slog-term = "2.9"
 fs_extra = "1.2.0"
-zip = "0.5.5"
+zip = { git = "https://github.com/zip-rs/zip.git" }
```
tizoc, line 43: @akoptelov it would be better to have a tag here, just in case future changes break compatibility

tizoc (approved):
@akoptelov looks good to me, the only thing I would change is to be more explicit about the "zip" dependency pointing to a repo (by setting a commit hash or tag)

sebastiencs (approved):

apps/node_monitoring/src/monitors/resource.rs
akoptelov: The timestamp should be the same, whether it is obtained from local now or from UTC now, to my understanding.

PR #969, opened 2022-03-04
Prepare develop for release/v1.16.1

akoptelov:


apps/deploy_monitoring/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "deploy-monitoring"
-version = "1.17.0"
```
tizoc, line 3: @akoptelov `1.16.1` right?

akoptelov: Looks like it is not handled by `version.sh`. I'll fix that.

tizoc: Ok, same for the title of the PR (I had exactly the same issue when releasing 1.16.0)

akoptelov: Closing this one in favor of https://github.com/tezedge/tezedge/pull/970

PR #974, opened 2022-03-09
Implement 012-PsIthaca operation contents

akoptelov:
- Added JSON encoding
- newtype for Fitness
- newtype for Bytes
- implement `encoding(reserve)`


akoptelov:
@binier Zura, now `Fitness` is a struct with proper ordering. We might also add protocol-specific debug formatting.

@vlad9486 Vlad, now almost all operations support binary and JSON encoding, with exception of `Origination` and `Transaction` with parameters, because of unimplemented script and parameter types. I'll look once more into your implementation to see if anything is still missing here.

akoptelov:
@dkuehr Daniel, can you check the `fuzzcheck` derive support for generics with constant params? Now the GH action reports an error...

tezos/messages/src/protocol/proto_012/operation.rs
```diff
@@ -3,69 +3,56 @@
 
 //! Operation contents. This is the contents of the opaque field [super::operation::Operation::data].
 //!
-//! See https://tezos.gitlab.io/protocols/012_ithaca.html
+//! Comparing to the previous protocol, new operation [`RegisterGlobalConstant`] has been added.
+//! See http://tezos.gitlab.io/protocols/011_hangzhou.html#global-constants.
```
tizoc, line 7: @akoptelov shouldn't this be in `proto_011/operation.rs` instead?

akoptelov: Sure, forgot to update the comment.

PR #975, opened 2022-03-10
Create snapshot from the in-memory context

sebastiencs:


shell_automaton/src/action.rs
```diff
@@ -232,6 +236,11 @@ pub enum Action {
     ProtocolRunnerSpawnServerError(ProtocolRunnerSpawnServerErrorAction),
     ProtocolRunnerSpawnServerSuccess(ProtocolRunnerSpawnServerSuccessAction),
 
+    ProtocolRunnerCurrentHeadInit(ProtocolRunnerCurrentHeadInitAction),
```
tizoc, line 239: @sebastiencs I would rename these so that it is not ambiguous (LatestCommitHashes or LatestContextHashes, so it is not confused with the "current head" concept from the shell).

shell_automaton/src/protocol_runner/protocol_runner_effects.rs
```diff
@@ -103,6 +107,37 @@ where
...
+                                store.dispatch(ProtocolRunnerCurrentHeadSuccessAction {
+                                    token,
+                                    latest_context_hashes,
+                                });
+                                continue;
+                            }
+                            ProtocolRunnerResult::GetCurrentHead((token, Err(error))) => {
+                                slog::error!(&store.state().log, "failed to get context's latest commits";
+                                    "error" => format!("{:?}", error));
+                                store.dispatch(ProtocolRunnerCurrentHeadSuccessAction {
```
binier, line 125: Shouldn't you dispatch error here? Probably you are dispatching success action because even if getting latest context hashes fails, you want it to work. But you still can do that by dispatching error and afterwards doing exact same thing as in case of success.

shell_automaton/src/service/protocol_runner_service.rs
```diff
@@ -49,6 +50,13 @@ pub enum ProtocolRunnerResult {
     ),
     InitContextIpcServer((ProtocolRunnerToken, Result<(), ProtocolServiceError>)),
 
+    GetCurrentHead(
```
binier, line 53: Please move this bellow:
```
GenesisCommitResultGet(
        (
            ProtocolRunnerToken,
            Result<CommitGenesisResult, ProtocolServiceError>,
        ),
    ),
```
To write them in same order as they are later called in. `GenesisCommitResultGet` is called before `GetCurrentHead`.

Also please rename it to `CurrentHeadGet` to make naming consistent.

shell_automaton/src/service/protocol_runner_service.rs
```diff
@@ -87,6 +95,7 @@ impl ProtocolRunnerResult {
             Self::InitRuntime((token, _)) => Some(*token),
             Self::InitContext((token, _)) => Some(*token),
             Self::InitContextIpcServer((token, _)) => Some(*token),
+            Self::GetCurrentHead((token, _)) => Some(*token),
```
binier, line 98: same here. Move down.

shell_automaton/src/service/protocol_runner_service.rs
```diff
@@ -131,6 +140,8 @@ pub trait ProtocolRunnerService {
 
     fn apply_block(&mut self, req: ApplyBlockRequest);
 
+    fn get_latest_context_hashes(&mut self, count: i64) -> ProtocolRunnerToken;
```
binier, line 143: Move up, so that it's after `genesis_commit_result_get_init`.

shell_automaton/src/service/protocol_runner_service.rs
```diff
@@ -340,4 +351,13 @@ impl ProtocolRunnerService for ProtocolRunnerServiceDefault {
             .blocking_send(ProtocolRunnerRequest::ShutdownServer(()))
             .unwrap();
     }
+
+    fn get_latest_context_hashes(&mut self, count: i64) -> ProtocolRunnerToken {
```
binier, line 355: Same here, move above.

shell_automaton/src/service/storage_service.rs
```diff
@@ -227,6 +228,55 @@ impl fmt::Debug for StorageServiceDefault {
     }
 }
 
+fn find_block_matching_context_storage_impl(
+    head_block_storage: BlockHeaderWithHash,
```
binier, line 232: Name is a bit confusing. Maybe it should be `head_block` or `head_block_from_storage`?

shell_automaton/src/protocol_runner/current_head/protocol_runner_current_head_state.rs
```diff
@@ -0,0 +1,29 @@
...
+use crypto::hash::ContextHash;
+use serde::{Deserialize, Serialize};
+use tezos_protocol_ipc_client::ProtocolRunnerError;
+
+use crate::protocol_runner::ProtocolRunnerToken;
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub enum ProtocolRunnerCurrentHeadState {
+    Init {
+        genesis_commit_hash: Option<ContextHash>,
```
tizoc, line 13: Why is this needed here? just so that it can be passed around for the final step? (because all these are not intermediary states before the initialization is complete)?

tezos/api/src/ffi.rs
```diff
@@ -179,6 +179,21 @@ pub struct BeginApplicationResponse {
     pub result: String,
 }
 
+#[cfg_attr(feature = "fuzzing", derive(fuzzcheck::DefaultMutator))]
+#[derive(Error, Serialize, Deserialize, Debug, Clone)]
+pub enum GetLastContextHashesError {
+    #[error("Failed to get the latest context hashes: {message}!")]
+    FailedToGetCurrentHead { message: String },
```
binier, line 186: In general it's better if we don't use `String` for errors and instead use concrete types, but I see `TezosErrorTrace` contains string errors as well, so we can refactor this later as a part of a refactor to get rid of strings in the error types.

binier (changes requested):
Please take a look at comments that I added, mostly cosmetic changes that I'm requesting.

Otherwise looks great! :)

shell_automaton/src/protocol_runner/protocol_runner_state.rs
```diff
@@ -27,6 +29,10 @@ pub enum ProtocolRunnerState {
     #[from]
     Init(ProtocolRunnerInitState),
 
+    /// Load context's current head
+    #[from]
+    GetCurrentHead(ProtocolRunnerCurrentHeadState),
```
tizoc, line 34: @sebastiencs I was thinking, do we really need a new separate message for this or is it something that would make sense as part of the context initialization step? (not worth to change it now that it is working, but something to keep in mind for the future once we start cleaning up all the protocol-runner stuff)

shell_automaton/src/service/storage_service.rs
```diff
@@ -227,6 +228,55 @@ impl fmt::Debug for StorageServiceDefault {
...
+    }
+
+    match find_block_matching_context_storage_impl(
+        head_block_storage.clone(),
+        &latest_context_hashes,
+        block_storage,
+    )? {
+        Some(head) => Ok(head),
+        None => {
+            slog::warn!(
```
tizoc, line 270: This should be an error, because the node will not be able to recover from this situation (other than by trying to find older context hashes to see if there is a match, or by re-starting the bootstrap process). But if this happens it is because something went very wrong and it is probably to just bail-out and notify the user of a potential corruption. What do you guys think?

tezos/context/src/gc/worker.rs
```diff
@@ -142,6 +228,41 @@ enum GCError {
...
+        error: IndexInitializationError,
+    },
+    #[error("The GC thread doesn't have access to the in-mem repository")]
+    InMemContextNotFound,
+    #[error("String not found in the string interner")]
+    StringNotFound,
+    #[error("Unable to create a valid snapshot path")]
+    InvalidSnapshotPath,
+    #[error("Call to renameat2 failed with {error}")]
+    RenameAt2Failed { error: std::io::Error },
```
tizoc, line 265: clippy says this is never used

tezos/context/src/gc/worker.rs
```diff
@@ -400,42 +609,308 @@ impl GCThread {
...
+        output: &mut SerializeOutput,
+        on_disk_repo: &mut Persistent,
+    ) -> Result<(), GCError> {
+        let data_len = output.len();
+        let hashes_len = on_disk_repo.hashes_in_memory_len();
+        let strings_len = on_disk_repo
+            .string_interner
+            .new_bytes_since_last_serialize();
+
+        let data_limit_reached = data_len >= 20_000_000;
```
tizoc, line 702: @sebastiencs can you declare these as constants? also document a bit the behavior here

tezos/context/src/gc/worker.rs
```diff
@@ -400,42 +609,308 @@ impl GCThread {
...
+            on_disk_repo.commit_to_disk(output).map_err(DBError::from)?;
+            on_disk_repo.set_is_commiting();
+            on_disk_repo.deallocate_strings_shapes();
+            output.clear();
+        }
+
         Ok(())
     }
 
+    fn find_frozen_commit(&self, commit_hash_id: HashId, bytes: &mut Vec<u8>) -> Option<HashId> {
```
tizoc, line 716: What is a frozen commit?

tezos/context/src/kv_store/persistent.rs
```diff
@@ -131,7 +132,7 @@ pub struct Persistent {
     /// This repeats 10 times
     sizes_file: File<{ TAG_SIZES }>,
     startup_check: bool,
-    last_commit_on_startup: Option<ObjectReference>,
+    last_commits_on_startup: VecDeque<ObjectReference>,
```
tizoc, line 135: `latest_commits_...` ?

tezos/context/src/gc/worker.rs
binier: probably he means commit for cemented block maybe. Blocks that are <= HEAD~2.

shell_automaton/src/protocol_runner/current_head/protocol_runner_current_head_state.rs
binier: I think it's to carry the value for `ProtocolRunnerState::Ready` state.

tizoc: It is kind of ugly, because on each step it then needs to be extracted and re-passed, but I guess there is no point in trying to change that now.

shell_automaton/src/service/storage_service.rs
tizoc: `find_block_with_context_hash` ?

tezos/api/src/ffi.rs
tizoc: The traces that we get from the protocol runner are strings containing the JSON representation of the OCaml error trace, we usually don't do anything with those other than displaying them for debugging purposes.

In the cases where it is needed we can check the "id" field of the json object to figure out what the error is and map it to something concrete (like here https://github.com/tezedge/tezedge/blob/97b3faa5b9ceee56a41bb03554fafba8c623eca0/tezos/api/src/ffi.rs#L532-L558)

tizoc:
@sebastiencs overall looks good, left some comments.

shell_automaton/src/protocol_runner/current_head/protocol_runner_current_head_state.rs
sebastiencs: @tizoc Yes this is to keep `genesis_commit_hash` until `ProtocolRunnerState::Ready` is reached.
Not sure what could be an alternative way to keep `genesis_commit_hash`

shell_automaton/src/service/protocol_runner_service.rs
sebastiencs: Alright, I will rename to `LatestContextHashesGet`

shell_automaton/src/service/storage_service.rs
sebastiencs: Ok, I will rename and add some documentation

shell_automaton/src/protocol_runner/protocol_runner_state.rs
sebastiencs: I did it in a separate message/step to keep initializing the protocol runner as fast as possible.
The initialization of the context doesn't take more than 1 second, however retrieving the latest context hashes takes at least 15 seconds.

shell_automaton/src/service/storage_service.rs
sebastiencs: I was thinking that in the case where the context storage is level 1000, and the block storage is level 100, that `find_block_matching_context_storage` will not be able to find the corresponding block, since we only get the last few context hashes from the context.

> case where the context storage is level 1000, and the block storage is level 100

Not sure if it is a realistic scenario though

shell_automaton/src/protocol_runner/protocol_runner_state.rs
tizoc: Uhhh, why does it take so long to obtain the context hashes? we have those in the index, which just needs to be scanned from the end, right?

tezos/context/src/gc/worker.rs
sebastiencs: Yes that's a commit at the current block level minus 3. I will add some documentation

shell_automaton/src/protocol_runner/protocol_runner_effects.rs
sebastiencs: Thanks, I will do that

shell_automaton/src/protocol_runner/protocol_runner_state.rs
sebastiencs: Currently, the commit index is available after computing the crc32 of all files, and after the deserialization of the strings and shapes:
https://github.com/tezedge/tezedge/blob/develop/tezos/context/src/kv_store/persistent.rs#L660

In the case of the in-memory repository (loading from snapshot), it also has to commit the full tree (loaded from disk) into the in-mem repository:
https://github.com/tezedge/tezedge/blob/develop/tezos/context/src/kv_store/in_memory.rs#L561

Though with some changes, we could make the context hashes available faster

tizoc: For the in-memory case we still want the context to be ready before continuing, so getting the context hash faster would probably not help much. Same for the persisted context, but I think that part about deserializing strings and shapes could be speed up right? currently everything is deserialized although most of it is probably not used (there are shapes that repeat a lot and many that are used only one, or a few times and never again, and we could avoid paying the cost of parsing those and keeping them in memory).

sebastiencs: > I think that part about deserializing strings and shapes could be speed up right? currently everything is deserialized although most of it is probably not used

hmm yes we could do that, but we would have to read the full tree to get the list of strings & shapes used.
That's something we can test yes, and compare with the current speed

tizoc: We kind of know which ones are going to be used more right? and those could be kept on their own index. The others get another index that is never loaded at startup and works as an LRU cache. Anwyay, that would be for a separate PR

shell_automaton/src/current_head/current_head_effects.rs
```diff
@@ -25,8 +26,18 @@ where
             let chain_id = store.state().config.chain_id.clone();
             let level_override = store.state().config.current_head_level_override;
             let storage_req_id = store.state().storage.requests.next_req_id();
+
+            let latest_context_hashes = match &store.state().protocol_runner {
+                ProtocolRunnerState::Ready(state) => state.latest_context_hashes.clone(),
+                _ => Vec::new(),
```
tizoc, line 32: @binier can this situation (we get to this point without the state being `Ready`) happen? if this is not something that should ever happen I would log a warning here just in case (cc @sebastiencs )

tizoc (approved):
Looks good to me.

PR #978, opened 2022-03-14
Disable prechecking for Ithaca

akoptelov:
Currently we don't support prechecking for the new protocol, but still there is a lot of actions that are executed and reporting errors (e.g. baking and endorsing rights calculation). We need to disable them basing on the block's protocol.


tizoc:
@akoptelov anything pending for this one? or can it be merged already?

PR #979, opened 2022-03-14
Do not always sort keys on fold

sebastiencs:


tizoc:
@sebastiencs this will still construct a vector with all entries in-memory right? that is what we want to avoid for the unsorted version (because after the flattening, there are directories that are huge).

tizoc (approved):

sebastiencs:
@tizoc I've updated the PR to not allocate every strings on fold. I will merge when the tests pass

tezos/context/src/working_tree/working_tree.rs
```diff
@@ -353,12 +360,12 @@ impl Iterator for TreeWalker {
                     if let Some((k, dir_entry)) = iter.next() {
                         match current_level.root.dir_entry_tree(dir_entry) {
                             Ok(root) => {
-                                // TODO: this is not very efficient, maybe we need to improve the key representation
-                                let mut key = current_level.key.clone();
-                                key.push(k.to_string());
+                                let mut keys = Vec::with_capacity(current_level.key.len() + 1);
```
tizoc, line 363: @sebastiencs why "keys" ? it is still a single key / path.

PR #980, opened 2022-03-14
Move context to a different directory than irmin

sebastiencs:


sebastiencs:
Replaced by https://github.com/tezedge/tezedge/pull/982

PR #981, opened 2022-03-14
Handle retried blocks consistently with Octez

akoptelov:
Block should be skipped iff:
* it has 012_PtHangz2 protocol
* it has been applied successfully after failed attempt because of context mismatch


tizoc (approved):
Looks good to me.

PR #982, opened 2022-03-14
Adjusts disk measurements taken by `node_monitoring`

tizoc:
Doesn't differentiate between context backends, always measures "context/"
Also replaces context-actions with context-stats-db


tizoc:
@adonagy btw, what do you think of changing `OCaml -> Octez` ? anything that would break if I change that?

tizoc:
https://github.com/tezedge/monitoring-test/pull/1 for the test image related changes

PR #1005, opened 2022-03-23
Feature - import-snapshot subcommand

adonagy:


light_node/src/snapshot_command.rs
```diff
@@ -432,3 +439,60 @@ fn resolve_block_reference(
...
     }
 }
+
+pub fn import_snapshot(from: &Path, tezos_data_dir: &Path) {
+    // remove existing data
+    let to_remove = [
+        tezos_data_dir.join("context"),
+        tezos_data_dir.join("bootstrap_db"),
+    ];
+    fs_extra::remove_items(&to_remove).expect("Failed to remove existing data");
```
tizoc, line 449: @adonagy because it can fail in the next step, I think it would be better to just ask the user to remove the files himself and bail-out here.

light_node/src/snapshot_command.rs
```diff
@@ -432,3 +439,60 @@ fn resolve_block_reference(
...
+
+    let tar_gz = File::open(from)
+        .unwrap_or_else(|_| panic!("Failed to open snapshot file {}", from.display()));
+
+    let buffered = BufReader::new(tar_gz);
```
tizoc, line 454: @adonagy if we go the "mirroring download" route (which I think is the best because it doesn't duplicate the amount of space needed), I think compression can be handled at transfer time (with gzip encoding), that way nothing special needs to be done about the files (neither during generation or during import)

tizoc: Btw "mirroring" is probably misleading, what I meant is using the manifest file as a guide to decide what to download (so the tool only needs to find that file, then download everything listed there and verify checksums, also, for files that already exists, if the checksum matches, then it can just skip the download -- it would be like resuming from an older interrupted attempt I guess)

light_node/src/snapshot_command.rs
adonagy: Oh right, good point 👍

light_node/etc/tezedge/tezedge.config
```diff
@@ -101,7 +101,7 @@
 
 # Path to a tezos protocol runner executable
 # --protocol-runner <PATH>
---protocol-runner=./target/debug/protocol-runner
```
tizoc, line 104: 👀 the day we get rid of this config system I will be so glad

PR #1031, opened 2022-04-01
Endorsements ithaca backend

akoptelov:


rpc/src/services/dev_services.rs
```diff
@@ -910,16 +914,79 @@ pub(crate) async fn get_shell_automaton_endorsing_rights(
...
+    round: Option<i32>,
+    base_time: Option<u64>,
     env: &RpcServiceEnvironment,
 ) -> anyhow::Result<serde_json::Value> {
+    use shell_automaton::mempool::mempool_actions::ConsensusOperationMatcher;
+    let matcher = match (block_hash, payload_hash, level, round) {
+        (Some(block_hash), None, None, None) => {
+            ConsensusOperationMatcher::endorsement_branch(block_hash)
+        }
+        //        (None, Some(payload_hash), None, None) => ConsensusOperationMatcher::endorsement_payload(payload_hash),
```
tizoc, line 948: what about this commented out case? not needed?

rpc/src/services/dev_services.rs
```diff
@@ -910,16 +914,79 @@ pub(crate) async fn get_shell_automaton_endorsing_rights(
...
+pub(crate) async fn get_shell_automaton_preendorsements_status(
+    payload_hash: Option<BlockPayloadHash>,
+    level: Option<i32>,
+    round: Option<i32>,
+    base_time: Option<u64>,
+    env: &RpcServiceEnvironment,
+) -> anyhow::Result<serde_json::Value> {
+    use shell_automaton::mempool::mempool_actions::ConsensusOperationMatcher;
+    let matcher = match (payload_hash, level, round) {
+        //        (Some(payload_hash), None, None) => ConsensusOperationMatcher::endorsement_payload(payload_hash),
```
tizoc, line 975: Another commented out case

rpc/src/services/dev_services.rs
akoptelov: Now it is not needed, a stub for the future.

rpc/src/services/dev_services.rs
akoptelov: The same as above.

PR #1035, opened 2022-04-04
Mempool set bound for refused operations

binier:


shell_automaton/src/mempool/mempool_state.rs
```diff
@@ -22,6 +22,13 @@ use crate::{
     prechecker::OperationDecodedContents, rights::Slot, service::rpc_service::RpcId, ActionWithMeta,
 };
 
+/// https://gitlab.com/tezedge/tezos/-/blob/tezedge-ffi/src/lib_shell/prevalidator.ml#L1452
```
tizoc, line 25: @binier lets switch this to a permalink using the commit hash (instead of the branch name `tezedge-ffi`, otherwise it is going to become stale if the code changes).

binier: Added tag in the url to prevent that.

PR #1041, opened 2022-04-07
Update issue templates

tizoc:
Issue template for bug reports


akoptelov (approved):
Doesn't it make sense to have expected behaviour right after describe the bug?

PR #1047, opened 2022-04-11
Mempool prevalidation priority

binier:


shell_automaton/src/mempool/mempool_state.rs
```diff
@@ -69,6 +69,71 @@ impl MempoolState {
...
+                    .map(|content| (hash, content))
+            })
+            .or_else(|| self.pending_operations.next_for_prevalidation())
+    }
+}
+
+#[derive(Serialize, Deserialize, Debug, Default, Clone)]
+pub struct MempoolPendingOperations {
+    ops: HashMap<OperationHash, Operation>,
+    consensus_ops: VecDeque<OperationHash>,
```
tizoc, line 89: 👌  for our current needs, this is perfect. Just one suggestion, wouldn't it be better to use a BTreeSet instead of a VecDeque? I see bellow that we need to iterate it comparing hashes to remove the entry when the operation is removed, which doesn't feel ideal.

shell_automaton/src/mempool/mempool_state.rs
```diff
@@ -69,6 +69,71 @@ impl MempoolState {
...
+            None
+        }
+    }
+
+    pub fn iter(&self) -> impl Iterator<Item = (&OperationHash, &Operation)> {
+        self.ops.iter()
+    }
+
+    /// Get next operation with highest priority for prevalidation.
+    pub fn next_for_prevalidation(&self) -> Option<(&OperationHash, &Operation)> {
```
tizoc, line 130: So this doesn't "pop" the operation right? it needs to be explicitly removed later for this call to returns something different (unless more operations are added), correct?

shell_automaton/src/mempool/mempool_state.rs
binier: @tizoc  I used `VecDeque` to guarantee FIFO order. I was thinking on using `BTreeSet`, but changed my mind because we don't really need that.

Only times when we need to remove an element from that:
- When we have validated an operation. It will always be first operation in that queue (removing of which is `O(1)`), unless it was injected + non-consensus operation that was validated. In that case yeah, we will iterate on whole list for no reason.
- When new block is applied. This happens every 15s and only once so I wouldn't worry about that being slow.

shell_automaton/src/mempool/mempool_state.rs
binier: Yes it doesn't pop. Only time we pop an operation is when operation was prevalidated or block was applied.

shell_automaton/src/mempool/mempool_state.rs
tizoc: Ok, sounds good, only thing I would add then is a comment mentioning this in `remove`, for our future selves.

shell_automaton/src/mempool/mempool_state.rs
```diff
@@ -69,6 +69,71 @@ impl MempoolState {
...
+    }
+
+    pub fn insert(&mut self, key: OperationHash, value: Operation) {
+        let op_kind = OperationKind::from_operation_content_raw(value.data().as_ref());
+        if self.ops.insert(key.clone(), value).is_none() {
+            if op_kind.is_consensus_operation() {
+                self.consensus_ops.push_back(key);
+            }
+        }
+    }
```
tizoc, line 112: One last thing, we are accumulating clippy warnings again, if we don't solve them we are soon going to be back at the same point we were a few weeks back.

tizoc (approved):

shell_automaton/src/mempool/mempool_state.rs
binier: let me fix those..

PR #1054, opened 2022-04-21
Fix bootstrap time

sebastiencs:


tizoc:
@sebastiencs solved in #1056

PR #1059, opened 2022-04-25
Add test to make a snashot and resume block application from it 

sebastiencs:
The test creates a snapshot from the in-memory context and restart the
node to apply blocks starting at the last context hash in the snapshot

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1128,6 +1128,103 @@ steps:
     # terminate the node gracefully and wait to allow `valgrind` dump data
     - kill -INT $(cat /var/run/tezedge.pid) && sleep 10
 
+- name: make-in-mem-snapshot-and-resume
+  image: tezedge/tezedge-ci-builder:rust-1.58.1-v12.0-octez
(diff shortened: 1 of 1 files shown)


.drone.yml
```diff
@@ -1128,6 +1128,103 @@ steps:
...
+        if [ -d "./replay-snapshot/context" ]; then
+          echo "found snapshot"
+          break
+        fi
+        attempts=$(($attempts + 1))
+      done
+      '
+    - echo "Bootstrapped. Killing the node and waiting for 10 seconds."
+    # terminate the node gracefully and wait to allow `valgrind` dump data
+    - kill -9 $(cat /var/run/tezedge.pid) && sleep 10
```
tizoc, line 1175: @sebastiencs is `-9` necessary? it is a bit aggressive, why doesn't SIGINT work?

sebastiencs: Ah yes, I used `-9` for testing but I forgot to change it back

PR #1060, opened 2022-04-26
Fix protocol-runner initialization with irmin

sebastiencs:
- Fix protocol-runner initialization with irmin
  Do not get latest context hashes with irmin-only storage. Irmin doesn't support that feature
- Use libc::syscall instead of libc::renameat2
  This handle when the wrapper is not available in the used libc


tizoc (approved):
Looks good.

PR #1061, opened 2022-04-26
Add action stats to the missing endorsement alerts and reports

binier:


apps/node_monitoring/src/monitors/delegate.rs
```diff
@@ -438,6 +460,13 @@ impl DelegatesMonitor {
             }
         }
 
+        let path = format!("/dev/shell/automaton/action_stats_for_blocks");
```
akoptelov, line 463: Shouldn't we query stats for near blocks here, like above?

apps/node_monitoring/src/monitors/delegate.rs
```diff
@@ -7,6 +7,7 @@ use std::{collections::HashMap, net::SocketAddr, time::Duration};
 
 use anyhow::Result;
 use crypto::hash::BlockHash;
+use itertools::Itertools;
```
akoptelov, line 10: Why do we need it?

akoptelov (approved):
There are a couple of comments/questions.

apps/node_monitoring/src/monitors/delegate.rs
binier: We need it for joining iterator of strings.

apps/node_monitoring/src/configuration.rs
```diff
@@ -437,6 +439,9 @@ impl DeployMonitoringEnvironment {
                 .unwrap_or("5")
                 .parse::<u64>()
                 .expect("Expected u64 value of seconds"),
+            explorer_url: args
```
akoptelov, line 442: You haven't introduced this argument to `clap`.

apps/node_monitoring/src/monitors/delegate.rs
binier: Yeah true, before we were only maintaining last 6 blocks, so I just simplified by storing all stats. Will change that.

apps/node_monitoring/src/configuration.rs
binier: thanks, fixed.

apps/node_monitoring/src/monitors/delegate.rs
binier: but nvm, actually there is a limit for rpc. on how many block stats can be retrieved. Limit is set to 64.
https://github.com/tezedge/tezedge/blob/2779c7e43988e51e28133c3c0b8fd81f822a0eea/shell_automaton/src/rpc/rpc_effects.rs#L145

So this will fetch last 64 blocks and store them. In report, more data might come in handy. @akoptelov what do you think?

PR #1062, opened 2022-04-26
Prevalidator rework

tizoc:
This PR switches to a new prevalidator implementation in the protocol runner (based on the current implementation in Octez), and uses a cleaner API.


tezos/interop/tests/test_rust_ocaml_conv.rs
```diff
@@ -447,7 +446,7 @@ fn test_validate_operation_request_conv() {
     let prevalidator = PrevalidatorWrapper {
         chain_id: ChainId::try_from(hex::decode(CHAIN_ID).unwrap()).unwrap(),
         protocol: get_protocol_hash(&[1, 2, 3, 4, 5, 6, 7, 8, 9]),
-        context_fitness: Some(vec![vec![0, 1], vec![0, 0, 1, 2, 3, 4, 5]]),
+        //context_fitness: Some(vec![vec![0, 1], vec![0, 0, 1, 2, 3, 4, 5]]),
```
binier, line 449: Shouldn't we just remove those commented `context_fitness` fields? Or do you wanna keep them in case we might need to add them back? I doubt we'll ever need it though.

binier (approved):
Just check out one comment that I wrote about `context_fitness`.

Otherwise it looks great! Much better api and code in general. Thanks :)

tezos/interop/tests/test_rust_ocaml_conv.rs
tizoc: I think we can remove them but I am not 100% sure yet. My current understanding is that the only reason this value was here is these tests (and I don't think we need that). Maybe it was used for something else before, but I don't see it anywhere else other than these tests.

PR #1064, opened 2022-04-28
add bisect exported functions: dump, reset_counters

dkuehr:
These functions will be available when libtezos is build with our fork of bisect_ppx instrumentation. This way they can be called from the Rust side to save coverage information of the OCaml components.


tezos/interop/src/lib.rs
```diff
@@ -23,6 +23,8 @@ mod tezos_ffi {
     use crate::TzResult;
 
     ocaml! {
+        pub fn dump(a: ());
```
tizoc, line 26: Let's move these (and the wrappers) to their own module (`tezos_fuzzing_ffi` or something like that) that is guarded by the fuzzing feature flag you use

tizoc (changes requested):
Looks good but lets separate the fuzzing stuff and make it only compile when fuzzing is enabled.

PR #1068, opened 2022-05-05
Interop bisect

dkuehr:
Allows building protocol-runner with coverage information which can be dumped from a running target by sending a SIGUSR2. This coverage information comes from two sources:
1. RUST: GCOV counters generated by rustc/llvm.
2. OCaml: bisect-ppx counters.

The SIGUSR2 handler is installed in the OCaml side. To dump the GCOV counters the `protocol_runner_dump_gcov` rust function is exported to OCaml.


protocol_runner/src/main.rs
```diff
@@ -17,6 +16,41 @@ use tikv_jemallocator::Jemalloc;
 #[global_allocator]
 static GLOBAL: Jemalloc = Jemalloc;
 
+#[cfg(dyncov)]
```
tizoc, line 19: Lets move all this dyncov stuff to it's own file, and then you can use a single

```rust
#[cfg(dyncov)]
mod dyncov;
```

Cargo.toml
```diff
@@ -59,7 +59,7 @@ opt-level = 3
 debug = 2
 debug-assertions = true  # TODO: test and disable if too slow
 overflow-checks = true
-lto = true
+lto = false
```
tizoc, line 62: why is this? can it be made optional just when fuzzing is enabled?

tizoc (changes requested):

Cargo.toml
dkuehr: This only affect the "fuzz" profile not normal builds. The change was needed because something has changed in nightly that breaks the build when building with gcov + lto.

tizoc (approved):

PR #1086, opened 2022-05-25
Protocol-runner call for preapplying a block without using RPC

tizoc:


tezos/api/src/ffi.rs
```diff
@@ -801,6 +805,31 @@ impl From<TezosErrorTrace> for FfiJsonEncoderError {
     }
 }
 
+#[derive(Serialize, Deserialize, Debug, Clone)]
```
tizoc, line 808: @binier here are the new request and response structs

tezos/protocol-ipc-client/src/lib.rs
```diff
@@ -456,6 +457,20 @@ impl ProtocolRunnerConnection {
         )
     }
 
+    /// Preapply block
+    pub async fn preapply_block(
```
tizoc, line 461: @binier and the new call that you can use

tezos/api/src/ffi.rs
```diff
@@ -801,6 +805,32 @@ impl From<TezosErrorTrace> for FfiJsonEncoderError {
...
+#[cfg_attr(feature = "fuzzing", derive(fuzzcheck::DefaultMutator))]
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct PreapplyBlockResponse {
+    pub level: Level,
+    pub proto: u8,
+    pub predecessor: BlockHash,
+    pub timestamp: Timestamp,
+    pub validation_pass: u8,
+    pub operations_hash: OperationListListHash,
+    pub fitness_bytes: Vec<u8>,
```
binier, line 829: Shouldn't this be `Vec<Vec<u8>>`?

PR #1088, opened 2022-05-26
Get validators from context

akoptelov:


akoptelov:
@binier Zura, this is the one for full Validators data.

PR #1108, opened 2022-06-09
Fix possible errors in RPC due to serde_json::Value conversions

adonagy:
Removed the old approach where the content was serialized as serde_json::Value and returned alongside the Response<Body> from the RPC handlers. Now we extract and deserialize the Response<Body> in the WebSocket RPC handler, so there is no need for additional conversions in the RPC crate. 

@tizoc This was causing additional errors for the stats RPCs uncovered yesterday by @binier . This approach also removes the concerns of performance regression in case of calling the hyper side (HTTP) of the RPCs (no additional serialization).

### DIFF
--- a/monitoring/src/websocket/ws_json_rpc.rs
+++ b/monitoring/src/websocket/ws_json_rpc.rs
@@ -2,7 +2,7 @@ use std::{collections::HashMap, pin::Pin};
 
 use rpc::{
     server::{parse_query_string, MethodHandler},
-    RpcServiceEnvironmentRef, ServiceResultBody,
+    RpcServiceEnvironmentRef,
 };
 use serde::{de::DeserializeOwned, Deserialize};
 use serde_json::Value;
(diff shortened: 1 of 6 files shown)


monitoring/src/websocket/ws_json_rpc.rs
```diff
@@ -58,7 +58,38 @@ pub async fn handle_request(
...
+                            "Failed to get bytes from hyper body",
+                        );
+                        return json_rpc_types::Response::error(
+                            req.jsonrpc,
+                            json_error,
+                            req.id.clone(),
+                        );
+                    }
+                };
+                let res = match serde_json::from_slice(&body_bytes) {
```
tizoc, line 77: @adonagy if you are always going to parse the already produced bytes here, cannot you just used `RawValue` to pass them as-is?

tizoc: iirc the previous hack was a workaround for serde not being able to properly parse a JSON value produced by the protocol, so we will have that issue again here (just not as bad because now the re-parsing happens only here and not in the regular handlers, so in the worst case only the WS RPC breaks).

PR #1122, opened 2022-06-16
Fix python test and use it for tezedge baker

vlad9486:


tezos/python-tests/daemons/baker.py
```diff
@@ -46,18 +46,14 @@ def __init__(
...
-                'run',
-                'with',
-                'local',
-                'node',
-                node_dir
-            ]
-            + accounts
-        )
+        if os.environ.get("RUN_TEZEDGE_BAKER") == "external":
+            cmd.extend([baker, '--base-dir', base_dir, '--endpoint', endpoint, '--archive'])
```
tizoc, line 52: @vlad9486 the initial `baker` here is not needed right? it is already part of the original `cmd` list

.drone.yml
```diff
@@ -1297,6 +1297,99 @@ volumes:
...
+      - cp drone-cache/build_files/tezedge-baker tezos-src/tezos/
+      - cp drone-cache/build_files/ffi/libtezos.so tezos-src/tezos/
+      # copy zcash-params init files - see constants.py light-node params
+      - cp drone-cache/build_files/ffi/sapling-spend.params tezos-src/tezos/
+      - cp drone-cache/build_files/ffi/sapling-output.params tezos-src/tezos/
+
+  - name: baker-endorser-test
+    <<: *python-test
+    environment:
+      PYTHON_TEST: test_baker_endorser.py
```
tizoc, line 1391: Two questions:

- Does this really need a full dedicated pipeline instead of being added as an extra test to the already existing pipelines?
- Other than the environment differences (using our own baker instead of octez), is there any difference in the test code itself?

tezos/python-tests/daemons/baker.py
vlad9486: @tizoc  you are right

.drone.yml
vlad9486: I improved one test `test_baker_endorser.py`. And maybe improve more after this merged.

PR #1127, opened 2022-06-17
Tezedge private network

akoptelov:


akoptelov:
@binier I think we should add your script here to grab keys from ledger's mnemonics,

PR #1130, opened 2022-06-20
Fix tezos-encoding `Dynamic` documentation + add further `derive`s for `Zarith` & `BigInt`

emturner:
# Overview
Thanks so much for these crates!  They've massively simplified interacting with tezos encodings from rust

- `Dynamic`/`ShortDynamic` docs *prefix length* look to be the wrong way round
- when using `Zarith` in encodings, wasn't able to auto-derive `PartialEq` & `Eq`, which are supported by `num_bigint`, so have added those derivations
- derive `Clone` for `BigInt`, which is really useful when writing tests involving `BigInt`


emturner:
thanks for the review @akoptelov!

I've updated the copyright headers to include my employer (Trili Tech) - is that ok?  I've also sent a message to @jurajselep letting him know

PR #1138, opened 2022-06-25
Allow None for pred metadata ops

akoptelov:
When using protocol activation block as predecessor (with validation_pass = 0), its ops metadata hash should be None, to match Ocaml behavior.


akoptelov:
@binier The fix is to allow predecessor with no-ops to have `None` as its `ops_metadata_hash`. According to @tizoc, this is how Octez node (and protocol) behaves.
See here: https://gitlab.com/tezedge/tezos/blob/a925eda16fc0520e5fddfb746ed025022b7add17/src/lib_store/store.ml#L725

tizoc (approved):
Looks good to me.

tizoc:
Btw, the address-sanitizer version of the Github actions runners is not able to finish anymore, I think the VM/container where the build run just doesn't have enough RAM to build the automaton, so it gets killed.

